PR #16, opened 2019-11-27
update debits module

qwer951123:
* replace Trait::Currency bound to MultiCurrencyExtended

* add Trait::GetStableCurrencyId

* fix ci tests


modules/cdp_engine/src/mock.rs
```diff
@@ -42,8 +44,9 @@ pub type Amount = i64;
 pub type CurrencyId = u32;
 pub const ALICE: AccountId = 1;
 pub const BOB: AccountId = 2;
+
+pub const ACA: CurrencyId = 0;
 pub const AUSD: CurrencyId = 0;
```
xlc:
```suggestion
pub const AUSD: CurrencyId = 1;
```

PR #70, opened 2020-01-17
feature/accounts

### LINKED ISSUE #58: Accounts module
### Accounts module

- Trait
  - FreeTransferCount: Get\<u8>
  - FreeTransferPeriod: Get\<Moment>
  - Time: Time
  - Call: Parameter + Dispatchable<Origin=<Self as frame_system::Trait>::Origin> + IsSubType\<orml_currencies::Module\<Self>, Self>;
  - Currency: Currency
- Storages
  - LastFreeTransfers: map AccountId => Vec\<Moment>
    - Store last `FreeTransferCount` free transfer time for each account
- Module
  - impl OnReapAccount
    - clear `LastFreeTransfers` for reaped account
  - try_free_transfer(who: AccountId) -> bool
    - Read `LastFreeTransfers`, remove all the expired entries (`value < now - FreeTransferPeriod`), return true and append `now` to the list if entries count less than `FreeTransferCount` otherwise return false
- ChargeTransactionPayment: SignedExtension
  - Similar to `ChargeTransactionPayment` in `pallet-transaction-payment` module
  - it additionaly call `try_free_transfer` and skip the withdraw fee part if the call is currencies.transfer and user still have free transfer
  - check `pallet-contracts` `CheckBlockGasLimit` for how to check call type
  - Use `transaction_payment::ChargeTransactionPayment::from(tip)::compute_fee` to reduce copy & paste

Update runtime to use the new `ChargeTransactionPayment` instead of the one in transaction payment module

We will still need additional requirements for an account to be able to make free transfer. This is still TBD.

wangjj9219:
Closes #58

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -2631,6 +2631,26 @@ dependencies = [
 ]
 
 [[package]]
+name = "module-accounts"
+version = "0.0.1"
+dependencies = [
+ "frame-support",
+ "frame-system",
+ "module-support",
+ "orml-currencies",
+ "orml-tokens",
+ "orml-traits",
+ "pallet-balances",
+ "pallet-transaction-payment",
+ "parity-scale-codec",
+ "serde",
+ "sp-core",
+ "sp-io",
+ "sp-runtime",
+ "sp-std",
+]
+
+[[package]]
 name = "module-auction-manager"
 version = "0.0.1"
 dependencies = [
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -2638,6 +2638,7 @@ dependencies = [
  "orml-tokens",
  "orml-traits",
  "pallet-balances",
+ "pallet-timestamp",
  "pallet-transaction-payment",
  "parity-scale-codec",
  "serde",
--- a/modules/accounts/Cargo.toml
+++ b/modules/accounts/Cargo.toml
@@ -1,0 +1,40 @@
+[package]
+name = "module-accounts"
+version = "0.0.1"
+authors = ["Acala Developers"]
+edition = "2018"
+
+[dependencies]
+serde = { version = "1.0", optional = true }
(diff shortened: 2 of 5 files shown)


wangjj9219:
`ChargeTransactionPayment ` in `pallet_transaction_payment` doesn't expose the `compute_fee` ，So I copy the code.

https://github.com/paritytech/substrate/blob/7248ac20c4386b30804dcad1bdeeed4c0a71b6ca/frame/transaction-payment/src/lib.rs#L154-L155

https://github.com/AcalaNetwork/Acala/blob/ac3e5a4b10226ffc5ea20c3fd3cb3b42b8bf6915/modules/accounts/src/lib.rs#L26-L34

And define `type Call` satisfied `IsSubType<orml_currencies::Module<Self>, Self>` ，If do not declare Trait derives from orml_currencies::Trait , will get following error ， is that ok？ 

```
error[E0277]: the trait bound `Self: orml_currencies::Trait` is not satisfied
  --> modules/accounts/src/lib.rs:26:1
   |
26 | / pub trait Trait: system::Trait + pallet_transaction_payment::Trait {
27 | |     type FreeTransferCount: Get<u8>;
28 | |     type FreeTransferPeriod: Get<MomentOf<Self>>;
29 | |     type Time: Time;
...  |
33 | |         + IsSubType<orml_currencies::Module<Self>, Self>;
34 | | }
   | |_^ the trait `orml_currencies::Trait` is not implemented for `Self`
   |
   = note: required because of the requirements on the impl of `sp_api_hidden_includes_decl_storage::hidden_include::Callable<Self>` for `orml_currencies::Module<Self>`
   = note: required by `sp_api_hidden_includes_decl_storage::hidden_include::IsSubType`
```

xlc:
Looks reasonable. I will open a PR upstream to make compute_fee public

xlc:
#72 should include https://github.com/paritytech/substrate/pull/4669

PR #80, opened 2020-01-27
accounts module update

### LINKED ISSUE #74: Accounts module update
- Trait
  - type FreeTransferDeposit: Get\<Balance>
  - type DepositCurrency: LockableCurrency
- Storages
  - FreeTransferEnabledAccounts: map AccountId => Option\<()>
- Calls
  - fn enable_free_transfer(origin)
    - set lock with `FreeTransferDeposit` amount, max expire date, all reasons
    - FreeTransferEnabledAccounts::put(origin, ())
  - fn disable_free_transfers(origin)
    - remove lock
    - FreeTransferEnabledAccounts::kill(origin)
- Module
  - update try_free_transfer
    - Check FreeTransferEnabledAccounts::exists(who) and return false if did not exist

Also 

https://github.com/AcalaNetwork/Acala/blob/c3bf35f8ece1bb06cfb1289c7fcda97b7a6fe051/modules/accounts/src/lib.rs#L134-L138

is same as `<Module<T>>::try_free_transfer(who)`

wangjj9219:
close #74

### DIFF
--- a/modules/accounts/src/lib.rs
+++ b/modules/accounts/src/lib.rs
@@ -24,9 +24,8 @@ use system::ensure_signed;
 mod mock;
 mod tests;
 
-const ACCOUNTS_ID: LockIdentifier = *b"accounts";
+const ACCOUNTS_ID: LockIdentifier = *b"ACA/acct";
 
-//type BalanceOf<T> = <<T as Trait>::Currency as MultiCurrency<<T as system::Trait>::AccountId>>::Balance;
 type MomentOf<T> = <<T as Trait>::Time as Time>::Moment;
 type PalletBalanceOf<T> =
 	<<T as pallet_transaction_payment::Trait>::Currency as Currency<<T as system::Trait>::AccountId>>::Balance;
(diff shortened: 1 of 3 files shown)


modules/accounts/src/lib.rs
```diff
@@ -4,47 +4,73 @@ use codec::{Decode, Encode};
...
 	transaction_validity::{
 		InvalidTransaction, TransactionPriority, TransactionValidity, TransactionValidityError, ValidTransaction,
 	},
 };
+use system::ensure_signed;
 
 mod mock;
 mod tests;
 
+const ACCOUNTS_ID: LockIdentifier = *b"accounts";
```
xlc: better to prefix those with `ACA`

```suggestion
const ACCOUNTS_ID: LockIdentifier = *b"ACA/acct";
```

modules/accounts/src/lib.rs
```diff
@@ -4,47 +4,73 @@ use codec::{Decode, Encode};
...
+
 //type BalanceOf<T> = <<T as Trait>::Currency as MultiCurrency<<T as system::Trait>::AccountId>>::Balance;
```
xlc: remove this if it is not used

Issue #65, opened 2020-01-14
Add staking module and treasury module


xlc:
Add `pallet-staking` and `pallet-treasury` module and connects them.

Make burn rate for treasury zero.

PR #84, opened 2020-02-03
Feature/intergrate staking

wangjj9219:
#65 

1. add pallet_treasury,  pallet_staking modules to runtime
2. add pallet_session to implement pallet_staking::Trait::SessionInterface
3. replace aura consensus to babe consensus because of pallet_session::Trait:: ShouldEndSession

Still need to do：
1. regenerate alphanet genesis according to babe consensus
2. now，pallet_session in runtime, need add pallet_authority_discovery module?

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -82,6 +82,7 @@ dependencies = [
  "pallet-timestamp",
  "pallet-transaction-payment",
  "pallet-transaction-payment-rpc-runtime-api",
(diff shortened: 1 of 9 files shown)


xlc:
We shouldn't need authority discovery because we won't have much validators in network. But if the code doesn't work without it, we can add it.

wangjj9219:
close #65

PR #111, opened 2020-02-16
Offchain worker

wangjj9219:
目前有几个问题还未解决：
1. rand包在module_cdp_engine模块中使用正常， 但集成到runtime时会提示无法找到rand包， 在 /runtime/Cargo.toml 同样配置有rand仍然提示错误。
2. 参考pallet_im_online模块配置了Authority， 用于offchain worker 签名 UnsignedTransaction ， 节点收到后会验证交易是否由Authority Set其中一员签名， 是否要在SessionKey中加入CdpEngineId？  这样当验证人节点配置CdpEngineId时， Authority Set就由pallet_session模块自动维护。 不加入SessionKey的话， 可能需要一套单独的维护方式（目前是由sudo账户添加）。 还是说不需要Authority， SignedExtension将不会对消息来源进行检查。
3.  对UnsignedTx的验证使用了不推荐的ValidateUnsigned， 稍后会SignedExtension替代


xlc:
避免直接用 `rand`，offchain_worker api 有 random_seed，可以配合 `RandomNumberGenerator` 用

https://github.com/paritytech/substrate/blob/2e8080e2902fc477bbce36512a8f5bcdc4b49f17/client/offchain/src/api.rs#L92

https://github.com/paritytech/substrate/blob/2e8080e2902fc477bbce36512a8f5bcdc4b49f17/primitives/runtime/src/random_number_generator.rs#L46

modules/loans/src/lib.rs
```diff
@@ -43,7 +43,7 @@ pub trait Trait: system::Trait {
 
 decl_storage! {
 	trait Store for Module<T: Trait> as Loans {
-		pub Debits get(fn debits): double_map hasher(blake2_256) T::AccountId, hasher(blake2_256) CurrencyIdOf<T> => T::DebitBalance;
+		pub Debits get(fn debits): double_map hasher(blake2_256) CurrencyIdOf<T>, hasher(blake2_256) T::AccountId => (T::DebitBalance, Option<(CurrencyIdOf<T>, T::AccountId)>);
```
xlc: I want to do some changes upstream to make iterator returns keys as well. so we can use blake2_128_concat hasher and getting keys as well.

Also this needs migration if we want to runtime upgrade Mandala. But after we change the key hasher so we don't need to migrate twice.

modules/cdp_engine/src/lib.rs
```diff
@@ -207,10 +249,126 @@ decl_module! {
...
+
+					Ok(OffchainWorkerLock {
+						previous_position: execute_position,
+						locked: true,
+					})
+				}
+				_ => Err(OffchainErr::FailedToAcquireLock),
+			}
+		})?;
+		let _ = require_lock.map_err(|_| OffchainErr::FailedToAcquireLock)?;
```
xlc: Can we do something like this to avoid `mut` execute_position. Also finding the current flow bit hard to understand.
`let OffchainWorkerLock { previous_position } = require_lock.map_err(|_| OffchainErr::FailedToAcquireLock)?;`

modules/loans/src/lib.rs
```diff
@@ -41,7 +41,7 @@ pub trait Trait: system::Trait {
 
 decl_storage! {
 	trait Store for Module<T: Trait> as Loans {
-		pub Debits get(fn debits): double_map hasher(blake2_256) T::AccountId, hasher(blake2_256) CurrencyIdOf<T> => T::DebitBalance;
+		pub Debits get(fn debits): double_map hasher(blake2_128_concat) CurrencyIdOf<T>, hasher(blake2_128_concat) T::AccountId => (T::DebitBalance, Option<(CurrencyIdOf<T>, T::AccountId)>);
```
xlc:
```suggestion
		pub Debits get(fn debits): double_map hasher(twox_64_concat) CurrencyIdOf<T>, hasher(twox_64_concat) T::AccountId => (T::DebitBalance, Option<(CurrencyIdOf<T>, T::AccountId)>);
```

CurrencyId and AccountId can use `twox_64_concat` as it is not something user can arbitrary crafted to clash the raw key.

Use `blake2_128_concat` for values that user have full control of, like `Balance`, `Vec<u8>`

modules/loans/src/lib.rs
```diff
@@ -89,6 +89,12 @@ decl_module! {
 }
 
 impl<T: Trait> Module<T> {
+	pub fn debits_iterator_with_collateral_prefix(
+		currency_id: CurrencyIdOf<T>,
+	) -> PrefixIterator<(T::DebitBalance, Option<(CurrencyIdOf<T>, T::AccountId)>)> {
```
xlc: I want to modify `PrefixIterator` in Substrate to make it return Keys as well but I guess for now we will just copy and modify the `PrefixIterator` from Substrate and later switch to whatever Substrate provides

https://github.com/paritytech/substrate/issues/4959

Issue #126, opened 2020-02-26
Add some operations related to emergency shutdown in offchain worker


wangjj9219:
run these operations by offchain worker when emergency shutdown occurs:

- [x] 1. settle CDPs that have debit (resolved by #128)
- [ ] 2. cancel surplus auctions and debit auctions
- [ ] 3. cancel collateral auctions which are still in forward stage

PR #128, opened 2020-02-26
offchain worker settles CDP after emergency shutdown happened

wangjj9219:
related to #126 

now,  offchain woker in cdp module will automatically settle all collaterals type CDPs which have debit after after emergency shutdown happend.


modules/cdp_engine/src/lib.rs
```diff
@@ -281,12 +287,12 @@ decl_module! {
 }
 
 impl<T: Trait> Module<T> {
-	fn required_liquidator_lock() -> Result<LiquidatorLock, OffchainErr> {
+	fn required_offchain_worker_lock() -> Result<OffchainWorkerLock, OffchainErr> {
```
xlc: I think `acquire_offchain_worker_lock ` is a better name?

modules/cdp_engine/src/lib.rs
```diff
@@ -354,66 +360,106 @@ impl<T: Trait> Module<T> {
...
 		Ok(())
 	}
 
 	fn liquidate_specific_collateral(currency_id: CurrencyIdOf<T>) {
 		for (_, key) in <loans::Module<T>>::debits_iterator_with_collateral_prefix(currency_id) {
 			if let Some((_, account_id)) = key {
-				// TODO: liquidate unsafe cdp before emergency shutdown, settle cdp with debit when emergency shutdown occurs.
 				if Self::is_unsafe_cdp(currency_id, &account_id) {
 					if let Err(e) = Self::submit_unsigned_liquidation_tx(currency_id, account_id.clone()) {
 						debug::debug!(
```
xlc: error case should use INFO/WARN and success case can be DEBUG

Issue #140, opened 2020-03-03
Put the actual amount got by swap into the returning result of functions in DEX module


wangjj9219:
cdp_treasury need the exact value when swap token to dex module

wangjj9219:
close by #141

PR #141, opened 2020-03-03
fix calculation to make sure get_supply_amount can afford target amount

### LINKED ISSUE #137: The result of `dex::get_supply_amount` is not correct
Because the div will discard the remainder, the result of the `dex::get_supply_amount` is not correct. use the incorrect result as supply amount to exchange with dex, the actual maxmum avalible target amount are most likely less than the specified target amount, lead to trading failed. 

For each div in the calculation process, if there is a remainder, the quotient needs to be increased by 1  to ensure that the final result is sufficient for trading.

wangjj9219:
fix #137 

In calculation steps, when multiple an FixedU128,  adding 1 to the result. The result may be a bit larger, but can ensure to swap target amount successfully.

Later, will decide whether to add 1 or not based on the remainder, that needs FixedU128 support new multiple function.

### DIFF
--- a/modules/cdp_engine/src/lib.rs
+++ b/modules/cdp_engine/src/lib.rs
@@ -608,7 +608,7 @@ impl<T: Trait> Module<T> {
 
 		// forth: handle bad debt and collateral
 		if !supply_amount.is_zero() 				// supply_amount must not be zero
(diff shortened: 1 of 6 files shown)


wangjj9219:
resolve #140

PR #174, opened 2020-04-01
cancel auction by offchain worker after shutdown

### LINKED ISSUE #126: Add some operations related to emergency shutdown in offchain worker

wangjj9219:
close #126


modules/auction_manager/src/lib.rs
```diff
@@ -112,10 +133,134 @@ decl_module! {
...
+						e,
+					);
+				}
+			}
+		}
 	}
 }
 
 impl<T: Trait> Module<T> {
+	fn acquire_offchain_worker_lock() -> Result<Timestamp, OffchainErr> {
```
xlc, line 160: we should be able to abstract the offchain worker lock logic into orml utilities to avoid code duplications

modules/auction_manager/src/lib.rs
```diff
@@ -690,3 +835,38 @@ impl<T: Trait> AuctionManager<T::AccountId> for Module<T> {
...
+					return InvalidTransaction::Stale.into();
+				}
+			} else if !<SurplusAuctions<T>>::contains_key(auction_id) && !<DebitAuctions<T>>::contains_key(auction_id) {
+				return InvalidTransaction::Stale.into();
+			}
+
+			Ok(ValidTransaction {
+				priority: TransactionPriority::max_value(),
+				requires: vec![],
+				provides: vec![(<system::Module<T>>::block_number(), auction_id).encode()],
```
xlc, line 864: no need to have block number in tag. we can't cancel an auction twice with different block numbers

add some unique byte to ensure the provides tag is unique.

PR #233, opened 2020-05-19
orml-tokens benchmarking

shaopengw:


wangjj9219:
It seems to be bug with orml tokens, I will look for it and fix

shaopengw:
As discussed with @xlc , existential deposit of tokens is introduced. It should be this change causing the build failures.

```rust
pub const TokenExistentialDeposit: Balance = 10 * MICROCENTS;
```

PR #245, opened 2020-05-25
add auction lots cap to avoid too much iterations affecting block produce

wangjj9219:


modules/cdp_treasury/src/lib.rs
```diff
@@ -44,6 +44,11 @@ pub trait Trait: system::Trait {
 
 	/// Dex manager is used to swap confiscated collateral assets to stable coin
 	type DEX: DEXManager<Self::AccountId, CurrencyId, Balance>;
+
+	/// the cap of lots number when create collateral auction on a liquidation
+	/// or to create debit/surplus auction on block end.
+	/// If set to 0, does not work.
+	type AuctionLotsCap: Get<u32>;
```
xlc, line 51: MaxAuctionsCount?

runtime/src/lib.rs
```diff
@@ -738,13 +739,18 @@ impl module_dex::Trait for Runtime {
 	type UpdateOrigin = pallet_collective::EnsureProportionMoreThan<_1, _2, AccountId, HonzonCouncilInstance>;
 }
 
+parameter_types! {
+	pub const AuctionLotsCap: u32 = 10;
```
xlc, line 743: 10 seems a bit too low?

PR #253, opened 2020-05-29
update collateral auctin end handler

### LINKED ISSUE #247: Improve collateral auction when below target using DEX
When collateral auction ended, if the bid is below target, check if sell to DEX can result a better price, and if yes, sell to DEX instead.

This should reduce debits in case people are not participate auctioning for whatever reason.

wangjj9219:
resolve #247 

@xlc  The stablecoin obtained from trading with DEX may still be larger than the target. Should the excess part be returned to auction owner? Or keep it as system surplus.

### DIFF
--- a/modules/auction_manager/src/lib.rs
+++ b/modules/auction_manager/src/lib.rs
@@ -32,7 +32,7 @@ use sp_std::{
 	cmp::{Eq, PartialEq},
 	prelude::*,
 };
-use support::{AuctionManager, CDPTreasury, OnEmergencyShutdown, PriceProvider, Rate};
+use support::{AuctionManager, CDPTreasury, CDPTreasuryExtended, DEXManager, OnEmergencyShutdown, PriceProvider, Rate};
 use utilities::{OffchainErr, OffchainLock};
 
 mod mock;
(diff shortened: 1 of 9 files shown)


xlc:
Ideally only enough collateral for target price, and remaining collateral refund to cdp owner.

But otherwise sell all collateral and refund extra aUSD to cdp owner could also work.

modules/auction_manager/src/lib.rs
```diff
@@ -687,34 +692,79 @@ impl<T: Trait> Module<T> {
...
-				// update auction records
-				TotalCollateralInAuction::mutate(collateral_auction.currency_id, |balance| {
-					*balance = balance.saturating_sub(collateral_auction.amount)
-				});
-				TotalTargetInAuction::mutate(|balance| *balance = balance.saturating_sub(collateral_auction.target));
-				<CollateralAuctions<T>>::remove(auction_id);
+			if !should_deal {
+				// shouldn't deal, exchange with DEX can get a better result.
+				// refund stable coin to the last bidder, ignore result to continue
+				if T::CDPTreasury::deposit_unbacked_debit_to(&bidder, bid_price).is_ok() {
```
xlc, line 712: if swap failed for whatever reason, should also fallback to deal auction

PR #255, opened 2020-05-29
optimize map storage

### LINKED ISSUE #249: Review and optimize storage usage
From Gav:

> It's bad form to have three separate map items (four if you count BountyApprovals, which should probably also be associated with BountyIndex) since it will place 3x the number of items in the trie needlessly. Unless each item contains unconstrained encode/decode complexity (they don't here), then just place them all in one item.
> 
> If you ever see multiple maps in the same pallet with the same key type, it's a red flag.

Basically try to merge storages as much as possible.

wangjj9219:
resolve #249

@xlc , I'm not sure if should merge `Collaterals` and `Debits` , `TotalCollaterals` and `TotalDebits` of `module_loans`. Because offchain worker of `module_cdp_engine` can only find cdp has debit by iterating `Debits`. If iterate the new item which merged by `Collaterals` and `Debits`, will get many cdp without debits.

### DIFF
--- a/modules/cdp_engine/src/lib.rs
+++ b/modules/cdp_engine/src/lib.rs
@@ -93,6 +93,32 @@ pub enum LiquidationStrategy {
 	Exchange,
 }
 
+/// Risk management params
+#[derive(Encode, Decode, Clone, RuntimeDebug, PartialEq, Eq, Default)]
+pub struct RiskManagementParams {
+	/// Maximum total debit value generated from it, when reach the hard cap,
+	/// CDP's owner cannot issue more stablecoin under the collateral type.
+	pub maximum_total_debit_value: Balance,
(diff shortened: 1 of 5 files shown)


wangjj9219:
@qwer951123 , this PR will change lots of storage, review changes and update related front end, SDK and bot

xlc:
There is `OffchainIndex` API allow validator to write to offchain worker DB some data when importing blocks so we can maintain a separate debit list in offchian DB.

But I am not sure how usable the offchain index API it is and this is very new (and untested). So we can do this later.

https://github.com/paritytech/substrate/blob/acca0ad3c0eae946231f2a9e27eeded6350dcbcf/primitives/io/src/lib.rs#L712-L726

PR #259, opened 2020-06-02
GH actions cache.

### LINKED ISSUE #258: Introduce cache in CI.
With clippy integrated in CI, now it takes about 1 hour to complete a GH action. It could be much faster with cache.

shaopengw:
Closes #258

### DIFF
--- a/.github/workflows/test.yml
+++ b/.github/workflows/test.yml
@@ -15,14 +15,13 @@ jobs:
     - uses: actions/checkout@v2
       with:
         submodules: recursive
-    - name: Cache cargo
+    - name: Cache cargo registry and index
       uses: actions/cache@v2
       with:
         path: |
           ~/.cargo/registry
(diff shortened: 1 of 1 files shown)


shaopengw:
It seems the cache is too big https://github.com/AcalaNetwork/Acala/actions/runs/122066457

xlc:
https://help.github.com/en/actions/configuring-and-managing-workflows/caching-dependencies-to-speed-up-workflows
> GitHub will remove any cache entries that have not been accessed in over 7 days. There is no limit on the number of caches you can store, but the total size of all caches in a repository is limited to 5 GB. If you exceed this limit, GitHub will save your cache but will begin evicting caches until the total size is less than 5 GB.

`target` is very likely larger than 5 GB.

shaopengw:
Will remove `target`, in the future may be we could use a self-hosted runner.

PR #262, opened 2020-06-04
update orml

xlc:


wangjj9219:
`saturating_from_rational` of `FixedU128` will panic when denominator is zero, so I replace `saturating_from_rational` by `checked_from_rational` in calculation.

PR #269, opened 2020-06-09
support fee refund for accounts according to pallet_transaction_payment

### LINKED ISSUE #267: Update accounts module to include new changes from pallet-transaction-payment
Need to update to include logic from

https://github.com/paritytech/substrate/pull/6165

wangjj9219:
close #267


modules/accounts/src/lib.rs
```diff
@@ -82,8 +85,8 @@ decl_module! {
 }
 
 impl<T: Trait> Module<T>
-where
-	T::Call: Dispatchable<Info = DispatchInfo, PostInfo = PostDispatchInfo>,
+// where
```
xlc, line 88: can be removed

PR #287, opened 2020-06-23
Fix accounts

wangjj9219:
1. add slippage limit when swap fee or open account deposit with DEX to avoid unreasonable losses.
2. configure Accounts as OnReceived of orml_tokens
3. fix some tests in integrate test


modules/accounts/src/lib.rs
```diff
@@ -285,34 +291,38 @@ impl<T: Trait> OnReceived<T::AccountId, CurrencyId, Balance> for Module<T> {
...
+					)
+					.is_ok()
+				{
+					// successful swap will cause changes in native currency,
+					// which also means that it will open a new account
+				} else {
+					// open account will fail because there's no enough native token,
+					// transfer all token as dust to treasury account.
+					let treasury_account = Self::treasury_account_id();
+					if who.clone() != treasury_account {
```
xlc, line 318:
```suggestion
					if *who != treasury_account {
```

Issue #320, opened 2020-07-23
Collateral auctions optimization


shaopengw:
### The Problem

In some collateral auctions, the amount is too big and needs to be divided to several lots for auction. The current strategy in implementation in such cases is like:
- Keep dividing remained collateral amount as `CollateralAuctionMaximumSize` per lot.
- If `MaxAuctionsCount` reached, put the remain collateral in one last auction.

This leads to imbalanced auction size in some cases, the last lot could be too small or big. For instance, in the case of *100* unit as max size, and *100* as max counts:
- When collateral amount is, say *210*, there will be 3 lots: *[100, 100, 10]*
- When collateral amount is too big, like *12000*, there will be 100 lots: *[100, 100, ..., 2000]*

### Optimization

To solve the imbalanced auction lots, we could optimize the auction division to always ensure equal sized lot as below:
1. Use `CollateralAuctionMaximumSize` to decide how many lots required, as `n`. And the size of each lot is `collateral_amount / n`.
2. If lots count exceeds `MaxAuctionsCount`, the size of each lot is `collateral_amount / MaxAuctionsCount`

With this strategy, the above examples would be
- When collateral amount is *210*, the 3 lots would be *[70, 70, 70]*.
- When collateral amount is *12000*, the 100 lots would be *[120, 120, ..., 120]*

shaopengw:
Closes by #322

PR #322, opened 2020-07-23
refactor collateral auction

wangjj9219:


wangjj9219:
resolve #320

modules/cdp_engine/src/lib.rs
```diff
@@ -713,9 +713,8 @@ impl<T: Trait> RiskManager<T::AccountId, CurrencyId, Balance, Balance> for Modul
 		collateral_balance: Balance,
 		debit_balance: Balance,
 	) -> DispatchResult {
-		let debit_value = Self::get_debit_value(currency_id, debit_balance);
-
-		if !debit_value.is_zero() {
+		if !debit_balance.is_zero() {
```
wangjj9219, line 716: fix bug maybe lead tiny debit balance bypass MininumDebitBalance check due to caculation loss

modules/auction_manager/src/lib.rs
```diff
@@ -59,7 +59,9 @@ pub struct CollateralAuctionItem<AccountId, BlockNumber> {
 	/// current collateral amount for sale
 	#[codec(compact)]
 	amount: Balance,
-	/// Target sales amount want to get by this auction
+	/// Target sales amount want to get by this auction,
+	/// if zero, collateral auction will never be reverse stage,
+	/// otherwise, target amount is the actual payment amount of active bidder
```
wangjj9219, line 64: Consider the case which the target is zero, although it is unlikely to happen in the system. If target is zero, the collateral auction will never in reverse stage.

modules/cdp_treasury/src/lib.rs
```diff
@@ -415,30 +415,39 @@ impl<T: Trait> CDPTreasuryExtended<T::AccountId> for Module<T> {
...
+			let lots_count = if max_auctions_count.is_zero()
+				|| collateral_auction_maximum_size.is_zero()
+				|| amount <= collateral_auction_maximum_size
+			{
+				One::one()
+			} else {
+				let mut count = amount
+					.checked_div(collateral_auction_maximum_size)
+					.expect("collateral auction maximum size is not zero; qed");
+				if !(amount % collateral_auction_maximum_size).is_zero() {
```
xlc, line 435: use `checked_rem`

PR #337, opened 2020-07-30
Refactor staking pool of homa

### LINKED ISSUE #330: refactor bind of Homa protocol
When bind DOT to issue LDOT by Homa protocol,  should determine whether to directly staking or put it into the free pool by `MinBondRatio` and `MaxBondRatio` of staking_pool, not always staking the DOT

wangjj9219:
1. fix bug of `redeem_by_unbond`: It's wrong that decrease `TotalBonded` while accumulating unbond of `NextEraUnbond`, this will cause loss to users who will redeem later.

2. resolve #330 , shouldn't bond all DOT when mint,  only bond part or all of DOT when the `communal_bonded_ratio` of staking pool is below `MinBondRatio`. The remainder is kept in `free_pool`.

3. make arithetic safer

### DIFF
--- a/modules/staking_pool/src/lib.rs
+++ b/modules/staking_pool/src/lib.rs
@@ -47,8 +47,8 @@ decl_event!(
 		<T as system::Trait>::AccountId,
 		Balance = Balance,
 	{
-		/// [who, issued_liquid, bonded_staking, left_in_free]
-		MintLiquid(AccountId, Balance, Balance, Balance),
+		/// [who, bond_staking, issued_liquid]
+		MintLiquid(AccountId, Balance, Balance),
 		/// [who, redeem_amount, unbond_amount]
 		RedeemByUnbond(AccountId, Balance, Balance),
 		/// [who, fee_in_liquid, liquid_amount_burned, staking_amount_retrived]
@@ -61,7 +61,6 @@ decl_event!(
 decl_error! {
 	/// Error for staking pool module.
 	pub enum Error for Module<T: Trait> {
-		StakingCurrencyNotEnough,
 		LiquidCurrencyNotEnough,
 		InvalidEra,
 		Overflow,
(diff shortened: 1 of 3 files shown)


modules/staking_pool/src/lib.rs
```diff
@@ -440,8 +487,47 @@ impl<T: Trait> OnNewEra<EraIndex> for Module<T> {
...
+			.checked_add(amount)
+			.and_then(|n| min_bond_ratio.checked_mul_int(n))
+			.map(|n| n.saturating_sub(communal_bonded))
+			.ok_or(Error::<T>::Overflow)?;
+
+		let amount_to_bond = sp_std::cmp::min(gap_amount, amount);
+		let remainer_to_free = amount.saturating_sub(amount_to_bond);
+
+		// bond
+		Self::bond(who, amount_to_bond)?;
```
xlc, line 519: I think we can just deposit everything into free pool. And let rebalance to do the bonding.

modules/staking_pool/src/lib.rs
```diff
@@ -239,56 +208,120 @@ impl<T: Trait> Module<T> {
...
+			// communal_bonded_staking_amount is not enough, re-calculate
+			if staking_amount_to_unbond > communal_bonded_staking_amount {
+				liquid_amount_to_redeem = liquid_exchange_rate
 					.reciprocal()
 					.unwrap_or_default()
-					.saturating_mul_int(free_unbonded);
+					.saturating_mul_int(communal_bonded_staking_amount);
+				staking_amount_to_unbond = liquid_exchange_rate
+					.checked_mul_int(liquid_amount_to_redeem)
+					.ok_or(Error::<T>::Overflow)?;
```
shaopengw, line 230: Here `staking_amount_to_unbound` calculation result would be equal to `communal_bonded_staking_amount`, if I didn't miss something.

PR #342, opened 2020-08-03
Merge collateral and debit storage into position.

shaopengw:
Loans module: merge collateral and debit storage into position, to improve performance and reduce storage size. (For the similar reason of pallet-balances refactor in https://github.com/paritytech/substrate/issues/4613 )


wangjj9219:
@qwer951123 need to update SDK, front end and data service after merging

PR #352, opened 2020-08-04
rpc support hex format number

### LINKED ISSUE #349: RPC should support hex format number
https://github.com/AcalaNetwork/Acala/blob/e97c45ea6f8fa8b58aab1b860889ed37f20a89a4/modules/dex/rpc/src/lib.rs#L19-L24

Should make it support hex format.

Substrate example:

https://github.com/paritytech/substrate/blob/270a992bf299cd41b2ed4e211814d2a958e3bfc8/primitives/runtime/src/generic/header.rs#L72-L86

Related https://github.com/polkadot-js/api/issues/2464

wangjj9219:
resolve #349

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -3345,7 +3345,9 @@ dependencies = [
  "serde",
  "sp-api",
  "sp-blockchain",
+ "sp-core",
  "sp-runtime",
+ "sp-std",
 ]
 
 [[package]]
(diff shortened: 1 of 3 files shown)


xlc:
Can you check if this works with decimal number input?
What polkadot.js does is searilze to number if it less than 53bits or hex if larger so we need to support both.

PR #356, opened 2020-08-06
Code review [WIP]

apopiak:
This PR exists to track the review of the Acala runtime for the launch.


apopiak:
## Code Review Notes
- General
    - consider going over the comments and improving the English (e.g. reduce typos, fix grammar)
    - make a decision on where to place atomicity boundaries (IMO at the boundary of any function that has more than one fallible mutation)
    - check duplicate DB accesses for performance impact
    - determine `on_finalize` weight
    - more documentation, especially for ORML/reusable stuff
- ORML
    - gradually update pallet looks like it needs a redesign
- Acala
    - use `repatriate_reserved` instead of `unreserve` and `transfer`
    - `EnabledCurrencyIds` cannot shrink currently
    - consider different Balance typedefs

modules/accounts/src/lib.rs
```diff
@@ -283,6 +302,9 @@ impl<T: Trait> Module<T> {
 			// by treasury account is sufficient to open an account.
 			if *k != treasury_account {
 				// send dust native currency to treasury account
+				// REVIEW: IIUC the account could have lots of reserved balance, just not any
+				//         free balance and thus should maybe not be considered dust just based
+				//         on the reserving failing?
 				let _ = <T as Trait>::Currency::transfer(
```
apopiak, line 308: Addressed: `open_account` is a private function that is only ever called with new accounts so they won't have reserved balance.

PR #372, opened 2020-08-14
Adjust governance

zjb0807:


runtime/dev/src/lib.rs
```diff
@@ -611,6 +632,115 @@ impl orml_auction::Trait for Runtime {
...
+	type Event = Event;
+	type Origin = Origin;
+	type PalletsOrigin = OriginCaller;
+	type Call = Call;
+	type Scheduler = Scheduler;
+	type AsOriginId = AuthoritysOriginId;
+	type AuthorityConfig = AuthorityConfigImpl;
+}
+
+pub struct AuthorityConfigImpl;
```
xlc, line 645: better to move those into a new file

PR #398, opened 2020-09-08
Add comments about ignored

### LINKED ISSUE #366: Ensure Result are not ignored without good reason
There are few places in our code uses `let _ = `

Either change them to use `.expect("reason; qed)"` or have really good explain why the results are ignored.

https://github.com/AcalaNetwork/Acala/pull/356/files#diff-92e2cb9814da46b1efd7bcd8b3868948R774

zjb0807:
Closes #366


modules/polkadot_bridge/src/lib.rs
```diff
@@ -97,6 +97,7 @@ decl_module! {
 		#[weight = 10_000]
 		pub fn simulate_withdraw_unbonded(origin) {
 			with_transaction_result(|| {
+				// ignore the result
```
xlc, line 100: ignore because we don't care who send the message

runtime/dev/src/lib.rs
```diff
@@ -138,6 +138,7 @@ where
 	GetModuleId: Get<ModuleId>,
 {
 	fn on_commission(currency_id: CurrencyId, amount: Balance) {
+		// it currently cannot fail, so ignore the result
```
xlc, line 141: this shouldn't overflow. but if it does, we will just burn the commission

modules/accounts/src/lib.rs
```diff
@@ -280,6 +280,7 @@ impl<T: Trait> Module<T> {
 			// by treasury account is sufficient to open an account.
 			if *k != treasury_account {
 				// send dust native currency to treasury account
+				// it currently cannot fail, so ignore the result
```
xlc, line 283: we transfer all free balances from a new account, so it shouldn't fail.
but even it failed, leave some dust storage is not a critical issue

xlc (approved):

PR #400, opened 2020-09-08
enable dex rpc

wangjj9219:
Still has problem to deserialize the uncertain types(hex string or integer): #349 , I modify the passing param type from generic `Balance` to `String`, use decimal string to do query.

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -68,6 +68,7 @@ dependencies = [
  "dev-runtime",
  "jsonrpc-core",
  "jsonrpc-pubsub",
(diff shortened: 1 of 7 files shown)


wangjj9219:
@qwer951123  need to update `BalanceRequest` type of SDK after merge

xlc:
Will this help? https://github.com/paritytech/substrate/blob/fd22c40755dd7657ef01915dcbbcf394a8dd0c59/primitives/rpc/src/number.rs#L35

xlc (approved):

PR #424, opened 2020-09-26
tokenize DEX lp share

wangjj9219:


modules/incentives/src/lib.rs
```diff
@@ -31,7 +34,30 @@ pub enum PoolId {
...
+}
+
+decl_event!(
+	pub enum Event<T> where
+		<T as frame_system::Trait>::AccountId,
+		Balance = Balance,
+		CurrencyId = CurrencyId,
+	{
+		/// Deposit DEX LP share. \[who, lp_currency_id, deposit_amount\]
+		DepositDEXLP(AccountId, CurrencyId, Balance),
```
xlc, line 52: `DepositDexShare` will be better?

modules/incentives/src/lib.rs
```diff
@@ -103,6 +136,46 @@ decl_module! {
...
+				Ok(())
+			})?;
+		}
+
+		#[weight = 10_000]
+		pub fn withdraw_dex_lp(origin, lp_currency_id: CurrencyId, amount: Balance) {
+			with_transaction_result(|| {
+				let who = ensure_signed(origin)?;
+
+				let share_amount: Share = amount.unique_saturated_into();
```
xlc, line 160: Do we still need a `Share` type?

xlc (approved):

PR #429, opened 2020-10-01
improve & refactor

xlc:
Closes #356

Should fix most of the issues except weights, storage RW optimizations, comments, and test coverages.


shaopengw (approved):
LGTM

PR #430, opened 2020-10-01
Add nft benchmark

zjb0807:


xlc (changes requested):
Please update to use WeightInfo

Examples:
https://github.com/paritytech/substrate/pull/7107
https://github.com/paritytech/substrate/pull/7168

From @shawntabrizi:

Each extrinsic should have its own T::WeightInfo::extrinsic_name() formula
This is generated automatically when adding --output ./path flag to the benchmark cli
You can add #[extra] to benchmarks that should not be included in the outputted file

xlc (approved):

Issue #433, opened 2020-10-04
Remove TransactionPayment from construct_runtime


xlc:
We shouldn't need it as we have module-accounts which is fork of it.

PR #436, opened 2020-10-06
Remove TransactionPayment from  construct_runtime

HZHAIX:
PR for #433

### DIFF
--- a/runtime/dev/src/lib.rs
+++ b/runtime/dev/src/lib.rs
@@ -1383,8 +1383,6 @@ construct_runtime!(
 
 		// Dev
 		Sudo: pallet_sudo::{Module, Call, Config<T>, Storage, Event<T>},
-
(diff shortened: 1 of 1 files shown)


xlc:
I wish this is easy as it is... but it is not. You will need to ensure the transaction payment RPC still works and everything still compiles.

xlc (changes requested):
I wish this is easy as it is... but it is not. You will need to ensure the transaction payment RPC still works and everything still compiles.

HZHAIX:
@xlc Is is ok to add `query_info` function in `module_account` ?

xlc:
This will make RPC not working.

This needs to be removed.
https://github.com/AcalaNetwork/Acala/blob/7135e5828a92899be03598226cd48a428ecff93d/runtime/dev/src/lib.rs#L293-L299

`query_info` needs to be implemented within `module-accounts`

PR #468, opened 2020-10-13
Refactor dex

wangjj9219:
BREAKING CHANGES!

1. Use new swap mechanism refer to Uniswap V2:
a. Support trading pairs between any tokens exclude dex share lp
b. Support use exact supply to swap target, and use supply to swap exact target
c. Can specify the swap path, for example: `[BTC, AUSD, ACA]` means swap BTC to AUSD firstly, then swap AUSD to ACA.
d. More efficient and simple calculation
2. refactor the trait `DEXManager`, affect multiple modules: cdp_treasury, cdp_engine, auction_manager, accounts
3. If the account which has not opened account receive non-native token and swap failed, no longer recycle non-native token. For the dust, it will be better to deal with it in currencies modules. 
4. remove dex rpc
5. some rename


modules/dex/src/lib.rs
```diff
@@ -316,419 +300,247 @@ decl_module! {
...
+				*pool_1 = pool_1.saturating_sub(target_decrement);
+			} else {
+				*pool_0 = pool_0.saturating_sub(target_decrement);
+				*pool_1 = pool_1.saturating_add(supply_increment);
+			}
 		});
+	}
 
-		Ok(other_currency_amount)
+	fn _swap_by_path(path: Vec<CurrencyId>, amounts: Vec<Balance>) {
```
xlc, line 447:
```suggestion
	fn _swap_by_path(path: &Vec<CurrencyId>, amounts: &Vec<Balance>) {
```

modules/dex/src/lib.rs
```diff
@@ -18,41 +17,40 @@ use orml_traits::{MultiCurrency, MultiCurrencyExtended};
...
-	/// The base currency as the core currency in all trading pairs
-	type GetBaseCurrencyId: Get<CurrencyId>;
+	/// Allowed trading pair list.
+	type EnabledTradingPairs: Get<Vec<(CurrencyId, CurrencyId)>>;
 
 	/// Trading fee rate
-	type GetExchangeFee: Get<Rate>;
+	/// The first item of the tuple is the numerator of the fee rate, second
+	/// item is the denominator, fee_rate = numerator / denominator
+	type GetExchangeFee: Get<(u32, u32)>;
```
xlc, line 50: why not using `Rate`?

wangjj9219: `target -> supply` calculation is the inverse of `supply -> target` calculation, in order to make the result of `target -> supply` can also meet `supply -> target`, we need add extra `1` after each internal `div` on `target -> supply` to correct the possible losses. However, the correction maybe make the result bigger, now we only have just 1 div on the end of `get_supply_amount`, and the final result is only 1 more at most.

These's a `div` in the `saturating_mul_int` of Rate, if use Rate, the formula also needs to be adjusted, and finally there will be 3 corrections. After 2 corrections, there are some other calculations, and the final result will have a large deviation. We have to minimize internal division operation.

xlc: Should add something to comment, like use (u32, u32) over Rate to minimize internal division operation.

xlc:
Maybe we should have new
```
struct TradingPair(p1: CurrencyId, p2: CurrencyId)

impl TradingPair {
  fn new(p1: CurrencyId, p2: CurrencyId) {
    // create with sorted order
  }
}
```

modules/dex/src/lib.rs
```diff
@@ -512,14 +539,22 @@ impl<T: Trait> DEXManager<T::AccountId, CurrencyId, Balance> for Module<T> {
 		Self::get_liquidity(currency_id_a, currency_id_b)
 	}
 
-	fn get_swap_target_amount(path: Vec<CurrencyId>, supply_amount: Balance) -> Option<Balance> {
-		Self::get_target_amounts(path, supply_amount, None)
+	fn get_swap_target_amount(
+		path: Vec<CurrencyId>,
```
xlc, line 543: make all the paths to be `&[CurrencyId]`?

xlc (approved):

PR #471, opened 2020-10-14
MultiCurrency precompile

### LINKED ISSUE #463: MultiCurrency & ERC20 integration
Bridge currencies to EVM by using custom precompiles

shaopengw:
Closes #463


modules/support/src/evm.rs
```diff
@@ -0,0 +1,98 @@
...
+use orml_traits::MultiCurrency;
+
+pub trait AccountIdConversion<AccountId> {
+	fn from_eth_address(address: &[u8; 20]) -> AccountId;
+}
+
+/// The `MultiCurrency` impl precompile.
+///
+/// The 1st byte of `input` indicates action.
+/// - `0`: Query total issuance. In this case, the 2nd byte is `currency_id`.
```
xlc, line 16: make a `#[repr(u8)] enum` instead using numbers

modules/support/src/evm.rs
```diff
@@ -0,0 +1,98 @@
...
+		let currency_id: CurrencyId = u8::from_be_bytes(currency_id_byte).into();
+
+		// Query total issuance.
+		if action == 0 {
+			let total_issuance = vec_u8_from_balance(MC::total_issuance(currency_id))?;
+			return Ok((ExitSucceed::Returned, total_issuance, 0));
+		}
+
+		// Query balance.
+		if action == 1 {
```
xlc, line 55: `match action`

modules/support/src/evm.rs
shaopengw: I can create an `Action` enum and make match more readable, but as the `intput` from evm are numbers, in the documentation it might be better to keep the numbers so that it's more clear for solidity side.

service/src/chain_spec.rs
```diff
@@ -219,6 +221,25 @@ pub fn mandala_testnet_config() -> Result<DevChainSpec, String> {
 	DevChainSpec::from_json_bytes(&include_bytes!("../../resources/mandala-dist.json")[..])
 }
 
+pub fn evm_genesis_accounts() -> BTreeMap<H160, GenesisAccount> {
+	let contracts_json = &include_bytes!("../../predeploy-contracts/resources/bytecodes.json")[..];
+	let contracts: Vec<(String, String)> = serde_json::from_slice(contracts_json).unwrap();
```
xlc, line 226:
```suggestion
	let contracts: Vec<(String, String)> = serde_json::from_slice(contracts_json).expect("predeploy contracts byte codes must be valid");
```

xlc (approved):

PR #519, opened 2020-11-09
Module EVM

ermalkaleci:


primitives/evm/src/lib.rs
```diff
@@ -0,0 +1,39 @@
+#![cfg_attr(not(feature = "std"), no_std)]
```
xlc, line 1: do we need to fork this?

if we do, I don't think this needs to be an standalone crate, can just be part of primitives create

xlc (approved):

xlc:
will get this merged now and make followup issues for remaining cleanups

primitives/evm/src/lib.rs
ermalkaleci: @xlc can move it into `module-evm`

PR #520, opened 2020-11-10
NFT Precompile

shaopengw:


modules/evm-accounts/src/lib.rs
```diff
@@ -253,6 +254,15 @@ impl<T: Trait> AddressMapping<AccountId32> for EvmAddressMapping<T> {
 	}
 }
 
+pub struct EvmAccountMapping<T>(sp_std::marker::PhantomData<T>);
+impl<T: Trait> AccountMapping<AccountId32> for EvmAccountMapping<T> {
```
xlc, line 258: can you do `where T::AccountId: From<AccountId32>` to avoid the encode / decode hack?

runtime/common/src/precompile/nft.rs
```diff
@@ -0,0 +1,134 @@
...
+impl<AccountId, AccountIdConverter, AccountMappingImpl, NFTImpl> Precompile
+	for NFTPrecompile<AccountId, AccountIdConverter, AccountMappingImpl, NFTImpl>
+where
+	AccountId: Debug + Clone,
+	AccountIdConverter: AddressMapping<AccountId>,
+	AccountMappingImpl: AccountMapping<AccountId>,
+	NFTImpl: NFT<AccountId, Balance = NFTBalance, ClassId = u64, TokenId = u64>,
+{
+	fn execute(input: &[u8], _target_gas: Option<usize>) -> result::Result<(ExitSucceed, Vec<u8>, usize), ExitError> {
+		debug::info!("----------------------------------------------------------------");
```
xlc, line 51: remove those debug logs

xlc (approved):
Should have some tests in later PR

PR #528, opened 2020-11-12
manage balance & fixes

ermalkaleci:
- Transfer the amount from caller to target/contract
- Fix `gas_used` calculation


modules/evm/src/runner/native.rs
```diff
@@ -53,14 +53,27 @@ impl<T: Trait> RunnerT<T> for Runner<T> {
...
 			Handler::<T>::new_with_precompile(&vicinity, gas_limit as usize, false, config, T::Precompiles::execute);
 
+		if !value.is_zero() {
+			substate
+				.transfer(Transfer { source, target, value })
+				.map_err(|_| Error::BalanceLow)?;
+		}
+
+		substate
+			.withdraw(source, U256::from(gas_limit))
```
xlc, line 63: we don't charge gas. the standard tx fee already handles it

ermalkaleci: I was trying to follow StackRunner. Anyway will address it later if needed

modules/evm/src/runner/native.rs
```diff
@@ -76,12 +72,16 @@ impl<T: Trait> RunnerT<T> for Runner<T> {
...
-			match reason {
-				ExitReason::Succeed(_) => TransactionOutcome::Commit(Ok(call_info)),
-				ExitReason::Revert(_) => TransactionOutcome::Rollback(Ok(call_info)),
-				ExitReason::Error(_) => TransactionOutcome::Rollback(Ok(call_info)),
-				ExitReason::Fatal(_) => TransactionOutcome::Rollback(Ok(call_info)),
+			if !reason.is_succeed() {
+				return TransactionOutcome::Rollback(Ok(call_info));
 			}
+
+			if let Err(e) = substate.transfer(Transfer { source, target, value }) {
```
xlc, line 79: why do we need this? `substate.execute` doesn't do transfer? but shouldn't we transfer first?

ermalkaleci: @xlc not it doesn't do transfer. I'm thinking about moving transfer inside `substrate.execute` to reduce complexity

modules/evm/src/runner/native.rs
```diff
@@ -268,7 +269,14 @@ impl<'vicinity, 'config, T: Trait> Handler<'vicinity, 'config, T> {
 		value: U256,
 		code: Vec<u8>,
 		input: Vec<u8>,
+		transfer: Option<Transfer>,
```
xlc, line 272: do we need additional parameter? `caller`, `address`, `value` should be `source`, `target` and `value`

Is there any case this does not do transfer? I think it should just always transfer.

xlc (commented):
This is getting complicated so should have bit more tests.

PR #530, opened 2020-11-13
check transaction priority

### LINKED ISSUE #440: Revisit transaction priorities
After #432

- [x] Check priority of unsigned transactions to see if they makes sense
- [x] Check all operational transaction to see if they should be operational

wangjj9219:
close #440 

1. `cancel_auction`, `settle_cdp` and `liquidate_cdp` could be generated in large quantities in a short time, if their dispatch class are Operational,  maybe block other business transactions.

2. All configuration update functions should be Operational so that take effect as soon as possible.

3. `settle_cdp` should have a higher priority than `cancel_auction`, so that users can retrieve their remain assets as soon as possible when emergency shutdown occurs

### DIFF
--- a/runtime/acala/src/lib.rs
+++ b/runtime/acala/src/lib.rs
@@ -27,7 +27,7 @@ use sp_runtime::{
 	curve::PiecewiseLinear,
 	generic, impl_opaque_keys,
 	traits::AccountIdConversion,
-	transaction_validity::{TransactionPriority, TransactionSource, TransactionValidity},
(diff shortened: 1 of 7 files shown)


runtime/acala/src/lib.rs
```diff
@@ -139,6 +139,14 @@ pub fn get_all_module_accounts() -> Vec<AccountId> {
 	]
 }
 
+// Priority of unsigned transactions
+parameter_types! {
```
xlc, line 143: could move those parameters to common

xlc (approved):

PR #540, opened 2020-11-16
Remove EcdsaSignature

zjb0807:


xlc (approved):

xlc:
Still keep `type EcdsaSignature = ecdsa::Signature;` to SDK types doesn't need change.

Also should update the one in ren bridge as well

https://github.com/AcalaNetwork/Acala/blob/master/ecosystem-modules/ren/renvm-bridge/src/lib.rs#L21

xlc (approved):

PR #544, opened 2020-11-18
Add context to precompile call.

shaopengw:


runtime/mandala/src/lib.rs
```diff
@@ -1244,13 +1246,28 @@ impl pallet_contracts::Trait for Runtime {
...
+			s.insert(H160::from_low_u64_be(PREDEPLOY_ADDRESS_START + i));
+		}
+		s
+	};
+}
+
+pub struct MultiCurrencyPrecompileCallerFilter;
+impl PrecompileCallerFilter for MultiCurrencyPrecompileCallerFilter {
+	fn is_allowed(caller: H160) -> bool {
+		MultiCurrencyPrecompileAllowedCallers::get().contains(&caller)
```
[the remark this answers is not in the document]
shaopengw: Yes, if that's safe enough for us.

runtime/common/src/precompile/mod.rs
```diff
@@ -33,12 +33,13 @@ where
 		address: H160,
 		input: &[u8],
 		target_gas: Option<usize>,
+		context: &Context,
 	) -> Option<core::result::Result<(ExitSucceed, Vec<u8>, usize), ExitError>> {
-		EthereumPrecompiles::execute(address, input, target_gas).or_else(|| {
+		EthereumPrecompiles::execute(address, input, target_gas, context).or_else(|| {
 			if address == H160::from_low_u64_be(PRECOMPILE_ADDRESS_START) {
```
xlc, line 39: maybe move the filter check to here?

xlc (approved):

runtime/common/src/precompile/mod.rs
shaopengw: We might need custom configurable allowed list for other precompiles, to support non-predeployed contracts. Like the NFT contract for linkdrop.

runtime/common/src/lib.rs
```diff
@@ -257,9 +257,32 @@ parameter_types! {
 	};
 }
 
-pub struct MultiCurrencyPrecompileCallerFilter;
-impl PrecompileCallerFilter for MultiCurrencyPrecompileCallerFilter {
+parameter_types! {
+	pub const ZeroBytesLength: usize = 12;
```
xlc, line 261: a normal const should be enough?

shaopengw: emmm, yep, will update

PR #550, opened 2020-11-18
Only apply precompile filter to system contracts.

shaopengw:


xlc (approved):
Should have some tests

runtime/common/src/lib.rs
```diff
@@ -247,11 +247,18 @@ impl<Balance: FixedPointOperand> module_staking_pool::FeeModel<Balance> for Curv
 
 pub const ZERO_BYTES_LEN: usize = 12;
```
xlc, line 248: move this into `is_system_contract` to avoid confusion

PR #554, opened 2020-11-19
Update how transfer is handled on dispatch methods

ermalkaleci:


modules/evm/src/runner/native.rs
```diff
@@ -56,9 +56,12 @@ impl<T: Trait> RunnerT<T> for Runner<T> {
 		substate.inc_nonce(source);
 
 		frame_support::storage::with_transaction(|| {
-			let transfer = Some(Transfer { source, target, value });
+			if substate.transfer(Transfer { source, target, value }).is_err() {
+				return TransactionOutcome::Rollback(Err(Error::BalanceLow));
```
xlc, line 60: shouldn't this propagate the error? it should be out of gas or some other errors

ermalkaleci: @xlc correct. This PR was to move transfer out of `substrate.execute` and handle them differently on Create/Call dispatch calls

ermalkaleci: it should fail to dispatch tx if transfer fails

PR #559, opened 2020-11-20
EVM: native runner cleanup

shaopengw:


modules/evm/src/runner/native.rs
```diff
@@ -364,6 +319,29 @@ impl<'vicinity, 'config, T: Trait> Handler<'vicinity, 'config, T> {
...
 
+/// Create `try_or_fail` and `try_or_rollback`.
+macro_rules! create_try {
+	( $map_err:expr ) => {
+		macro_rules! try_or_fail {
+			( $e:expr ) => {
+				match $e {
+					Ok(v) => v,
+					Err(e) => return Capture::Exit($map_err(e)),
+					}
```
xlc, line 330: indent is wrong

modules/evm/src/runner/native.rs
```diff
@@ -364,6 +319,29 @@ impl<'vicinity, 'config, T: Trait> Handler<'vicinity, 'config, T> {
 	}
 }
 
+/// Create `try_or_fail` and `try_or_rollback`.
+macro_rules! create_try {
+	( $map_err:expr ) => {
+		macro_rules! try_or_fail {
+			( $e:expr ) => {
```
xlc, line 326: why this needs to be macro? can we make it generic function?

shaopengw: Generic function cannot cover the requirement that "if err, return a non err value, else do nothing". Usually `try!` and `?` syntax could handle this scenario but the handler functions doesn't return `Result`.

modules/evm/src/runner/native.rs
shaopengw: I did it right, and then cargo fmt made it wrong.😂

modules/evm/src/runner/native.rs
xlc: ok. we can refactor the code so it runs in a method return `Result` later

xlc (approved):

PR #565, opened 2020-11-21
gas to weight converter

ermalkaleci:


modules/evm/src/lib.rs
```diff
@@ -214,30 +219,25 @@ decl_module! {
...
+			} else {
+				Module::<T>::deposit_event(Event::<T>::ExecutedFailed(target, info.exit_reason, info.output));
 			}
 
-			Ok(Pays::No.into())
+			let used_gas: u32 = info.used_gas.unique_saturated_into();
+
+			Ok(PostDispatchInfo {
+				actual_weight: Some(T::GasToWeight::convert(used_gas)),
+				pays_fee: Pays::No
```
xlc, line 234:
```suggestion
				pays_fee: Pays::Yes
```

We want to use our custom fee handling logic to handle the tx fee instead of evm gas fee

xlc (changes requested):

Issue #525, opened 2020-11-12
Remove EnsureAddressOrigin and refactor AddressMapping


xlc:
- Remove EnsureAddressOrigin
- Move AddressMapping to primitives
- Merge AccountMapping into AddressMapping
  - fn to_account(evm: &EvmAddress) -> T::AccountId
  - fn to_evm_address(account: &T::Account) -> Option\<EvmAddress>

PR #569, opened 2020-11-23
Refactor address mapping

zjb0807:
Closes: #525


modules/evm/src/lib.rs
```diff
@@ -29,59 +29,25 @@ use sp_std::{marker::PhantomData, vec::Vec};
...
-pub struct IdentityAddressMapping;
-
-impl AddressMapping<H160> for IdentityAddressMapping {
-	fn into_account_id(address: H160) -> H160 {
-		address
-	}
-}
-
 /// Hashed address mapping.
 pub struct HashedAddressMapping<H>(sp_std::marker::PhantomData<H>);
```
xlc, line 33: we don't use this. so can just move this into test

xlc (approved):

Issue #499, opened 2020-10-30
EVM: Contract Existential Deposit


xlc:
- module-evm::Trait
  - type ContractExistentialDeposit: Get<Balance>
    - default to 0 for mandala so we don't break existing Solidity projects
    - make it 1 ACA for mock so we can test this
- primitives::evm::Vicinity
  - creating: bool
    - set to true when calling runner create / create2

When new contract is created, if is `vicinity.creating` is true:
  - transfer `ContractExistentialDeposit` from `vicinity.origin` to contract address
  - reserve this amount

if `vicinity.creating` is false:
  - transfer `ContractExistentialDeposit` from `caller` to contract address
  - reserve this amount

When contract is deleted, unreserve `ContractExistentialDeposit` before calling merge account

----

This means when deploying a contract, the `tx.origin` will pay for the deposit for the deployed contract and all other contracts deployed in constructor

When interacting a contract, the factory contract is responsible for deposit. It is up to the contract to decide if it should require user to pay for the deposit or cover the deposit.

PR #577, opened 2020-11-25
Add ContractExistentialDeposit

zjb0807:
Closes: #499 

`vicinity.origin` always is `tx.caller`, so don't need `vicinity.creating`

### DIFF
--- a/modules/evm/src/runner/handler.rs
+++ b/modules/evm/src/runner/handler.rs
@@ -1,7 +1,8 @@
 #![allow(clippy::type_complexity)]
 
 use crate::{
-	AccountCodes, AccountNonces, AccountStorages, AddressMapping, Event, Log, MergeAccount, Module, Trait, Vicinity,
+	AccountCodes, AccountNonces, AccountStorages, AddressMapping, BalanceOf, Event, Log, MergeAccount, Module, Trait,
+	Vicinity,
 };
 use evm::{
 	Capture, Context, CreateScheme, ExitError, ExitReason, ExitSucceed, ExternalOpcode, Opcode, Runtime, Stack,
@@ -12,12 +13,12 @@ use evm_runtime::{Config, Handler as HandlerT};
 use frame_support::{
 	debug,
 	storage::{StorageDoubleMap, StorageMap},
-	traits::{Currency, ExistenceRequirement, Get},
+	traits::{Currency, ExistenceRequirement, Get, ReservableCurrency},
 };
 use sha3::{Digest, Keccak256};
 use sp_core::{H160, H256, U256};
 use sp_runtime::{
-	traits::{One, UniqueSaturatedInto},
+	traits::{One, UniqueSaturatedInto, Zero},
 	SaturatedConversion, TransactionOutcome,
 };
 use sp_std::{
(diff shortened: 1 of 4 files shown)


modules/evm/src/runner/handler.rs
```diff
@@ -368,11 +368,21 @@ impl<'vicinity, 'config, T: Trait> HandlerT for Handler<'vicinity, 'config, T> {
 				value,
 			}));
 
-			try_or_rollback!(self.transfer(Transfer {
-				source: self.vicinity.origin,
-				target: address,
-				value: U256::from(T::ContractExistentialDeposit::get().saturated_into::<u128>())
-			}));
+			let contract_existential_deposit = if self.vicinity.creating {
```
xlc, line 371:
```
let contract_existential_deposit = 
				Transfer {
					source: if self.vicinity.creating { self.vicinity.origin } else { caller }
					target: address,
					value: U256::from(T::ContractExistentialDeposit::get().saturated_into::<u128>()),
				};
```

xlc (approved):

modules/evm/src/tests.rs
```diff
@@ -491,12 +511,66 @@ fn should_transfer_from_contract() {
...
+		let result =
+			<Test as Trait>::Runner::call(alice(), result.address, create_contract, amount, 1000000000).unwrap();
+		assert_eq!(result.exit_reason, ExitReason::Succeed(ExitSucceed::Stopped));
+		assert_eq!(
+			balance(alice()),
+			INITIAL_BALANCE - amount - <Test as Trait>::ContractExistentialDeposit::get()
+		);
+		assert_eq!(
+			balance(factory_contract_address),
+			amount - <Test as Trait>::ContractExistentialDeposit::get()
```
ermalkaleci, line 562: @xlc @zjb0807 This isn't making sense to me. `factory_contract_address` balance should `amount`. Alice is the one who should pay the deposit.

modules/evm/src/tests.rs
```diff
@@ -513,11 +587,21 @@ fn contract_should_deploy_contracts() {
...
+			balance(alice()),
+			INITIAL_BALANCE - <Test as Trait>::ContractExistentialDeposit::get()
+		);
+		assert_eq!(
+			reserved_balance(result.address),
+			<Test as Trait>::ContractExistentialDeposit::get()
+		);
 
 		// Factory.createContract
+		// need factory contract pay for the ContractExistentialDeposit. But factory not
```
ermalkaleci, line 600: @xlc @zjb0807 I don't see why factory should be payable. Alice is willing to create the contract so she should pay the deposit

modules/evm/src/tests.rs
xlc: For indirect contract creation via call instead create / create2, it is the contract paying the deposit, not the caller. Contract may require caller to send token for the contract deposit but it can also cover the fee for user.

modules/evm/src/tests.rs
xlc: non payable method means caller cannot send ACA and call the method. and we should not charge extra token from caller.

modules/evm/src/tests.rs
ermalkaleci: In that case `alice` balance should be `INITIAL_BALANCE - amount`, right?

PR #580, opened 2020-11-25
Simplify EVM handler

xlc:
The deleted BTreeSet is a nice to have optimization to avoid unnecessary storage changes on revert, but for now correctness is more important than performance.


shaopengw (approved):
LGTM

Issue #566, opened 2020-11-22
Update account existence


xlc:
Depends on: 
  - https://github.com/open-web3-stack/open-runtime-module-library/issues/324
  - https://github.com/paritytech/substrate/pull/7363

Remove open account deposit & related code
Update all the pallets to update consumer / provider count 
Update integration test to remove irrelevant ones and add new one

wangjj9219:
closed by #724

PR #588, opened 2020-11-29
refactor module-accounts

wangjj9219:
Partially completed #566 

1. Remove open account deposit & related code of `module_accounts`
2. Config `System` instead of `Accounts` as the `AccountStore` for `pallet_balances`
3. rename `module_accounts` to `module_transaction_payment`
4. redefine some hooks of `module_evm_accounts` and `module_evm`


modules/accounts/src/lib.rs
```diff
@@ -384,103 +336,6 @@ decl_module! {
...
-				<T as Trait>::MultiCurrency::free_balance(currency_id, who),
-				Some(T::MaxSlippageSwapWithDEX::get()),
-			);
-		}
-	}
-}
-
 /// Fork StoredMap in frame_system,  still use `Account` storage of
 /// frame_system.
 impl<T: Trait> StoredMap<T::AccountId, T::AccountData> for Module<T> {
```
xlc, line 341: I don't think we need this anymore?

runtime/karura/src/lib.rs
```diff
@@ -742,13 +739,34 @@ impl DataFeeder<CurrencyId, Price, AccountId> for AggregatedDataProvider {
 	}
 }
 
+pub struct NonNativeTokenOnDust;
+impl OnDust<AccountId, CurrencyId, Balance> for NonNativeTokenOnDust {
```
xlc, line 743: I think we can move this to orml-tokens?

```
struct TransferDust<TreasuryAccount: Get<AccountId>, Currency: MultiCurrency>()
```

xlc (commented):
should use a non-zero existential deposit for tests

xlc (approved):

PR #592, opened 2020-11-30
EVM: deploy system contract

### LINKED ISSUE #546: EVM: deploy system contract
For security reason we introduced system contract filter, and only system contracts are allowed to call precompile contracts. A system contract's address starts with at least 12 zero bytes.

To support non-predeployed contract as system contract, EVM needs deploy system contracts dispatch call, which will deploy the contract with next available system contract id as address.

shaopengw:
Closes #546


xlc (approved):

ermalkaleci:
What happens if you provide an existing address? Will it fail due to collision or overwrite it?

xlc:
This should be able to deploy to an existing address as it updates network_contract_index every time.

PR #596, opened 2020-11-30
update revert message decoding

ermalkaleci:


modules/evm/rpc/src/lib.rs
```diff
@@ -39,15 +40,19 @@ fn error_on_execution_failure(reason: &ExitReason, data: &[u8]) -> Result<()> {
...
-				.into_iter()
-				.filter(|x| !x.is_ascii_control())
-				.collect();
-			let message = String::from_utf8_lossy(&msg_data).trim().to_owned();
+			let mut message = "VM Exception while processing transaction: revert".to_string();
+			// A minimum size of error function selector (4) + offset (32) + string length
+			// (32) should contain a utf-8 encoded revert reason.
+			if data.len() > 68 {
+				let message_len = U256::from(&data[36..68]).saturated_into::<usize>();
+				let body: &[u8] = &data[68..68 + message_len];
```
xlc, line 48: add length check to make sure this can't index out of bounds

xlc (changes requested):

modules/evm/rpc/src/lib.rs
```diff
@@ -64,6 +52,23 @@ fn error_on_execution_failure(reason: &ExitReason, data: &[u8]) -> Result<()> {
 	}
 }
 
+fn decode_revert_message(data: &[u8]) -> String {
+	let invalid: String = "invalid revert message".into();
+	// A minimum size of error function selector (4) + offset (32) + string length
+	// (32) should contain a utf-8 encoded revert reason.
+	if data.len() > 68 {
+		let message_len = U256::from(&data[36..68]).saturated_into::<usize>();
+		if data.len() < message_len {
```
xlc, line 61: should be `if data.len() < message_len + 68 {`?
and can you add some tests to `decode_revert_message` using data you get from evm

PR #604, opened 2020-12-02
Dex improvements

### LINKED ISSUE #568: DEX improvements
Goal:
- Able to add new pair via governance call
- Improved initial price discovery
  - Currently the first liquidity provider decide the initial price, which is bad

module-dex

- Types
  - struct TradingPairProvisionParameters
    - min_contribution: (Balance, Balance)
    - target: (Balance, Balance)
    - not_before: BlockNumber
  - enum TradingPairStatus
    - NotEnabled
    - Provisioning(TradingPairProvisionParameters)
    - Enabled
- Trait
  - type ListingOrigin: EnsureOrigin
  - type Scheduler: ScheduleNamed
- Storages
  - TradingPairStatuses: map TradingPair => TradingPairStatus
  - ProvisioningPool: double_map TradingPair, AccountId => (Balance, Balance)
- Call
  - fn enable_pair(origin, pair: TradingPair, parameters: TradingPairProvisionParameters)
  - fn disable_pair(origin, pair: TradingPair)
    - only allow LP to withdraw liquidity

Notes:
- Every time TradingPairStatuses is accessed, check if should convert it from `Provisioning` to `Enabled`
- When enabling, use the total pool size to determine the initial price
- Everyone in the ProvidioningPool getting their share based on their asset value using initial price for calculation
- Example: https://docs.google.com/spreadsheets/d/1T9Hd14TT0hwtBZ6CE1y3NpQH0IADYIvCYCys_V17lm0/edit?usp=sharing

wangjj9219:
resolve #568


modules/dex/src/lib.rs
```diff
@@ -215,6 +385,122 @@ impl<T: Trait> Module<T> {
 		T::ModuleId::get().into_account()
 	}
 
+	/// Access status of specific trading_pair,
+	/// if status is Provisioning and able to be `Enabled`, update it and return
+	/// `Enabled`
+	fn access_trading_pair_status(trading_pair: TradingPair) -> TradingPairStatus<Balance, T::BlockNumber> {
```
xlc, line 391: we should do this check after add liquidity not before add liquidity

modules/dex/src/lib.rs
```diff
@@ -215,6 +385,122 @@ impl<T: Trait> Module<T> {
...
+		let trading_pair_status = Self::trading_pair_statuses(trading_pair);
+		match trading_pair_status {
+			TradingPairStatus::<_, _>::Provisioning(provision_parameters) => {
+				// convert Provisioning to Enabled if possible
+				if frame_system::Module::<T>::block_number() >= provision_parameters.not_before
+					&& !provision_parameters.accumulated_provision.0.is_zero()
+					&& !provision_parameters.accumulated_provision.1.is_zero()
+					&& provision_parameters.accumulated_provision.0 >= provision_parameters.target_provision.0
```
xlc, line 399: one of the target reach is enough

modules/dex/src/lib.rs
```diff
@@ -215,6 +385,122 @@ impl<T: Trait> Module<T> {
...
+			_ => return Err(Error::<T>::MustProvisioningTradingPair.into()),
+		};
+		let (contribution_0, contribution_1) = if currency_id_a == trading_pair.0 {
+			(contribution_a, contribution_b)
+		} else {
+			(contribution_b, contribution_a)
+		};
+
+		ensure!(
+			contribution_0 >= provision_parameters.min_contribution.0
```
xlc, line 459: one of the pair above limit is enough. so people can only add aUSD which is basically market buy order

primitives/src/lib.rs
```diff
@@ -107,6 +107,40 @@ pub enum CurrencyId {
 	DEXShare(TokenSymbol, TokenSymbol),
 }
 
+impl CurrencyId {
+	pub fn is_token_currency_id(&self) -> bool {
+		match self {
```
xlc, line 112: `matches!(self, CurrencyId::Token(_))`

https://doc.rust-lang.org/core/macro.matches.html

xlc (changes requested):

xlc (approved):

PR #625, opened 2020-12-11
module_currencies which supports ERC20 currency

ermalkaleci:


ecosystem-modules/ren/renvm-bridge/src/mock.rs
```diff
@@ -4,7 +4,7 @@
 
 use super::*;
 use frame_support::{impl_outer_event, impl_outer_origin, parameter_types};
-use orml_currencies::BasicCurrencyAdapter;
+use module_currencies::BasicCurrencyAdapter;
```
xlc, line 7: no need to replace those in mocks unless you are explicitly testing if this pallet works with module-currencies

modules/currencies/src/lib.rs
```diff
@@ -0,0 +1,800 @@
...
+pub trait Config: frame_system::Config {
+	type Event: From<Event<Self>> + Into<<Self as frame_system::Config>::Event>;
+	type MultiCurrency: MergeAccount<Self::AccountId>
+		+ MultiCurrencyExtended<Self::AccountId, CurrencyId = CurrencyId>
+		+ MultiLockableCurrency<Self::AccountId, CurrencyId = CurrencyId>
+		+ MultiReservableCurrency<Self::AccountId, CurrencyId = CurrencyId>;
+	type NativeCurrency: BasicCurrencyExtended<Self::AccountId, Balance = BalanceOf<Self>, Amount = AmountOf<Self>>
+		+ BasicLockableCurrency<Self::AccountId, Balance = BalanceOf<Self>>
+		+ BasicReservableCurrency<Self::AccountId, Balance = BalanceOf<Self>>;
+	type GetNativeCurrencyId: Get<CurrencyIdOf<Self>>;
```
xlc, line 63: no need to have this

modules/currencies/src/lib.rs
```diff
@@ -0,0 +1,770 @@
...
+		Ok(())
+	}
+}
+
+impl<T: Config> MultiCurrency<T::AccountId> for Module<T> {
+	type CurrencyId = CurrencyIdOf<T>;
+	type Balance = BalanceOf<T>;
+
+	fn minimum_balance(currency_id: Self::CurrencyId) -> Self::Balance {
+		if currency_id.is_erc20() {
```
xlc, line 215: all those should just be a `match`

modules/currencies/src/lib.rs
```diff
@@ -0,0 +1,770 @@
...
+		to: &T::AccountId,
+		amount: Self::Balance,
+	) -> DispatchResult {
+		if amount.is_zero() || from == to {
+			return Ok(());
+		}
+
+		if let CurrencyId::ERC20(contract) = currency_id {
+			let source = T::AddressMapping::to_evm_address(&from).ok_or(Error::<T>::AccountNotFound)?;
+			let address = T::AddressMapping::to_evm_address(&to).ok_or(Error::<T>::AccountNotFound)?;
```
xlc, line 304: this means we can't have treasury or other system account to hold ERC20 tokens... let me think how can we improve this...

modules/currencies/src/lib.rs
```diff
@@ -0,0 +1,770 @@
...
+		}
+		Self::deposit_event(RawEvent::BalanceUpdated(currency_id, who.clone(), by_amount));
+		Ok(())
+	}
+}
+
+impl<T: Config> MultiLockableCurrency<T::AccountId> for Module<T> {
+	type Moment = T::BlockNumber;
+
+	fn set_lock(lock_id: LockIdentifier, currency_id: Self::CurrencyId, who: &T::AccountId, amount: Self::Balance) {
```
xlc, line 383: I think we will need to modify `MultiLockableCurrency` to have those return `DispatchResult` so we don't ignore error here

modules/currencies/src/lib.rs
```diff
@@ -0,0 +1,770 @@
...
+			if currency_id == T::GetNativeCurrencyId::get() {
+				T::NativeCurrency::remove_lock(lock_id, who);
+			} else {
+				T::MultiCurrency::remove_lock(lock_id, currency_id, who);
+			}
+		}
+	}
+}
+
+impl<T: Config> MultiReservableCurrency<T::AccountId> for Module<T> {
```
xlc, line 414: I think we should implement reserve for erc20 by transfer fund to a system account

xlc (changes requested):

modules/currencies/src/lib.rs
```diff
@@ -0,0 +1,770 @@
...
+			value,
+			status,
+		)
+	}
+}
+
+pub type NativeCurrencyOf<T> = Currency<T, <T as Config>::GetNativeCurrencyId>;
+
+/// Adapt other currency traits implementation to `BasicCurrency`.
+pub struct BasicCurrencyAdapter<T, Currency, Amount, Moment>(marker::PhantomData<(T, Currency, Amount, Moment)>);
```
shaopengw, line 606: The adapter and implementations is not this module specific and duplicated with orml-currencies.

modules/currencies/src/lib.rs
wangjj9219: return error when currency id is ERC20?

xlc: yes. that's better than ignore error.

modules/currencies/src/lib.rs
xlc: https://github.com/AcalaNetwork/Acala/pull/635 should help this

modules/currencies/src/lib.rs
ermalkaleci: @xlc @shaopengw the adapter depends on `orml_currencies::Config`. I suggest we merge this PR and address this on another PR since a lot of changes are required on `orml` to make it generic enough

Issue #601, opened 2020-12-02
CI: check with `with-ethereum-compatibility` feature


xlc:
Add a step to check with `with-ethereum-compatibility` feature enabled.

PR #627, opened 2020-12-14
Add ci test with-ethereum-compatibility

zjb0807:
Closes: #601


xlc (commented):
so this only tests if constructor run. I don't think we need too many of the files. just have Address, Counters, TokenMock should be enough.

xlc (approved):

PR #635, opened 2020-12-16
Update evm account

xlc:
so every Substrate account also have a default evm account as well.


modules/evm-accounts/src/lib.rs
```diff
@@ -207,22 +193,50 @@ where
...
 			}
 		})
 	}
+
+	fn get_or_create_evm_address(account_id: &T::AccountId) -> EvmAddress {
+		Self::get_evm_address(account_id).unwrap_or_else(|| {
+			let addr = account_to_default_evm_address(account_id);
+
+			// create reverse mapping
+			Accounts::<T>::insert(&addr, account_id);
```
zjb0807, line 212:
```
EvmAddresses::<T>::insert(account_id, addr);
```

zjb0807 (approved):

PR #640, opened 2020-12-18
Rococo V1

shaunxw:


bin/acala/service/src/chain_spec/mandala.rs
```diff
@@ -108,8 +108,8 @@ pub fn latest_mandala_testnet_config() -> Result<ChainSpec, String> {
 	let wasm_binary = mandala_runtime::WASM_BINARY.ok_or("Mandala runtime wasm binary not available")?;
 
 	Ok(ChainSpec::from_genesis(
-		"Acala Mandala TC5",
-		"mandala5",
+		"Acala Mandala PC1",
```
xlc, line 111: we should have two of those, one for TC5, and one for PC2

shaopengw: Could we use `acala-dev` for TC5?

xlc: oh you are right.

PR #642, opened 2020-12-18
EVM name consistency

ermalkaleci:


xlc (changes requested):
I will prefer `EvmAddress` over `EVMAddress`

For example, in other crates, it is `Url` instead of `URL`, `Http` instead of `HTTP`
https://docs.rs/url/2.2.0/url/struct.UrlQuery.html

ermalkaleci:
@xlc I prefer uppercase. Anyway in that case I suggest we change everything to `Evm`

xlc:
We just stick to whatever naming convention the standard library is using, usually by looking how they name HTTP / URL related code.

Issue #590, opened 2020-11-30
EVM: set_code


xlc:
Depends on https://github.com/AcalaNetwork/Acala/issues/533

module-evm

- Types
  - ContractInfo
    - is_public: bool
- Traits
  - type MaxCodeSize: Get\<u32>
    - make it 60 * 1024 for acala and 24576 when `with-ethereum-compatibility` is enabled
- Calls
  - fn set_code(origin, contract: H160, code: Vec\<u8>)
    - ensure maintainer or root
    - ensure contract is not public if not root
    - check max code size
    - update contract code
    - make sure ref count is updated accordingly
    - make sure storage quota is updated accordingly
  - fn selfdestruct(origin, contract: H160)
    - ensure maintainer
    - ensure contract is not public
    - delete contract & storage & refund to maintainer
    - make sure ref count is updated accordingly
    - make sure storage quota is updated accordingly
- Module
  - when deploy contract
    - check max code size or fail with out of gas error
    - https://github.com/ethereum/EIPs/blob/master/EIPS/eip-170.md
    - https://github.com/rust-blockchain/evm/blob/bc92b4408a949938afc96248152e19eced0f7c8d/src/executor/stack.rs#L529

PR #651, opened 2021-01-05
Add set code

zjb0807:
Closes: #590


modules/evm/src/lib.rs
```diff
@@ -313,6 +324,10 @@ decl_event! {
 		ContractDevelopmentDisabled(AccountId),
 		/// Deployed contract. \[contract\]
 		ContractDeployed(EvmAddress),
+		/// Seted contract code. \[contract\]
+		ContractSetedCode(EvmAddress),
```
xlc, line 328:
```suggestion
		ContractSetCode(EvmAddress),
```

There is no `Seted`, just `Set`

modules/evm/src/lib.rs
```diff
@@ -156,6 +166,7 @@ pub struct ContractInfo<T: Config> {
 	pub existential_deposit: BalanceOf<T>,
 	pub maintainer: EvmAddress,
 	pub deployed: bool,
+	pub is_public: bool,
```
xlc, line 169: should just use `deployed`, no need to have another `is_public`

xlc (changes requested):

xlc (approved):

Issue #654, opened 2021-01-05
EVM state rent redesign


xlc:
The previous state rent design is too complicated in terms of implementation as well as developer experience (hard for contract developer to use it). So we want something simpler instead.

- Revert most of the previous state rent changes.
- No more default quota.
- Add `storage_limit: u32` to `call` / `create` / `create2`.
- On new contract storage add, reserve `StorageDepositPerByte` from the origin account. i.e. the one paying the gas fee
  - Similar thing when storage is removed, refund the deposit to the account sending transaction

PR #659, opened 2021-01-07
Change state rent

zjb0807:
Closes: #654


modules/evm/src/lib.rs
```diff
@@ -264,7 +247,7 @@ decl_storage! {
 				);
 
 				if !account.code.is_empty() { // if code len > 0 then it's a contract
-					<Module<T>>::on_contract_initialization(address, &EvmAddress::default(), account.code.clone(), Some(account.storage.len() as u32)).expect("Genesis contract shouldn't fail");
+					<Module<T>>::on_contract_initialization(address, None, &EvmAddress::default(),  account.code.clone(), 100000, Some(account.storage.len() as u32)).expect("Genesis contract shouldn't fail");
```
xlc, line 250:
```suggestion
					<Module<T>>::on_contract_initialization(address, None, &EvmAddress::default(), account.code.clone(), 100000, Some(account.storage.len() as u32)).expect("Genesis contract shouldn't fail");
```

modules/evm/src/lib.rs
```diff
@@ -1025,42 +875,27 @@ impl<T: Config> Module<T> {
...
+			if delta > storage_limit {
+				return Err(Error::<T>::StorageExceedsStorageLimit.into());
+			}
 
-					if *maintainer_storage_usage > *maintainer_storage_quota {
-						return Err(Error::<T>::StorageQuotaNotEnough.into());
-					}
+			let reserve_amount = T::StorageDepositPerByte::get().saturating_mul(delta.into());
+			let origin_account = T::AddressMapping::get_account_id(&origin);
+			T::Currency::reserve(&origin_account, reserve_amount)?;
```
xlc, line 891: the reserve & unreserve may happen on different accounts. e.g. one account store 100 bytes and another account remove those bytes

so need to transfer the deposit from origin to contract, and reserve on the contract account.

on refund, use repatriate_reserved to unreserve and send refund to origin. ignore the error if no enough fund to unreserve

modules/evm/src/lib.rs
```diff
@@ -876,6 +884,7 @@ impl<T: Config> Module<T> {
 	}
 
 	fn do_update_storage_usage(
```
xlc, line 886: add `#[require_transactional]` to make sure this can only be called within transactional because it now have return error after state change.

modules/evm/src/runner/handler.rs
```diff
@@ -70,6 +73,10 @@ impl<'vicinity, 'config, T: Config> Handler<'vicinity, 'config, T> {
 			)
 	}
 
+	pub fn used_storage(&self) -> usize {
+		self.storagemeter.total_used_storage() - self.storagemeter.refunded_storage()
```
xlc, line 77: it is possible refunded is more than used

modules/evm/src/runner/handler.rs
```diff
@@ -605,3 +585,78 @@ impl<'vicinity, 'config, T: Config> HandlerT for Handler<'vicinity, 'config, T>
 		Ok(())
 	}
 }
+
+/// Storagemeter.
+#[derive(Clone)]
+pub struct Storagemeter {
```
xlc, line 591: move this to a new file `storage_meter.rs` and have some tests in that file

modules/evm/src/runner/handler.rs
```diff
@@ -605,3 +585,78 @@ impl<'vicinity, 'config, T: Config> HandlerT for Handler<'vicinity, 'config, T>
...
+	storage_limit: usize,
```
xlc, line 592: use `u32` everywhere to keep things simple

xlc (commented):
Looks good so far. I will do a more detailed review later.

xlc:
There are still few things can be improved but I think it is good enough for this to be merged now. I will do some refactor later.

xlc (approved):

Issue #664, opened 2021-01-08
EVM: New tests to very the behavior when calling a non-exist contract


xlc:
Add unit test for this to ensure the behavior https://github.com/AcalaNetwork/Acala/pull/662/commits/b62ce366bd45b4feb20fe66fbd62abd03cdb887b

PR #672, opened 2021-01-13
Fix call non-exist contracts

zjb0807:
Closes: #664


xlc (approved):

xlc:
I merged this but actually I think this needs more changes so reverted.

PR #674, opened 2021-01-14
Fix call non-exist contracts

xlc:
Reverts AcalaNetwork/Acala#673

#672

### DIFF
--- a/modules/evm/src/runner/handler.rs
+++ b/modules/evm/src/runner/handler.rs
@@ -178,15 +178,15 @@ impl<'vicinity, 'config, T: Config> Handler<'vicinity, 'config, T> {
 	}
 
 	// is contract && not deployed
-	pub fn is_undeployed_contract(&self, address: &H160) -> Result<bool, ExitError> {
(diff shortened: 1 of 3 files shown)


modules/evm/src/runner/handler.rs
```diff
@@ -178,15 +178,15 @@ impl<'vicinity, 'config, T: Config> Handler<'vicinity, 'config, T> {
...
 		if let Some(AccountInfo {
 			contract_info: Some(ContractInfo { deployed, .. }),
 			..
 		}) = Accounts::<T>::get(address)
 		{
-			!deployed
+			Ok(!deployed)
 		} else {
-			false
+			Err(ExitError::Other("Contract not exists".into()))
```
xlc, line 189: We shouldn't need to have a custom contract not exist error. Should still let evm try to execute it and let evm to handle the error instead ourself

xlc (commented):
Can you add new test that calls non-exist contract and assert the result

xlc:
Can't approve my own PR but this looks good to me

zjb0807 (approved):

Issue #655, opened 2021-01-05
EVM bridge update


xlc:
Depends on #654

- Types
  - enum ExecutionMode
    - Execute
    - View
      - Discard any state changes
    - EstimateGas
      - Also discard any state changes and use estimate gas mode for evm config
  - trait EVM
    - fn execute
      - context: InvokeContext
      - input: Vec\<u8>
      - value: Balance
      - gas_limit: u32
      - storage_limit: u32
      - mode: ExecutionMode
      - Note: the storage rent should be paid by / refunded to `context.origin`
  - struct InvokeContext
    - contract: EvmAddress
    - sender: EvmAddress
      - rename from `source`, similar to  `msg.sender`
    - origin: EvmAddress
      - similar to `tx.origin`
- EVMBridge
  - make `total_supply` & `balance_of` use `ExecutionMode::View`
    - make sure document is updated

PR #676, opened 2021-01-14
Update evm-bridge

zjb0807:
Closes: #655

### DIFF
--- a/modules/evm/src/lib.rs
+++ b/modules/evm/src/lib.rs
@@ -38,7 +38,7 @@ use sp_runtime::{
 	Either,
 };
 use sp_std::{marker::PhantomData, vec::Vec};
-use support::{EVMStateRentTrait, EVM as EVMTrait};
+use support::{EVMStateRentTrait, ExecutionMode, InvokeContext, EVM as EVMTrait};
 
 /// Type alias for currency balance.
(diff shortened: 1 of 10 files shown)


modules/evm/src/lib.rs
```diff
@@ -837,29 +837,33 @@ impl<T: Config> EVMTrait for Module<T> {
...
 		storage_limit: u32,
-		config: Option<evm::Config>,
+		mode: ExecutionMode,
 	) -> Result<CallInfo, sp_runtime::DispatchError> {
+		let mut config = T::config().clone();
+		if let ExecutionMode::EstimateGas = mode {
+			config.estimate = true;
+		}
+
 		let info = Runner::<T>::call(
```
xlc, line 852: if mode is view:
- don't emit event
- put the call in `with_transaction` and rollback the changes

modules/evm/src/lib.rs
```diff
@@ -837,29 +837,33 @@ impl<T: Config> EVMTrait for Module<T> {
...
-			source,
-			target,
+			context.origin,
```
xlc, line 853: should add new parameter to `Runner::<T>::call` to take both the sender & origin. sender should remain the same but vicinity.origin should be origin

xlc (changes requested):

modules/evm/src/lib.rs
xlc: make sure you have tests

xlc (approved):

Issue #677, opened 2021-01-14
EVM: Signed extension so evm-bridge can get real tx origin


xlc:
Followup of #676

Add new signed extension to get real tx origin and have it saved in storage so we can get the real origin account and charge storage rent from the origin.

PR #686, opened 2021-01-20
Add evm SignedExtension

zjb0807:
Closes: #677 

There still have a bug, which causes `https://polkadot.js.org` to fail to call the module's Call

### DIFF
--- a/modules/currencies/src/lib.rs
+++ b/modules/currencies/src/lib.rs
@@ -70,7 +70,7 @@ pub trait Config: frame_system::Config {
 
 	/// Mapping from address to account id.
 	type AddressMapping: AddressMapping<Self::AccountId>;
-	type EVMBridge: EVMBridge<BalanceOf<Self>>;
+	type EVMBridge: EVMBridge<Self::AccountId, BalanceOf<Self>>;
 }
 
 decl_event!(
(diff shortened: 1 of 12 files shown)


modules/currencies/src/lib.rs
```diff
@@ -295,11 +295,13 @@ impl<T: Config> MultiCurrency<T::AccountId> for Module<T> {
 			CurrencyId::ERC20(contract) => {
 				let sender = T::AddressMapping::get_evm_address(&from).ok_or(Error::<T>::AccountNotFound)?;
 				let address = T::AddressMapping::get_or_create_evm_address(&to);
+				let origin = T::EVMBridge::get_origin();
+				let origin_address = T::AddressMapping::get_evm_address(&origin).ok_or(Error::<T>::AccountNotFound)?;
```
xlc, line 299: use `get_or_create_evm_address` here

modules/evm/src/lib.rs
```diff
@@ -882,6 +885,16 @@ impl<T: Config> EVMTrait for Module<T> {
 			}
 		})
 	}
+
+	/// Get the real origin account and charge storage rent from the origin.
+	fn get_origin() -> T::AccountId {
+		ExtrinsicOrigin::<T>::get().expect("There must be a value through `SignedExtension::pre_dispatch`")
```
xlc, line 891: this can be called in `on_intialize` / `on_finalize` so there could be no origin. so should make it return `Option<T::AccountId>` instead

And in module-currencies, use unwrap_or_default when get origin

xlc (changes requested):

xlc (approved):

PR #697, opened 2021-01-26
Allow people to claim default EVM account

xlc:


modules/evm/src/lib.rs
```diff
@@ -446,7 +446,7 @@ decl_module! {
 		#[transactional]
 		pub fn deploy(origin, contract: EvmAddress) {
 			let who = ensure_signed(origin)?;
-			let address = T::AddressMapping::get_or_create_evm_address(&who);
+			let address = T::AddressMapping::get_evm_address(&who).ok_or(Error::<T>::AddressNotMapped)?;
 			T::Currency::transfer(&who, &T::TreasuryAccount::get(), T::DeploymentFee::get(), ExistenceRequirement::AllowDeath)?;
 			Self::mark_deployed(contract, Some(address))?;
```
4meta5, line 451: The call to `mark_deployed` can fail with error `ContractNotFound`, but the transfer in the line above could have succeeded. So order should be switched to prevent that scenario.

xlc: But if we switch order, the mark_deployed could success and transfer could fail.

That's why we have `#[transactional]` for all our calls so we don't need to worry about this anymore.

4meta5: Cool, did not know about this annotation, thanks for sharing!

Issue #631, opened 2020-12-14
EVM: Scheduler API - part 1


xlc:
API to allow EVM contracts to schedule contract call.

Precompiles to allow contract to schedule a call with `min_delay, gas_limit, target, input, value`.

Reserve `gas_limit * DefaultGasPrice`

New call `fn scheduled_call(origin, from, target, input, value, gas_limit)`, ensure_root and otherwise similar to `call`, refund unused gas and move used gas payment to treasury.

Use pallet-scheduler to schedule a call with `scheduled_call`

Predeployed contract to call the precompiles.

Some example code of using this feature.

PR #699, opened 2021-01-27
Add evm schedule

zjb0807:
Closes: #631


modules/evm/src/lib.rs
```diff
@@ -344,6 +344,39 @@ decl_module! {
...
+			value: BalanceOf<T>,
+			gas_limit: u32,
+			storage_limit: u32,
+		) -> DispatchResultWithPostInfo {
+			ensure_root(origin)?;
+
+			let from_account = T::AddressMapping::get_account_id(&from);
+			// unreserve the deposit for gas_limit and storage_limit
+			let total_fee = T::StorageDepositPerByte::get().saturating_mul(storage_limit.into()).saturating_add(gas_limit.into());
+			T::Currency::unreserve(&from_account, total_fee);
```
4meta5, line 362: Consider unreserve after the Runner `call` because it is fallible. Would not want to unreserve if the Runner::call fails and returns `Error::<T>::NoPermission` for example. Then, someone could call this without permission on purpose to unreserve that total fee amount.

zjb0807: @xlc 
Maybe should remove unreserve and reserve. 
Just judge whether the free_balance >= total_fee

xlc: The fee is not handle correctly here, but can fix it in later PR https://github.com/AcalaNetwork/Acala/issues/700

runtime/common/src/precompile/schedule_call.rs
```diff
@@ -0,0 +1,145 @@
...
+
+				let delay = DispatchTime::After(min_delay);
+				let origin = Origin::root().caller().clone();
+
+				let from_account = AddressMapping::get_account_id(&from);
+
+				// reserve the deposit for gas_limit and storage_limit
+				let total_fee = Runtime::StorageDepositPerByte::get()
+					.saturating_mul(storage_limit.into())
+					.saturating_add(gas_limit.into());
```
xlc, line 119: this isn't right, but can happen with https://github.com/AcalaNetwork/Acala/issues/700

runtime/common/src/precompile/schedule_call.rs
```diff
@@ -0,0 +1,145 @@
...
+					)),
+					delay,
+					None,
+					0,
+					origin,
+					call,
+				)
+				.map_err(|_| ExitError::Other("Scheduler failed".into()))?;
+
+				Ok((ExitSucceed::Returned, vec![], 0))
```
xlc, line 140: make the return the schedule task ID so it can be later used by https://github.com/AcalaNetwork/Acala/issues/701

runtime/common/src/precompile/schedule_call.rs
```diff
@@ -0,0 +1,145 @@
...
+				Runtime::Currency::reserve(&from_account, total_fee).map_err(|e| {
+					let err_msg: &str = e.into();
+					ExitError::Other(err_msg.into())
+				})?;
+
+				Scheduler::schedule_named(
+					Encode::encode(&(
+						&"ScheduleCall",
+						from,
+						input_data,
```
xlc, line 129: I think this is just too much data in the name. Maybe we can just have a `"ScheduleCall" + sequential ID`?
The ID can be storage with
```
parameter_types! {
     pub storage EvmSchedullerNextID: u32 = 0u32;
}
```

https://github.com/paritytech/polkadot/pull/2167/files#diff-8a10805ad559865410470a16150cfa44ea0c156becadfc749c10aab87e8c93beR37

xlc (changes requested):

xlc (approved):

Issue #698, opened 2021-01-26
Weights & benchmarking & tests for claim default EVM account #697


xlc:
Need weights & benchmarking for the new call and some tests for updated behavior for #697

xlc:
Closed by #721

PR #721, opened 2021-02-03
Evm claim/weights

brettkolodny:
Added benchmarking and weight functions for `claim_default_account` in the `evm-accounts` module. #698

### DIFF
--- a/modules/evm-accounts/src/lib.rs
+++ b/modules/evm-accounts/src/lib.rs
@@ -43,6 +43,7 @@ pub mod module {
 
 	pub trait WeightInfo {
 		fn claim_account() -> Weight;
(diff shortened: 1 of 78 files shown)


xlc (approved):

modules/evm-accounts/src/lib.rs
```diff
@@ -149,7 +150,7 @@ pub mod module {
 		}
 
 		// TODO: weights & benchmarking
```
xlc, line 152: I clicked merge button too fast... can you delete this TODO?

brettkolodny: Will do

Issue #701, opened 2021-01-27
EVM: Scheduler part 3


xlc:
New API to cancel previous scheduled call by scheduled task ID
New API to adjust the delay of scheduled call by scheduled task ID

PR #731, opened 2021-02-08
Schedule add reschecule and cancel

zjb0807:
Closes: #701

### DIFF
--- a/modules/support/src/lib.rs
+++ b/modules/support/src/lib.rs
@@ -361,9 +361,3 @@ impl<AccountId, Balance: Default + Copy, NegativeImbalance: Imbalance<Balance>>
 		Ok(())
 	}
 }
-
-pub trait EVMSchedulerManager<BlockNumber, AccountId, Balance> {
-	fn schedule(caller: &AccountId, task_id: &[u8], block_number: BlockNumber, fee: Balance) -> DispatchResult;
-	fn cancel(caller: &AccountId, task_id: &[u8]) -> Result<Balance, DispatchError>;
-	fn reschedule(caller: &AccountId, task_id: &[u8], new_block_number: BlockNumber) -> DispatchResult;
-}
--- a/modules/support/src/lib.rs
+++ b/modules/support/src/lib.rs
@@ -326,7 +326,7 @@ pub trait EVMStateRentTrait<AccountId, Balance> {
 
 pub trait TransactionPayment<AccountId, Balance, NegativeImbalance> {
 	fn reserve_fee(who: &AccountId, weight: Weight) -> Result<Balance, DispatchError>;
-	fn unreserve_fee(_who: &AccountId, _fee: Balance);
+	fn unreserve_fee(who: &AccountId, fee: Balance);
 	fn unreserve_and_charge_fee(
 		who: &AccountId,
 		weight: Weight,
--- a/modules/support/src/lib.rs
+++ b/modules/support/src/lib.rs
@@ -325,7 +325,8 @@ pub trait EVMStateRentTrait<AccountId, Balance> {
 }
 
 pub trait TransactionPayment<AccountId, Balance, NegativeImbalance> {
-	fn reserve_fee(who: &AccountId, weight: Weight) -> DispatchResult;
+	fn reserve_fee(who: &AccountId, weight: Weight) -> Result<Balance, DispatchError>;
+	fn unreserve_fee(_who: &AccountId, _fee: Balance);
 	fn unreserve_and_charge_fee(
 		who: &AccountId,
 		weight: Weight,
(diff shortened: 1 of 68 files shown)


xlc (commented):
https://github.com/paritytech/substrate/blob/29aca981db5e8bf8b5538e6c7920ded917013ef3/frame/scheduler/src/lib.rs#L703-L712

Should be using those methods for cancel / reschedule

runtime/common/src/precompile/schedule_call.rs
```diff
@@ -161,29 +171,78 @@ impl<AccountId, AddressMapping, Scheduler, ChargeTransactionPayment, Call, Origi
...
+					ExitError::Other(err_msg.into())
+				})?;
+
+				#[cfg(not(feature = "with-ethereum-compatibility"))]
+				{
+					//// unreserve the transaction fee for gas_limit
+					ChargeTransactionPayment::unreserve_fee(&from_account, fee);
+				}
+
+				Scheduler::cancel_named(task_id).map_err(|_| ExitError::Other("Cancel schedule failed".into()))?;
```
zjb0807, line 208: Here use the scheduler  to do `cancel/reschedule`

zjb0807: EVMSchedulerManager just saves the caller and checks whether it has permission.

zjb0807:
It is found that `schedule_named` and `cancel_named` are called with root, and the origin cannot be checked.
Redefine `task_id = id + sender + fee`.
So check origin directly through `task_id`.

modules/support/src/lib.rs
```diff
@@ -325,7 +325,8 @@ pub trait EVMStateRentTrait<AccountId, Balance> {
 }
 
 pub trait TransactionPayment<AccountId, Balance, NegativeImbalance> {
-	fn reserve_fee(who: &AccountId, weight: Weight) -> DispatchResult;
+	fn reserve_fee(who: &AccountId, weight: Weight) -> Result<Balance, DispatchError>;
+	fn unreserve_fee(_who: &AccountId, _fee: Balance);
```
xlc, line 329:
```suggestion
	fn unreserve_fee(who: &AccountId, fee: Balance);
```

runtime/common/src/precompile/schedule_call.rs
```diff
@@ -155,37 +161,97 @@ impl<AccountId, AddressMapping, Scheduler, ChargeTransactionPayment, Call, Origi
...
 			}
 		}
 	}
 }
 
-fn vec_u8_from_tuple(task_address: TaskAddress<BlockNumber>) -> Vec<u8> {
-	let mut be_bytes_0 = [0u8; 32];
-	U256::from(task_address.0).to_big_endian(&mut be_bytes_0[..]);
+fn join_task_id(id: Vec<u8>, sender: H160, fee: Balance) -> Vec<u8> {
+	let mut be_bytes = [0u8; 96];
```
xlc, line 240: you can do `Encode::encode((id, sender, Compact(fee)))`, and `Decode::<(Vec<u8>, H160, Compact<Balance>)>::decode(input)`

Also use compact to save some bytes

runtime/common/src/precompile/schedule_call.rs
```diff
@@ -155,37 +161,97 @@ impl<AccountId, AddressMapping, Scheduler, ChargeTransactionPayment, Call, Origi
...
+
+				// task_id = id + sender + reserve_fee
+				let (_, sender, fee) = split_task_id(&task_id);
+				ensure!(sender == from, ExitError::Other("NoPermission".into()));
+
+				#[cfg(not(feature = "with-ethereum-compatibility"))]
+				{
+					// unreserve the transaction fee for gas_limit
+					let from_account = AddressMapping::get_account_id(&from);
+					ChargeTransactionPayment::unreserve_fee(&from_account, fee.into());
```
xlc, line 203: this can only happen if cancel is successful. i.e. we have verified the task id is valid

xlc (changes requested):

runtime/common/src/precompile/schedule_call.rs
```diff
@@ -13,17 +13,25 @@ use frame_support::{
...
-use codec::Encode;
+use super::input::{Input, InputT, PER_PARAM_BYTES};
+use codec::{Decode, Encode};
 use pallet_scheduler::TaskAddress;
 
 parameter_types! {
 	pub storage EvmSchedulerNextID: u32 = 0u32;
 }
 
+#[derive(Debug, PartialEq, Encode, Decode)]
```
xlc, line 27:
```suggestion
#[derive(RuntimeDebug, PartialEq, Encode, Decode)]
```

runtime/common/src/precompile/schedule_call.rs
```diff
@@ -13,17 +13,25 @@ use frame_support::{
...
+struct TaskInfo {
+	id: Vec<u8>,
```
xlc, line 29:
```suggestion
	prefix: &'static str,
	id: u32,
```

to avoid another encode

xlc (approved):

runtime/common/src/precompile/tests.rs
```diff
@@ -215,14 +219,65 @@ fn schedule_call_precompile_should_work() {
...
+		let event = TestEvent::pallet_scheduler(pallet_scheduler::RawEvent::Scheduled(3, 0));
+		assert!(System::events().iter().any(|record| record.event == event));
 
-		// returned block + index
-		let mut expected_output = [0u8; 64];
-		U256::from(3).to_big_endian(&mut expected_output[..32]);
-		U256::from(0).to_big_endian(&mut expected_output[32..64]);
+		let mut num = [0u8; 8];
+		num[..].copy_from_slice(&output[32 - 8..32]);
+		let task_id_len: usize = usize::from_be_bytes(num);
```
xlc, line 229: usize is u64 in std and u32 in wasm, so I think should use u32 here

PR #742, opened 2021-02-11
precomiple error as fatal

ermalkaleci:
temporary workaround that treats precompile error as fatal in order to exit execution with error message

### DIFF
--- a/modules/evm/src/runner/handler.rs
+++ b/modules/evm/src/runner/handler.rs
@@ -6,7 +6,7 @@ use crate::{
 	AccountInfo, AccountStorages, Accounts, AddressMapping, Codes, Config, ContractInfo, Error, Event, Log,
 	MergeAccount, Pallet, Vicinity,
 };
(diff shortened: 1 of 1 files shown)


xlc (commented):
All different exit type are super confusing... Could you describe what are the difference between Exit & ExitFatal? 
What is the effect on the client side?
Some more test to cover this change?

ermalkaleci:
@xlc when evm enters a sub-call it will go through handler and then match exit_reason. Evm is design to ignore ExitError and continue execution https://github.com/rust-blockchain/evm/blob/2213a2020c943f5d3912bbcdf51df181d17c94a8/runtime/src/eval/system.rs#L403 That's why I thought maybe this is a bug but apparently it was by design. ExitFatal will stop execution with the error. So converting precompile ExitError into ExitFatal with stop execution and get out module error otherwise it will continue execution until revert or stop without any error. Maybe this is an issue we need to have further discussions

ermalkaleci:
to be discussed later

Issue #685, opened 2021-01-18
EVM: improve gas estimation logic


xlc:
https://github.com/paritytech/frontier/commit/fe13ec2675165014e0bb92de6325bfe0765cfbbe

PR #752, opened 2021-02-18
Improve gas estimation logic

zjb0807:
Closes: #685


xlc (approved):
I think we can enable this by default

PR #762, opened 2021-02-25
Fix bugs

zjb0807:
1. fix pre-deploy contract currency id.
2. fix estimate_gas API(If the transaction fails, there will be an endless loop).


xlc (approved):
Some test of the RPC will be helpful

zjb0807:
> Some test of the RPC will be helpful

Testing RPC is not easy to implement, because `substrate_test_runtime_client` is not published.

xlc:
> > Some test of the RPC will be helpful
> 
> Testing RPC is not easy to implement, because `substrate_test_runtime_client` is not published.

should we make an issue on Substrate?

PR #769, opened 2021-03-01
update prices to convert the feed prices from oracle with decimal consideration

wangjj9219:


modules/prices/src/lib.rs
```diff
@@ -129,19 +136,26 @@ impl<T: Config> PriceProvider<CurrencyId> for Pallet<T> {
 		}
 	}
 
-	/// get price in USD
+	/// get the exchange rate of specific currency to USD
```
xlc, line 139:
```suggestion
	/// get the exchange rate of specific currency to USD
	/// Note: this returns the price for 1 basic unit
```

runtime/mandala/src/lib.rs
```diff
@@ -770,6 +770,13 @@ parameter_types! {
 	pub StableCurrencyFixedPrice: Price = Price::saturating_from_rational(1, 1);
 }
 
+parameter_type_with_key! {
+	pub TokenDecimals: |currency_id: CurrencyId| -> u32 {
```
xlc, line 774: this can be in runtime common, all runtimes should share same config

PR #779, opened 2021-03-02
config token decimal

wangjj9219:
config tokens decimal:
ACA => 13
KAR => 12
AUSD/KUSD => 12
DOT/LDOT => 10
KSM/LKSM => 12
and default decimal is 18

### DIFF
--- a/runtime/common/src/lib.rs
+++ b/runtime/common/src/lib.rs
@@ -11,14 +11,14 @@ use frame_support::{
 };
 use frame_system::limits;
 pub use module_support::{ExchangeRate, PrecompileCallerFilter, Price, Rate, Ratio};
-use primitives::{CurrencyId, PRECOMPILE_ADDRESS_START, PREDEPLOY_ADDRESS_START};
+use orml_traits::{parameter_type_with_key, GetByKey};
+use primitives::{Balance, CurrencyId, TokenSymbol, PRECOMPILE_ADDRESS_START, PREDEPLOY_ADDRESS_START};
 use sp_core::H160;
 use sp_runtime::{
 	traits::{Convert, Saturating},
 	transaction_validity::TransactionPriority,
 	FixedPointNumber, FixedPointOperand, Perbill,
 };
-
 use static_assertions::const_assert;
 
 pub mod precompile;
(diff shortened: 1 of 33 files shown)


runtime/common/src/lib.rs
```diff
@@ -339,13 +339,64 @@ parameter_types! {
...
+			RENBTC => 8,
+			POLKABTC => 8,
+
+			_ => 18,
+		}
 	};
 }
 
+pub fn dollar(currency_id: CurrencyId) -> Balance {
+	let xixi = TokenDecimals::get(&currency_id);
```
xlc, line 380: what is xixi

runtime/common/src/lib.rs
```diff
@@ -339,13 +339,64 @@ parameter_types! {
...
+pub const KAR: CurrencyId = CurrencyId::Token(TokenSymbol::KAR);
+pub const KUSD: CurrencyId = CurrencyId::Token(TokenSymbol::KUSD);
+pub const KSM: CurrencyId = CurrencyId::Token(TokenSymbol::KSM);
+pub const LKSM: CurrencyId = CurrencyId::Token(TokenSymbol::LKSM);
+pub const XBTC: CurrencyId = CurrencyId::Token(TokenSymbol::XBTC);
+pub const RENBTC: CurrencyId = CurrencyId::Token(TokenSymbol::RENBTC);
+pub const POLKABTC: CurrencyId = CurrencyId::Token(TokenSymbol::POLKABTC);
+pub const PLM: CurrencyId = CurrencyId::Token(TokenSymbol::PLM);
+pub const PHA: CurrencyId = CurrencyId::Token(TokenSymbol::PHA);
+pub const SDN: CurrencyId = CurrencyId::Token(TokenSymbol::SDN);
```
xlc, line 355: I think we can generate those in with macro

https://github.com/AcalaNetwork/Acala/blob/master/primitives/src/currency.rs#L13

runtime/common/src/lib.rs
```diff
@@ -339,13 +339,64 @@ parameter_types! {
...
+
+parameter_type_with_key! {
 	pub TokenDecimals: |currency_id: CurrencyId| -> u32 {
-		// TODO: config
-		18
+		match *currency_id {
```
xlc, line 359: maybe we should just have a `trait GetDecimals` and have CurrencyId implement it

and we can also have it include helpers like `dollar`

xlc (approved):

shaunxw (approved):

PR #788, opened 2021-03-04
make prices module provide price for lp token

wangjj9219:


modules/dex/src/lib.rs
```diff
@@ -527,18 +527,18 @@ impl<T: Config> Pallet<T> {
 					|| provision_parameters.accumulated_provision.1 >= provision_parameters.target_provision.1)
 			{
 				// calculate initial price
-				let initial_price_0_in_1: Price = Price::checked_from_rational(
-					provision_parameters.accumulated_provision.1,
+				let initial_price_1_in_0: Price = Price::checked_from_rational(
```
xlc, line 530: why this change? should we just always use the big number divide by small number?

xlc (approved):

modules/dex/src/lib.rs
wangjj9219: ok, when calculating the initial share amount, always use the big number divide by the small number in order to minimize the loss. In this way, the decimals of lptoken cannot be determined directly,  the front end can use the default 18 to handle display.

PR #797, opened 2021-03-10
Add token decimals

zjb0807:


primitives/src/currency.rs
```diff
@@ -107,25 +109,25 @@ create_currency_id! {
...
-		ACA(13) = 0,
-		AUSD(12) = 1,
-		DOT(10) = 2,
-		LDOT(10) = 3,
-		XBTC(8) = 4,
-		RENBTC(8) = 5,
-		POLKABTC(8) = 6,
-		PLM(18) = 7,
-		PHA(18) = 8,
+		ACA("Acala Token", 13) = 0,
```
xlc, line 112: @bette7 can you provide all the token names?

zjb0807: copy from https://github.com/airalab/robonomics/blob/bd17fecc0aedf50e5895f4a4ae16d946ed87eab4/bin/node/primitives/src/lib.rs#L83-L100

qwer951123: @xlc token HDT is in current @acala-network/types, but missing in the current chain config, should it update?

xlc: Right. @zjb0807 can you also add HDT = 9, decimal = 12, HydraDX Token

xlc (approved):

PR #805, opened 2021-03-14
Use cumulus xcm-handler to execute XCM locally.

shaunxw:
Companion of https://github.com/open-web3-stack/open-runtime-module-library/pull/401


runtime/karura/src/lib.rs
```diff
@@ -1250,6 +1252,13 @@ impl cumulus_pallet_xcm_handler::Config for Runtime {
 	type AccountIdConverter = LocationConverter;
 }
 
+pub struct HandleXcm;
+impl XcmHandlerT<AccountId> for HandleXcm {
```
xlc, line 1256: I think it is safe to move this into common?

shaunxw: The impl needs `XcmHandler` from `construct_runtime` of each runtime. I think they are not accessible in common runtime.

xlc (approved):

Issue #754, opened 2021-02-18
Charge transaction fee for renvm-bridge mint


xlc:
Use `TransactionPayment` trait to charge fee after mint successed

PR #806, opened 2021-03-15
Add charge transaction fee

zjb0807:
Closes: #754


ecosystem-modules/ren/renvm-bridge/src/lib.rs
```diff
@@ -107,13 +121,20 @@ decl_module! {
...
 			#[compact] amount: Balance,
-			_n_hash: [u8; 32],
+			n_hash: [u8; 32],
 			sig: EcdsaSignature,
 		) {
 			ensure_none(origin)?;
-			Self::do_mint(who, amount, sig)?;
+			Self::do_mint(&who, amount, &sig)?;
+
+			let len = (32 + p_hash.len() + n_hash.len() + sig.0.len()) as u32;
```
xlc, line 132: this is not accurate as it doesn't include everything in a transaction.
should just make build a dummy mint transaction, measure the length, and hardcode the value here

ecosystem-modules/ren/renvm-bridge/src/lib.rs
```diff
@@ -40,16 +46,24 @@ type EcdsaSignature = ecdsa::Signature;
...
-	type Currency: BasicCurrency<Self::AccountId, Balance = Balance>;
+/// Type alias for currency balance.
+pub type BalanceOf<T> = <<T as Config>::Currency as Currency<<T as frame_system::Config>::AccountId>>::Balance;
+pub type NegativeImbalanceOf<T> =
+	<<T as Config>::Currency as Currency<<T as frame_system::Config>::AccountId>>::NegativeImbalance;
+
+pub trait Config: frame_system::Config {
+	type Event: From<Event<Self>> + Into<<Self as frame_system::Config>::Event>;
+	type Currency: Currency<Self::AccountId>;
+	type CurrencyAdapter: BasicCurrency<Self::AccountId, Balance = Balance>;
```
xlc, line 57:
```suggestion
	type BridgedTokenCurrency: BasicCurrency<Self::AccountId, Balance = Balance>;
```

ecosystem-modules/ren/renvm-bridge/src/lib.rs
```diff
@@ -113,7 +127,16 @@ decl_module! {
 			sig: EcdsaSignature,
 		) {
 			ensure_none(origin)?;
-			Self::do_mint(who, amount, sig)?;
+			Self::do_mint(&who, amount, &sig)?;
+
+			// Calculated the transaction length from the unit test.
+			let len: u32 = 166;
```
xlc, line 133: make this a `const MINT_TX_LENGTH: u32 = 166` and have test check the const

xlc (approved):

ecosystem-modules/ren/renvm-bridge/src/lib.rs
zjb0807: I think the dynamic calculation is better because the amount is compact.

PR #809, opened 2021-03-15
Inline Documentation for EVM

brettkolodny:


modules/evm/src/lib.rs
```diff
@@ -870,6 +886,15 @@ impl<T: Config> Pallet<T> {
 		Ok(())
 	}
 
+	/// Selfdestruct a contract at a given address.
+	///
+	/// - Ensure that the given maintainer is the contract maintainer.
```
xlc, line 891: Try avoid just describe the code in English, should describe the parameters (and constraints), what happens (without all the details), and the returns

Also no need to have detailed docs for internal method unless the method is complicated.

modules/evm/src/lib.rs
```diff
@@ -644,6 +652,7 @@ pub mod module {
 			Ok(().into())
 		}
 
+		/// Set the code of a contract at a given address.
```
xlc, line 655: We want have more docs for all the extrinsics including the input parameters, and constraints.

xlc (approved):

Issue #804, opened 2021-03-14
Replace RPC estimateGas to estimateResources


xlc:
Make it return storage usage and adjusted weight fee from `payment_queryFeeDetails`

Return type will be `{gas: used_gas, storage: used_storage, weight_fee: adjusted_weight_fee}`

Used storage count be negative which means refund, in that case make it zero.

PR #815, opened 2021-03-16
Estimate resources

zjb0807:
Closes: #804


primitives/src/evm.rs
```diff
@@ -47,7 +47,7 @@ pub struct CreateInfo {
 	pub address: EvmAddress,
 	pub output: Vec<u8>,
 	pub used_gas: U256,
-	pub used_storage: i32,
+	pub used_storage: U256,
```
xlc, line 50: make it i32 as this value could be negative

modules/evm/rpc/src/call_request.rs
```diff
@@ -39,3 +39,16 @@ pub struct CallRequest {
...
+
+/// EstimateResources response
+#[derive(Debug, Default, PartialEq, Serialize, Deserialize, Clone)]
+#[serde(deny_unknown_fields)]
+#[serde(rename_all = "camelCase")]
+pub struct EstimateResourcesResponse {
+	/// Used gas
+	pub gas: U256,
+	/// Used storage
+	pub storage: U256,
```
xlc, line 51:
```suggestion
	pub storage: i32,
```

modules/evm/rpc/src/call_request.rs
```diff
@@ -39,3 +39,16 @@ pub struct CallRequest {
...
+	/// Adjusted weight fee
+	pub weight_fee: U256,
```
xlc, line 53: should be whatever original adjusted_weight_fee type

xlc (changes requested):
used_storage could be neg

modules/evm/rpc/src/call_request.rs
zjb0807: The serde does not support u128, use U256 instead of u128.

modules/evm/rpc/src/lib.rs
```diff
@@ -190,12 +190,43 @@ where
...
 	}
 
-	fn estimate_resources(
-		&self,
-		extrinsic: Bytes,
-		request: CallRequest,
-		_: Option<B>,
-	) -> Result<EstimateResourcesResponse> {
+	fn estimate_resources(&self, extrinsic: Bytes, _: Option<B>) -> Result<EstimateResourcesResponse> {
+		let utx = mandala_runtime::UncheckedExtrinsic::decode(&mut &*extrinsic).map_err(|_| Error {
```
xlc, line 195: this needs to be generic otherwise we can't use it on other runtime

One way is to have a EvmCallMatcher trait and each runtime will provide its implementation

xlc (changes requested):

xlc (approved):

xlc:
@ntduan this will update rpc method name and breaks bodhi.js

xlc (approved):

PR #835, opened 2021-03-23
add KILT Coin as NonNativeCurrency

weichweich:
- Adds the kilt coin to the Mandala Runtime

### DIFF
--- a/primitives/src/currency.rs
+++ b/primitives/src/currency.rs
@@ -138,6 +138,7 @@ create_currency_id! {
 		// Reserve for POLKABTC = 134
 		SDN("Shiden", 18) = 135,
 		// Reserve for PHA = 136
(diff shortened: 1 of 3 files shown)


primitives/src/currency.rs
```diff
@@ -138,6 +138,7 @@ create_currency_id! {
 		// Reserve for POLKABTC = 134
 		SDN("Shiden", 18) = 135,
 		// Reserve for PHA = 136
+		KILT("Kilt", 15) = 137,
```
xlc, line 141: Is KILT going to be a Kusama parachain or Polkadot parachain? Make it 10 for Polkadot parachain

weichweich: we are going to be a Kusama chain for now.

PR #838, opened 2021-03-23
integrate chainsafe

wangjj9219:


ecosystem-modules/chainsafe/src/lib.rs
```diff
@@ -0,0 +1,214 @@
...
+			dest_chain_id,
+			resource_id,
+			recipient,
+			sp_core::U256::from(amount.saturated_into::<u128>()),
+		)
+	}
+
+	fn is_origin_chain_resource(resource_id: ResourceId) -> bool {
+		let origin_chain_id = <T as chainbridge::Config>::ChainId::get();
+		resource_id[31] == origin_chain_id
```
xlc, line 212: is possible for this to panic due to access invalid index?

xlc (approved):
later we need to think about how to map chain safe asset to ERC20 to support e.g. ampl crosschain transfer

ecosystem-modules/chainsafe/src/lib.rs
wangjj9219: `ResourceId` is alias of `[u8; 32]`

PR #849, opened 2021-03-29
Add add_liquidity and remove_liquidity api

zjb0807:


modules/dex/src/lib.rs
```diff
@@ -766,6 +766,8 @@ impl<T: Config> Pallet<T> {
 		})
 	}
 
+	/// Ensured atomic.
```
xlc, line 769: No need to have this in comment.

xlc (approved):

PR #853, opened 2021-03-30
Added Bit.Country Token

justinphamnz:
Added BCG to supported currencies.

### DIFF
--- a/primitives/src/currency.rs
+++ b/primitives/src/currency.rs
@@ -127,7 +127,7 @@ create_currency_id! {
 		PLM("Plasm", 18) = 7,
 		PHA("Phala", 18) = 8,
 		HDT("HydraDX", 12) = 9,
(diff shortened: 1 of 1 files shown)


primitives/src/currency.rs
```diff
@@ -127,6 +127,7 @@ create_currency_id! {
 		PLM("Plasm", 18) = 7,
 		PHA("Phala", 18) = 8,
 		HDT("HydraDX", 12) = 9,
+		BCG("Bit.Country", 18) = 10,
```
xlc, line 130: Actually can you make this 11 and reserve 10 for Kilt? Just to keep everything consistent

xlc (approved):

primitives/src/currency.rs
justinphamnz: Yes, no problem. I already updated to 11.

Issue #848, opened 2021-03-29
Replace let _ = ... in the Runtime with debug_assert!


xlc:
Something like https://github.com/paritytech/substrate/issues/8106

wangjj9219:
closed by #865

PR #865, opened 2021-04-05
Issue 848

ferrell-code:
relates #848, be aware the rust compiler panicked on one build which was strange but should work fine now :)


xlc:
Thanks for the PR!

Seems like it does catch something unexpected. @wangjj9219 can you take a look of the failed benchmarking test?

wangjj9219:
already fixed by #896

PR #869, opened 2021-04-06
Update currency.rs

drewstone:
Adds Edgeware & Hedgeware.


xlc:
Thanks for the PR. We are undergoing some major refactor while resulting no need to add everything parachain token to our CurrencyId. Once that's done, we will also remove existing ones so I will close this.

PR #871, opened 2021-04-07
Integrate new cross-chain assets transfer impl

shaunxw:


primitives/src/currency.rs
```diff
@@ -130,7 +130,7 @@ create_currency_id! {
 		BCG("Bit.Country", 18) = 11,
 
 		// Kusama Ecosystem
-		KAR("Karura", 12) = 128,
+		KAR("Karura", 12) = 128,:
```
xlc, line 133:
```suggestion
		KAR("Karura", 12) = 128,
```

xlc (approved):

PR #883, opened 2021-04-08
NFT refactor

xlc:
- Use the Currency from pallet_proxy
- Token deposit is now reserved on token owner account, instead of class owner account. This allow it to be used as existential deposit for token owner.
- Transfer token will also transfer the token deposit reserve
- Destroy class will also remove proxy and transfer all free balance out


zjb0807 (approved):
LGTM

PR #885, opened 2021-04-08
accumulate loans interest by timestamp

wangjj9219:


modules/cdp-engine/src/lib.rs
```diff
@@ -119,7 +120,12 @@ pub mod module {
 	use super::*;
 
 	#[pallet::config]
-	pub trait Config: frame_system::Config + loans::Config + SendTransactionTypes<Call<Self>> {
+	pub trait Config:
+		frame_system::Config
+		+ loans::Config
+		+ pallet_timestamp::Config<Moment = Moment>
```
xlc, line 126: Better to use this UnixTime

https://github.com/paritytech/substrate/blob/4898eb350d636d439827cb43f0f26e72e66492fa/frame/support/src/traits/misc.rs#L160

modules/cdp-engine/src/lib.rs
```diff
@@ -506,6 +500,41 @@ pub mod module {
...
 impl<T: Config> Pallet<T> {
+	fn accumulate_interest(now: Moment, last_accumulate_timestamp: Moment) -> u32 {
+		let mut count: u32 = 0;
+
+		if !T::EmergencyShutdown::is_shutdown() && !now.is_zero() {
+			let interval = now.saturating_sub(last_accumulate_timestamp);
+			let multiplier = Rate::checked_from_rational(interval, ONE_YEAR).unwrap_or_default();
+
+			for currency_id in T::CollateralCurrencyIds::get() {
+				let annual_interest_rate = Self::get_annual_interest_rate(currency_id);
```
xlc, line 511: Should have some tests to verify that running this for a year will actually result something similar to target rate to ensure we don't have too much precision lose due to arithmetic

modules/cdp-engine/src/lib.rs
```diff
@@ -506,6 +500,41 @@ pub mod module {
...
+				let total_debits = <LoansOf<T>>::total_positions(currency_id).debit;
+
+				if !annual_interest_rate.is_zero() && !total_debits.is_zero() {
+					let debit_exchange_rate = Self::get_debit_exchange_rate(currency_id);
+					let debit_exchange_rate_increment =
+						debit_exchange_rate.saturating_mul(annual_interest_rate.saturating_mul(multiplier));
```
xlc, line 517: This doesn't do compounding so we need to run time numbers to verify if this still result values within acceptable range

wangjj9219: em, I ignored compounding. Perhaps set the interest rate of a second to simplify calculation?

xlc: that will still have the same issue.
I think we should just check if the number is not off too much. otherwise do real compounding calculation

wangjj9219: Assuming per 12 seconds can produce a block, to calculate the real compound interest rate by seconds, the calculation consumption and loss are acceptable.

xlc (approved):

PR #909, opened 2021-04-14
add new call to close cdp by dex

wangjj9219:


modules/honzon/src/lib.rs
```diff
@@ -122,6 +122,19 @@ pub mod module {
 			Ok(().into())
 		}
 
+		#[pallet::weight(<T as Config>::WeightInfo::close_loan_has_debit_by_dex())]
+		#[transactional]
+		pub fn close_loan_has_debit_by_dex(
+			origin: OriginFor<T>,
+			currency_id: CurrencyId,
+			maybe_path: Option<sp_std::vec::Vec<CurrencyId>>,
```
xlc, line 130: should we require to pass a path?

xlc (approved):

modules/honzon/src/lib.rs
wangjj9219: This operation allows users to use collateral locked in CDP to trade with DEX,  I think should allow them to choose the best trading path.

PR #919, opened 2021-04-17
check storage limit on finish

### LINKED ISSUE #916: EVM: storage limit estimation issue
Currently estimate resource only give the final storage usage, and it is possible the maximum usage during contract execution is very high and therefore the final usage value is not suitable as a limit.

We should just ignore storage limit during charge and enforce the limit at very end of the contract execution. We still charge gas for all the writes so if should be ok to allow user doing unnecessary work and revert at very end.

ermalkaleci:
Closes #916

### DIFF
--- a/modules/evm/src/runner/storage_meter.rs
+++ b/modules/evm/src/runner/storage_meter.rs
@@ -376,6 +376,25 @@ mod tests {
 	}
 
 	#[test]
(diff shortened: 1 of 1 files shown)


zjb0807 (approved):

xlc (approved):
add new test for low limit should success with hight storage & refund

PR #938, opened 2021-04-21
Update evm bridge

zjb0807:


modules/currencies/src/mock.rs
```diff
@@ -205,26 +203,51 @@ frame_support::construct_runtime!(
 	}
 );
 
-pub fn alice() -> AccountId {
-	<Runtime as Config>::AddressMapping::get_account_id(
-		&H160::from_str("1000000000000000000000000000000000000001").unwrap(),
-	)
+pub fn alice() -> EvmAddress {
```
shaunxw, line 206: I think we can keep `alice` function to return `AccountId`, and name the new one returning `EvmAddress` as `alice_evm_addr`.

modules/evm-bridge/src/lib.rs
```diff
@@ -172,7 +171,11 @@ impl<T: Config> EVMBridgeTrait<AccountIdOf<T>, BalanceOf<T>> for Pallet<T> {
 			ExecutionMode::Execute,
 		)?;
 
-		Self::handle_exit_reason(info.exit_reason)
+		Self::handle_exit_reason(info.exit_reason)?;
+
+		// In order to avoid empty contracts, must check for return values
+		ensure!(info.output.len() == 32, Error::<T>::InvalidReturnValue);
```
xlc, line 177: actually I think we should do something similar to SafeERC20 https://github.com/OpenZeppelin/openzeppelin-contracts/blob/165e6f19489786c9c6abfcc9bdc8f2815d807935/contracts/token/ERC20/utils/SafeERC20.sol#L66

It is fine to have no return, but if there is return, it should be true

xlc (changes requested):

modules/evm-bridge/src/lib.rs
```diff
@@ -172,7 +171,11 @@ impl<T: Config> EVMBridgeTrait<AccountIdOf<T>, BalanceOf<T>> for Pallet<T> {
 			ExecutionMode::Execute,
 		)?;
 
-		Self::handle_exit_reason(info.exit_reason)
+		Self::handle_exit_reason(info.exit_reason)?;
+
+		// In order to avoid empty contracts, must check for return values
```
shaunxw, line 176:
```suggestion
		// Check return value to make sure not calling on empty contracts.
```

modules/evm-manager/src/lib.rs
```diff
@@ -175,9 +175,8 @@ impl<T: Config> CurrencyIdMapping for EvmCurrencyIdMapping<T> {
 	fn encode_currency_id(val: CurrencyId) -> Option<[u8; 32]> {
 		let mut bytes = [0u8; 32];
 		match val {
-			CurrencyId::Token(_) => {
-				let id: u32 = val.try_into().expect("CurrencyId::Token into u32 is success; qed");
-				bytes[12..16].copy_from_slice(&id.to_be_bytes()[..]);
+			CurrencyId::Token(token) => {
+				bytes[31] = token as u8;
```
shaunxw, line 179: `as` usage looks dangerous. I would prefer to have `Into<u8>` impl for `TokenSymbol` and call it, to ensure type safety.

runtime/mandala/tests/integration_test.rs
```diff
@@ -51,13 +54,10 @@ const ALICE: [u8; 32] = [4u8; 32];
...
 const LPTOKEN: CurrencyId =
 	CurrencyId::DexShare(DexShare::Token(TokenSymbol::AUSD), DexShare::Token(TokenSymbol::RENBTC));
-const LPTOKEN_ERC20: CurrencyId =
-	CurrencyId::DexShare(DexShare::Erc20(ERC20_ADDRESS_0), DexShare::Erc20(ERC20_ADDRESS_1));
 
-pub const ERC20_ADDRESS_0: H160 = H160([32, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 1]);
-pub const ERC20_0: CurrencyId = CurrencyId::Erc20(ERC20_ADDRESS_0);
-pub const ERC20_ADDRESS_1: H160 = H160([32, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 2]);
-pub const ERC20_1: CurrencyId = CurrencyId::Erc20(ERC20_ADDRESS_1);
+pub fn evm_alice() -> EvmAddress {
```
shaunxw, line 58:
```suggestion
pub fn alice_evm_addr() -> EvmAddress {
```

modules/evm-bridge/src/lib.rs
zjb0807: https://github.com/AcalaNetwork/Acala/pull/938/files#diff-7efb0e44e274d6054a16c345d5e935867bf0b44dc8960e5219e1a605aac72aa3R194

In order to support `balance_of(Non-existent address) = 0`, the `ExitSucceed::Stopped` have been added. But this is also the case when the contract is empty, so it must not be empty.

xlc: ok I think it is fine to require output, but can you also require output is true

xlc (approved):

ermalkaleci (approved):

PR #946, opened 2021-04-21
Weight meter

brettkolodny:


brettkolodny:
To run:

`cargo bench --features=bench | cargo run --bin weight-gen`

PR #969, opened 2021-04-30
replace manual convert with num_enum macro

ermalkaleci:


zjb0807:
https://github.com/PureStake/moonbeam/pull/358/files#diff-8f1282276e25bba33280af309d5461c656fbf5ea732594d757cdfc01b2828335R73

@xlc he suggests use function-selector instead of the enum.

ermalkaleci:
> https://github.com/PureStake/moonbeam/pull/358/files#diff-8f1282276e25bba33280af309d5461c656fbf5ea732594d757cdfc01b2828335R73
> 
> @xlc he suggests use function-selector instead of the enum.

I think this will create confusion especially if method name changes with will generate different selectors

xlc (approved):
This is fine as it is but later I think it is better to switch to use function-selector so that we can avoid a lot of method argument marshalling work and avoid some predeploy contracts. i.e. reduce codes

Once this Acala EVM out on prod, it is public interface and we can never change them, so no need to worry about name changes because there cannot be any changes.

PR #979, opened 2021-05-04
check the range of the user's contract address

zjb0807:
Closes: #964

### DIFF
--- a/modules/evm/src/runner/handler.rs
+++ b/modules/evm/src/runner/handler.rs
@@ -205,7 +205,7 @@ impl<'vicinity, 'config, T: Config> Handler<'vicinity, 'config, '_, T> {
 		});
 	}
 
-	pub fn create_address(scheme: CreateScheme) -> Result<H160, DispatchError> {
+	pub fn create_address(scheme: CreateScheme) -> Result<H160, ExitError> {
 		let address = match scheme {
 			CreateScheme::Create2 {
 				caller,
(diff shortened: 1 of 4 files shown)


modules/evm/src/runner/handler.rs
```diff
@@ -227,6 +227,14 @@ impl<'vicinity, 'config, T: Config> Handler<'vicinity, 'config, '_, T> {
 				H256::from_slice(Keccak256::digest(&stream.out()).as_slice()).into()
 			}
 			CreateScheme::Fixed(naddress) => naddress,
+		};
+
+		if address.as_bytes().starts_with(&SYSTEM_CONTRACT_ADDRESS_PREFIX) {
+			Err(DispatchError::Other(
```
xlc, line 233: should just change `DispatchError` to `ExitError` to avoid another convert later

xlc (commented):
~some test when it fails?~

it seems like hard to write test to make it fail, so should be fine

modules/evm/src/runner/handler.rs
```diff
@@ -230,9 +230,10 @@ impl<'vicinity, 'config, T: Config> Handler<'vicinity, 'config, '_, T> {
 		};
 
 		if address.as_bytes().starts_with(&SYSTEM_CONTRACT_ADDRESS_PREFIX) {
-			Err(DispatchError::Other(
+			use sp_std::borrow::Cow;
+			Err(ExitError::Other(Cow::Borrowed(
```
xlc, line 234:
```suggestion
			Err(ExitError::Other("contract address conflicts with the system contract".into()))
```

modules/evm/src/runner/mod.rs
```diff
@@ -78,6 +78,7 @@ impl<T: Config> Runner<T> {
 				CreateScheme::Legacy { caller: source }
 			};
 			Handler::<T>::create_address(scheme)
+				.map_err(|_| DispatchError::Other("contract address conflicts with the system contract"))
```
xlc, line 81: add an error to `enum Error` and have `Error::<T>::ConflictContractAddress`

xlc (approved):

Issue #824, opened 2021-03-18
EVM Oracle is not decimal aware


xlc:
Needs some logic to handle token decimal

PR #1004, opened 2021-05-14
Fix oracle price

zjb0807:
Closes: #824 

https://github.com/AcalaNetwork/evm-examples/pull/22


wangjj9219:
The current calculation is expected. I will add more detailed comments.

zjb0807:
The PR fix #824

xlc (changes requested):
I think we shouldn't ExitError for non critical error and not sure if the precision calculation is correct

xlc (approved):

PR #1020, opened 2021-05-20
Upgrade ORML dependencies.

shaunxw:


runtime/karura/src/lib.rs
```diff
@@ -118,7 +118,7 @@ pub const VERSION: RuntimeVersion = RuntimeVersion {
 	spec_name: create_runtime_str!("karura"),
 	impl_name: create_runtime_str!("karura"),
 	authoring_version: 1,
-	spec_version: 1000,
+	spec_version: 1001,
```
xlc, line 121: no need to change this until we have mainnet

xlc (approved):

runtime/karura/src/lib.rs
shaunxw: Oh the change was for running the wasm upgrade test, and I forgot to reset it.

xlc (approved):

PR #1035, opened 2021-05-21
Fix launch

zjb0807:


xlc:
#1038 should also fix some other issues

zjb0807:
```
$ docker run --rm acala/karura-node:latest build-spec --chain=karura-latest --disable-default-bootnode
Trace: Error: Error: ENOENT: no such file or directory, open 'output/karura-latest-2000.json'
```

Fix the output directory

xlc (approved):

PR #1036, opened 2021-05-21
Migrate to Bounded Vec and Derive MaxEncodedLen

### LINKED ISSUE #1034: Migrate collator-selection Module to BoundedVec
#917

brettkolodny:
Closes #1034

### DIFF
--- a/modules/collator-selection/src/lib.rs
+++ b/modules/collator-selection/src/lib.rs
@@ -81,7 +81,7 @@ pub mod pallet {
 		dispatch::DispatchResultWithPostInfo,
 		inherent::Vec,
 		pallet_prelude::*,
-		traits::{Currency, EnsureOrigin, ReservableCurrency},
+		traits::{Currency, EnsureOrigin, MaxEncodedLen, ReservableCurrency},
 		PalletId,
 	};
 	use frame_support::{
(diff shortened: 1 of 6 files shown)


brettkolodny:
There are already two types for `MaxCandidates` and `MaxInvulnerables` that are tied to the two storage items that need to be migrated to `BoundedVec` but those traits are used for benchmarking only.

Should these be left as `Vec` or should two new types be implemented that serve as maximum amount that should ever be possible?

xlc:
We have make them strictly enforced

modules/collator-selection/src/lib.rs
```diff
@@ -249,23 +243,10 @@ pub mod pallet {
 		pub fn set_invulnerables(origin: OriginFor<T>, new: Vec<T::AccountId>) -> DispatchResultWithPostInfo {
 			T::UpdateOrigin::ensure_origin(origin)?;
 			// we trust origin calls, this is just a for more accurate benchmarking
```
xlc, line 245: this comment can go

xlc (changes requested):
I think we still want desired_candidates as a soft limit can be increased without runtime upgrade.

xlc (approved):

PR #1041, opened 2021-05-24
Add BoundedVec and MaxEncodedLen to nominees-election

### LINKED ISSUE #1037: Migrate nominees-election Module to BoundedVec
#917

brettkolodny:
Closes #1037

### DIFF
--- a/modules/nominees-election/src/lib.rs
+++ b/modules/nominees-election/src/lib.rs
@@ -417,7 +417,7 @@ impl<T: Config<I>, I: 'static> Pallet<T, I> {
 		}
 	}
 
-	fn rebalance() -> Result<(), Error<T, I>> {
+	fn rebalance() {
 		let mut voters = Votes::<T, I>::iter().collect::<Vec<(T::NomineeId, Balance)>>();
 
 		voters.sort_by(|a, b| b.1.cmp(&a.1));
@@ -428,11 +428,9 @@ impl<T: Config<I>, I: 'static> Pallet<T, I> {
 			.map(|(nominee, _)| nominee)
 			.collect::<Vec<_>>()
 			.try_into()
-			.map_err(|_| Error::<T, I>::NominateesCountExceeded)?;
+			.expect("Only took from voters");
 
 		Nominees::<T, I>::put(new_nominees);
-
-		Ok(())
 	}
 }
 
@@ -445,12 +443,6 @@ impl<T: Config<I>, I: 'static> NomineesProvider<T::NomineeId> for Pallet<T, I> {
 impl<T: Config<I>, I: 'static> OnNewEra<EraIndex> for Pallet<T, I> {
 	fn on_new_era(era: EraIndex) {
 		CurrentEra::<T, I>::put(era);
-		Self::rebalance().map_or_else(
-			|_| {
-				debug_assert!(false);
-				()
-			},
-			|_| (),
-		);
+		Self::rebalance();
 	}
 }
--- a/modules/nominees-election/src/lib.rs
+++ b/modules/nominees-election/src/lib.rs
@@ -162,7 +162,6 @@ pub mod module {
 		#[pallet::constant]
 		type MaxUnlockingChunks: Get<u32>;
 		type RelaychainValidatorFilter: Contains<Self::NomineeId>;
-		type MaxNominees: Get<u32>;
 	}
 
 	#[pallet::error]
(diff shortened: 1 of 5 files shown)


modules/nominees-election/src/lib.rs
```diff
@@ -132,16 +162,18 @@ pub mod module {
 		#[pallet::constant]
 		type MaxUnlockingChunks: Get<u32>;
 		type RelaychainValidatorFilter: Contains<Self::NomineeId>;
+		type MaxNominees: Get<u32>;
```
xlc, line 165:
```suggestion
		#[pallet::constant]
		type MaxNominees: Get<u32>;
```

xlc: can we just use `NominateesCount` instead of `MaxNominees`?

modules/nominees-election/src/lib.rs
```diff
@@ -156,16 +188,23 @@ pub mod module {
...
-		StorageMap<_, Twox64Concat, T::AccountId, Vec<<T as Config<I>>::NomineeId>, ValueQuery>;
+	pub type Nominations<T: Config<I>, I: 'static = ()> = StorageMap<
+		_,
+		Twox64Concat,
+		T::AccountId,
+		BoundedVec<<T as Config<I>>::NomineeId, T::NominateesCount>,
+		ValueQuery,
+	>;
+
+	// type Unlocking<T> = BoundedVec<UnlockChunk, <T as Config>::Max>
```
xlc, line 199:
```suggestion
```

xlc (changes requested):

modules/nominees-election/src/lib.rs
brettkolodny: I was originally thinking that we may want the flexibility not to have them connected but after thinking it through I think it make sense to have them be the same.

modules/nominees-election/src/lib.rs
```diff
@@ -79,7 +91,9 @@ impl BondingLedger {
 					false
 				}
 			})
-			.collect();
+			.collect::<Vec<UnlockChunk>>()
+			.try_into()
+			.expect("Maximum UnlockChunks exceeded");
```
xlc, line 96: The message in expect should be a proof of why this can never fail. In this case it is because filter can only reduce number of elements.

modules/nominees-election/src/lib.rs
```diff
@@ -110,10 +124,26 @@ impl BondingLedger {
 			}
 		}
 
+		self.unlocking = inner_vec.try_into().expect("Maximum UnlockChunks exceeded");
```
xlc, line 127: only pop elements from inner_vec

modules/nominees-election/src/lib.rs
xlc: or you should just use `retain`

modules/nominees-election/src/lib.rs
```diff
@@ -110,10 +124,26 @@ impl BondingLedger {
...
 		self
 	}
 }
 
+impl<T> Default for BondingLedger<T>
+where
+	T: Get<u32>,
+{
+	fn default() -> Self {
+		let unlocking: BoundedVec<UnlockChunk, T> = vec![].try_into().expect("Maximum UnlockChunks exceeded");
```
xlc, line 137:
```suggestion
		let unlocking: BoundedVec<UnlockChunk, T> = Default::default();
```

modules/nominees-election/src/lib.rs
```diff
@@ -383,30 +426,40 @@ impl<T: Config<I>, I: 'static> Pallet<T, I> {
 		}
 	}
 
-	fn rebalance() {
+	fn rebalance() -> Result<(), Error<T, I>> {
```
xlc, line 429:
```suggestion
	fn rebalance() {
```

this shouldn't fail

modules/nominees-election/src/lib.rs
```diff
@@ -383,30 +426,40 @@ impl<T: Config<I>, I: 'static> Pallet<T, I> {
...
 
-		let new_nominees = voters
+		let new_nominees: BoundedVec<<T as Config<I>>::NomineeId, <T as Config<I>>::NominateesCount> = voters
 			.into_iter()
 			.take(T::NominateesCount::get().saturated_into())
 			.map(|(nominee, _)| nominee)
-			.collect::<Vec<_>>();
+			.collect::<Vec<_>>()
+			.try_into()
+			.map_err(|_| Error::<T, I>::NominateesCountExceeded)?;
```
xlc, line 440: we can have `expect` here as the max numbers of elements is `NominateesCount` guaranteed by `take`

xlc (changes requested):

xlc (approved):

Issue #1044, opened 2021-05-25
Migrate polkadot-bridge to BoundedVec and derive MaxEncodedLen


brettkolodny:
#917

PR #1045, opened 2021-05-25
Migrated polkadot-bridge Storage to BoundedVec

brettkolodny:
#1044

### DIFF
--- a/modules/polkadot-bridge/src/lib.rs
+++ b/modules/polkadot-bridge/src/lib.rs
@@ -36,26 +36,15 @@ use support::{
 pub use module::*;
 
 /// The params related to rebalance per era
-#[derive(Encode, Decode, Clone, RuntimeDebug, PartialEq, Eq)]
-pub struct SubAccountStatus<T: Config> {
+#[derive(Encode, Decode, Clone, RuntimeDebug, PartialEq, Eq, Default)]
+pub struct SubAccountStatus<Unbonding> {
 	/// Bonded amount
 	pub bonded: Balance,
 	/// Free amount
 	pub available: Balance,
 	/// Unbonding list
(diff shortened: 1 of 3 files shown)


modules/polkadot-bridge/src/lib.rs
```diff
@@ -36,17 +36,28 @@ use support::{
...
 	pub bonded: Balance,
 	/// Free amount
 	pub available: Balance,
 	/// Unbonding list
-	pub unbonding: Vec<(EraIndex, Balance)>,
+	pub unbonding: BoundedVec<(EraIndex, Balance), T::MaxUnbonding>,
 	pub mock_reward_rate: Rate,
 }
 
+impl<T: Config> Default for SubAccountStatus<T> {
```
xlc, line 50: the derived Default doesn't work?

brettkolodny: No, it fails due to unsatisfied trait bounds

xlc: Right. The issue is it will also add `T: Default` which doesn't make sense. The workaround is avoid pass `Config` directly but pass `MaxUnbonding` instead

modules/polkadot-bridge/src/lib.rs
```diff
@@ -36,17 +36,28 @@ use support::{
...
+	fn default() -> Self {
+		Self {
+			bonded: Balance::default(),
+			available: Balance::default(),
+			mock_reward_rate: Rate::default(),
+			unbonding: vec![].try_into().expect("Exceeded MaxUnbonding"),
```
xlc, line 56: should just use default

xlc (approved):

PR #1049, opened 2021-05-26
Fix collator selection

zjb0807:
https://github.com/AcalaNetwork/Acala/blob/84ad69eea4479a9538a6a84b5a4b386579606e00/modules/collator-selection/src/lib.rs#L361

The kick_threshold time is too short cause the candidates are cleared.


zjb0807:
https://github.com/AcalaNetwork/Acala/blob/6a7b2f78be123000e62e06ddae46efa77681ef59/modules/collator-selection/src/lib.rs#L296

```
last_block: frame_system::Pallet::<T>::block_number() + T::KickThreshold::get(), 
```
Maybe change this is better?

modules/collator-selection/src/lib.rs
```diff
@@ -406,30 +397,83 @@ pub mod pallet {
 	/// Play the role of the session manager.
 	impl<T: Config> SessionManager<T::AccountId> for Pallet<T> {
 		fn new_session(index: SessionIndex) -> Option<Vec<T::AccountId>> {
-			log::info!(
-				"assembling new collators for new session {} at #{:?}",
+			let candidates = Self::candidates()
+				.iter()
+				.map(|candidate| candidate.who.clone())
```
xlc, line 402:
```suggestion
				.into_iter()
				.map(|candidate| candidate.who)
```

I think this can avoid the clone

modules/collator-selection/src/lib.rs
```diff
@@ -294,6 +296,7 @@ pub mod pallet {
 			let incoming = CandidateInfo {
 				who: who.clone(),
 				deposit,
+				validator: false,
```
xlc, line 299: should avoid the word `validator` because parachain is using collator

modules/collator-selection/src/lib.rs
```diff
@@ -386,7 +365,19 @@ pub mod pallet {
 			let candidates_len = <Candidates<T>>::mutate(|candidates| -> usize {
 				let mut candidates_inner = candidates.to_vec();
 				if let Some(found) = candidates_inner.iter_mut().find(|candidate| candidate.who == author) {
+					log::debug!(
+						"note author {:?} authored a block at #{:?}",
+						author,
+						<frame_system::Pallet<T>>::block_number(),
+					);
 					found.last_block = frame_system::Pallet::<T>::block_number();
```
xlc, line 373: I think we don't nee to track `last_block` anymore

modules/collator-selection/src/lib.rs
```diff
@@ -386,7 +365,19 @@ pub mod pallet {
 			let candidates_len = <Candidates<T>>::mutate(|candidates| -> usize {
 				let mut candidates_inner = candidates.to_vec();
 				if let Some(found) = candidates_inner.iter_mut().find(|candidate| candidate.who == author) {
```
xlc, line 367: I think we can just check if SessionPoints have this author

modules/collator-selection/src/lib.rs
```diff
@@ -406,30 +397,83 @@ pub mod pallet {
...
 		}
-		fn start_session(_: SessionIndex) {
-			// we don't care.
+
+		fn start_session(index: SessionIndex) {
+			<Candidates<T>>::mutate(|candidates| {
+				let mut candidates_inner = candidates.to_vec();
+				candidates_inner.iter_mut().for_each(|candidate| {
+					if candidate.validator {
+						<SessionPoints<T>>::insert(&candidate.who, 0);
```
xlc, line 427: I think we need to do this for all new authors. otherwise in `end_session`, one that never produce block will not in `SessionPoints`

modules/collator-selection/src/lib.rs
```diff
@@ -406,30 +397,83 @@ pub mod pallet {
...
+			<SessionPoints<T>>::iter().for_each(|(who, point)| {
+				if point == 0 {
+					log::debug!(
+						"end session {:?} at #{:?}, remove candidate: {:?}",
+						index,
+						<frame_system::Pallet<T>>::block_number(),
+						who,
+					);
+
+					let outcome = Self::try_remove_candidate(&who);
```
xlc, line 462: better to collect all the authors needs remove, and remove all of them in one go

xlc (changes requested):

modules/collator-selection/src/lib.rs
zjb0807: If new authors in `SessionPoints`, it will be removed in `end_session`.
Here delay one session.

xlc: we know on which session index the candidate will become author, so we can insert all new authors

modules/collator-selection/src/lib.rs
```diff
@@ -294,7 +296,7 @@ pub mod pallet {
 			let incoming = CandidateInfo {
 				who: who.clone(),
 				deposit,
-				last_block: frame_system::Pallet::<T>::block_number(),
+				start_session: T::ValidatorSet::session_index() + 2,
```
xlc, line 299:
```suggestion
                // pallet-session have 1 session delay so this will be applied on 1 session after next session
                // Tracking issue: https://github.com/paritytech/substrate/issues/8650
				start_session: T::ValidatorSet::session_index() + 2,
```

xlc (approved):

Issue #1057, opened 2021-05-31
update collator-selection to avoid hardcoded 1 session delay


xlc:
Instead of magically figure out when a candidate become author, just use `ValidatorSet::validators` to fetch current authors.

PR #1059, opened 2021-06-01
Use ValidatorSet::validators to get authors

zjb0807:
Closes: #1057

### DIFF
--- a/modules/collator-selection/src/lib.rs
+++ b/modules/collator-selection/src/lib.rs
@@ -83,16 +83,21 @@ pub mod pallet {
 		BoundedVec, PalletId,
 	};
 	use frame_support::{
-		sp_runtime::{traits::AccountIdConversion, RuntimeDebug},
+		sp_runtime::{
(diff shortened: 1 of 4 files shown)


modules/collator-selection/src/lib.rs
```diff
@@ -404,26 +404,28 @@ pub mod pallet {
...
-						<SessionPoints<T>>::insert(&candidate.who, 0);
-					}
-				});
-				(candidates.len(), collator)
+			let validators = T::ValidatorSet::validators();
+			let candidates = Self::candidates();
+			let mut collators = vec![];
+
+			candidates.iter().for_each(|candidate| {
+				let who = ValidatorIdOf::<T>::convert(candidate.who.clone())
```
xlc, line 412: I think you can avoid this by having `type ValidatorSet: ValidatorSet<Self::AccountId, ValidatorId = Self::AccountId>;`

xlc (approved):

xlc (approved):

Issue #1067, opened 2021-06-06
collator-selection: MinimumCandidateCount


xlc:
don't remove candidate if current candidate count is this amount

this should ensure we don't have too less candidate and allow to help recover the network incase there is some major bugs on client side

PR #1068, opened 2021-06-06
Only update points of candidates in validators

zjb0807:
Closes: #1067 

Here only remove the candidate in `Candidates` storage.

The `Invulnerables` will not be removed and a minimum number of candidates can be kept.


zjb0807:
@StrawberryFlavor The launch parachains only work 1 day because the `LeasePeriod = 1 * DAYS`
https://github.com/paritytech/polkadot/blob/80256e98925deb8f90115664da142acc3306b235/runtime/rococo/src/lib.rs#L761

xlc (approved):
Some tests will be helpful

PR #1098, opened 2021-06-15
Update orml token

zjb0807:
1. Update orml.

### DIFF
--- a/modules/loans/src/lib.rs
+++ b/modules/loans/src/lib.rs
@@ -90,8 +90,6 @@ pub mod module {
 	#[pallet::error]
 	pub enum Error<T> {
 		AmountConvertFailed,
-		/// Balance is too low.
-		BalanceTooLow,
 	}
 
 	#[pallet::event]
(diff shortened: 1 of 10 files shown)


modules/loans/src/lib.rs
```diff
@@ -194,8 +196,18 @@ impl<T: Config> Pallet<T> {
...
 
 		if collateral_adjustment.is_positive() {
+			ensure!(
+				T::Currency::total_balance(currency_id, who).saturating_sub(collateral_balance_adjustment)
+					>= T::Currency::minimum_balance(currency_id),
+				Error::<T>::BalanceTooLow
+			);
 			T::Currency::transfer(currency_id, who, &module_account, collateral_balance_adjustment)?;
 		} else if collateral_adjustment.is_negative() {
+			ensure!(
```
xlc, line 206: why do we need those checks? we can just let `transfer` to fail?
if user already have other assets that provides provider, we should allow reduce this asset to zero

zjb0807:
Need merge https://github.com/open-web3-stack/open-runtime-module-library/pull/522

xlc (approved):

wangjj9219 (approved):

Issue #859, opened 2021-04-01
Integration test for #858


xlc:
#858

Need to update existence deposit first

PR #1101, opened 2021-06-16
Add evm account integration test

zjb0807:
Closes: #859 

https://github.com/open-web3-stack/open-runtime-module-library/blob/32666fd132119cfb0a6299b10909d903b0e9f5c8/tokens/src/lib.rs#L477

Tokens also add system provider.

### DIFF
--- a/runtime/mandala/src/lib.rs
+++ b/runtime/mandala/src/lib.rs
@@ -54,9 +54,7 @@ use module_evm_accounts::EvmAddressMapping;
 pub use module_evm_manager::EvmCurrencyIdMapping;
 use module_transaction_payment::{Multiplier, TargetedFeeAdjustment};
 use orml_tokens::CurrencyAdapter;
-use orml_traits::{
-	create_median_value_data_provider, parameter_type_with_key, DataFeeder, DataProviderExtended, Handler,
-};
+use orml_traits::{create_median_value_data_provider, parameter_type_with_key, DataFeeder, DataProviderExtended};
 use pallet_transaction_payment::{FeeDetails, RuntimeDispatchInfo};
 use sp_api::impl_runtime_apis;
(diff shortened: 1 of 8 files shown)


runtime/mandala/src/lib.rs
```diff
@@ -1146,9 +1147,9 @@ impl module_transaction_payment::Config for Runtime {
 pub struct EvmAccountsOnClaimHandler;
 impl Handler<AccountId> for EvmAccountsOnClaimHandler {
 	fn handle(who: &AccountId) -> DispatchResult {
-		if System::providers(who) == 0 {
-			// no provider. i.e. no native tokens
-			// ensure there are some native tokens, which will add provider
+		if Balances::total_balance(who).is_zero() {
```
xlc, line 1150: If user have other tokens (e.g. DOT), we don't need to do this

zjb0807: Delete  the line `if Balances::total_balance(who).is_zero() {` 
 and all cases do `TransactionPayment::ensure_can_charge_fee` ?

xlc: I mean the old code should be fine

zjb0807: https://github.com/open-web3-stack/open-runtime-module-library/blob/32666fd132119cfb0a6299b10909d903b0e9f5c8/tokens/src/lib.rs#L477

If user have some tokens (e.g. ACA or DOT), `providers(who) = 1` and never 
 do `TransactionPayment::ensure_can_charge_fee`.

if user doesn't have any tokens, `providers(who) = 0` and do `TransactionPayment::ensure_can_charge_fee` doesn't make any sense.

xlc (approved):

PR #1104, opened 2021-06-17
The return of the xcm

shaunxw:


xlc (approved):

runtime/karura/src/lib.rs
```diff
@@ -1288,166 +1291,223 @@ impl module_evm_bridge::Config for Runtime {
...
+	IsNativeConcrete<CurrencyId, CurrencyIdConvert>,
+	AccountId,
+	LocationToAccountId,
+	CurrencyId,
+	CurrencyIdConvert,
+>;
+
+//TODO: use token registry currency type encoding
+fn native_currency_location(id: CurrencyId) -> MultiLocation {
+	X3(Parent, Parachain(ParachainInfo::get().into()), GeneralKey(id.encode()))
```
apopiak, line 1436: As `Multilocation`s are meant to be relative to the context. Why not just use `X1(GeneralKey(id.encode()))`?

runtime/mandala/src/lib.rs
```diff
@@ -1556,166 +1559,222 @@ impl ecosystem_chainsafe::Config for Runtime {
...
+	type XcmpMessageHandler = XcmpQueue;
+	type ReservedXcmpWeight = ReservedXcmpWeight;
 }
 
 impl parachain_info::Config for Runtime {}
 
-impl cumulus_pallet_aura_ext::Config for Runtime {}
+parameter_types! {
+	pub const DotLocation: MultiLocation = MultiLocation::X1(Parent);
+	pub const RelayNetwork: NetworkId = NetworkId::Polkadot;
```
apopiak, line 1581: Not sure where your testnets are deployed, but ff you want to deploy this to Westend the `RelayNetwork` should be `NetworkId::Named(b"Westend".to_vec())` I guess.

runtime/karura/src/lib.rs
```diff
@@ -1288,166 +1291,223 @@ impl module_evm_bridge::Config for Runtime {
...
+	type Trader = FixedRateOfConcreteFungible<KsmPerSecond, ()>;
+	type ResponseHandler = (); // Don't handle responses for now.
+}
+
+parameter_types! {
+	pub MaxDownwardMessageWeight: Weight = RuntimeBlockWeights::get().max_block / 10;
+}
+
+/// No local origins on this chain are allowed to dispatch XCM sends/executions.
+pub type LocalOriginToLocation = (SignedToAccountId32<Origin, AccountId, RelayNetwork>,);
```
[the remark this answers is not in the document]
shaunxw: I think you're right. I used the same type statemine and didn't double check the comments https://github.com/paritytech/cumulus/blob/1bb5e2ae4ba91647e2c26b5a820211a7274ed53a/polkadot-parachains/statemine-runtime/src/lib.rs#L571-L572

runtime/karura/src/lib.rs
shaunxw: We plan to have a standard way(the *token registry* in the code comment), to represent all local and foreign token ids in the network. This would make it easier for all parachain projects to identify and transfer tokens. The scheme for parachain tokens will be `(Parent, Parachain(id), GeneralKey)`.

PR #1105, opened 2021-06-17
Fix sudo key

zjb0807:


xlc (changes requested):
Interesting because somehow the old code seems working?
But I think the better fix is update the config.yml format to `sudo: alice`

xlc (approved):

Issue #1093, opened 2021-06-14
script to verify bootnodes are all good


xlc:
For each bootnode, run something like this and verify the node is able to connect to 1 peer

`docker run --rm acala/karura-node:1.0.1 --chain=karura --reserved-only --reserved-nodes=/dns/karura-bootnode-2.aca-api.network/tcp/30333/p2p/12D3KooWMosCgKRA18CVozYg3QhWvEEV85iqAaN7FcENavgUs4vm -- --chain=dev`

PR #1127, opened 2021-06-24
Add script to verify bootnodes

zjb0807:
Closes: #1093

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -140,3 +140,7 @@ srtool-build-wasm-karura:
 .PHONY: generate-tokens
 generate-tokens:
 	./scripts/generate-tokens-and-predeploy-contracts.sh
+
+.PHONY: verify-bootnodes-karura
+verify-bootnodes-karura:
+	./scripts/verify-bootnodes.sh karura
--- a/Makefile
+++ b/Makefile
@@ -140,7 +140,3 @@ srtool-build-wasm-karura:
 .PHONY: generate-tokens
 generate-tokens:
 	./scripts/generate-tokens-and-predeploy-contracts.sh
-
-.PHONY: verify-bootnodes-karura
-verify-bootnodes-karura:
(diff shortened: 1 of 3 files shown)


scripts/verify-bootnodes.sh
```diff
@@ -0,0 +1,59 @@
...
+elif [ $1 = "acala" ]; then
+	IMAGE=acala/acala-node:1.0.1
+	CHAIN_NAME=acala
+	GENESIS=
+else
+	echo "not support $1"
+	exit 1
+fi
+
+docker rm -f bootnotes > /dev/null 2>&1 || true
```
xlc, line 18: why need this?

scripts/verify-bootnodes.sh
```diff
@@ -0,0 +1,59 @@
+#!/usr/bin/env bash
+
+set -e
+
+if [ $1 = "karura" ]; then
+	IMAGE=acala/karura-node:1.0.1
```
xlc, line 6:
```suggestion
	IMAGE=acala/karura-node:latest
```

Makefile
```diff
@@ -140,3 +140,7 @@ srtool-build-wasm-karura:
 .PHONY: generate-tokens
 generate-tokens:
 	./scripts/generate-tokens-and-predeploy-contracts.sh
+
+.PHONY: verify-bootnodes-karura
+verify-bootnodes-karura:
```
xlc, line 145: I don't think we need this in makefile, developer don't need to run this

xlc (commented):
can you also setup a Github action to run this every week?

scripts/verify-bootnodes.sh
zjb0807: Deal with the situation of `ctrl+c` in local testing. 
Will not affect the process.

xlc (approved):

Issue #1131, opened 2021-06-26
More collator-selection improvements


xlc:
- New Config `CollatorKickThreshold: Get<Permill>`
- Update kick collator to:
  - `total_session_point = blocks_per_session * point_per_block`
  - `average_session_point = total_session_point / collator_count`
  - `required_point = average_session_point * CollatorKickThreshold`
  - `if session_point(collator_id) < required_point { kick_collator(collator_id) }`
- Check how long it take to fully kick a collator. I think author change have 2 session delay so a kick will only applies after 2 session?
- new `const POINT_PER_BLOCK = 10`

PR #1139, opened 2021-06-28
Improve collator selection

zjb0807:
Closes: #1131

### DIFF
--- a/modules/collator-selection/src/lib.rs
+++ b/modules/collator-selection/src/lib.rs
@@ -146,10 +146,6 @@ pub mod pallet {
 		#[pallet::constant]
 		type MaxInvulnerables: Get<u32>;
 
-		/// Blocks number per session.
-		#[pallet::constant]
-		type BlocksPerSession: Get<Self::BlockNumber>;
-
 		/// Will be kicked if block is not produced in threshold.
 		#[pallet::constant]
 		type CollatorKickThreshold: Get<Permill>;
@@ -385,6 +381,7 @@ pub mod pallet {
 	{
 		fn note_author(author: T::AccountId) {
 			log::debug!(
+				target: "collator-selection",
 				"note author {:?} authored a block at #{:?}",
 				author,
 				<frame_system::Pallet<T>>::block_number(),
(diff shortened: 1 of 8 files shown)


zjb0807:
will test it with launch

modules/collator-selection/src/lib.rs
```diff
@@ -437,15 +449,22 @@ pub mod pallet {
...
 		fn end_session(index: SessionIndex) {
-			let mut candidates_len = 0;
 			let mut removed_len = 0;
-			<SessionPoints<T>>::drain().for_each(|(who, point)| {
-				if point == 0 {
+			let session_points = <SessionPoints<T>>::drain().collect::<Vec<_>>();
+			let candidates_len: u32 = session_points.len() as u32;
+			for (who, point) in session_points {
+				let blocks_per_session: u32 = TryInto::try_into(T::BlocksPerSession::get()).unwrap_or_default();
+				let total_session_point: u32 = blocks_per_session.checked_mul(POINT_PER_BLOCK).unwrap_or_default();
```
xlc, line 457: Should sum all the points. Later we may have some other conditions to add more or less points. e.g. include some unsigned tx via offchain worker can give more point or failed to include it may result less reward points.

modules/collator-selection/src/lib.rs
```diff
@@ -437,15 +449,22 @@ pub mod pallet {
...
+				let average_session_point: u32 = total_session_point.checked_div(candidates_len).unwrap_or_default();
+				let required_point: u32 = T::CollatorKickThreshold::get().mul_floor(average_session_point);
+				if point < required_point {
 					log::debug!(
-						"end session {:?} at #{:?}, remove candidate: {:?}",
+						"end session {:?} at #{:?}, remove candidate: {:?}, point: {:?}, required_point: {:?}",
```
xlc, line 462: should add `target: "collator-selection"` to all the loggings

modules/collator-selection/src/lib.rs
```diff
@@ -437,29 +448,42 @@ pub mod pallet {
...
+
+			let total_session_point: u32 = session_points.iter().fold(0, |mut sum, (_, point)| {
+				sum += point;
+				sum
+			});
+			let average_session_point: u32 = total_session_point.checked_div(candidates_len).unwrap_or_default();
+			let required_point: u32 = T::CollatorKickThreshold::get().mul_floor(average_session_point);
+			for (who, point) in session_points {
+				// required_point maybe is zero
+				if point < required_point || point.is_zero() {
```
xlc, line 463:
```suggestion
				if point <= required_point {
```

I think this is better? We already taking a floor for required point.

xlc (approved):

PR #1140, opened 2021-06-28
calculate fair price for lp token

wangjj9219:
according to https://blog.alphafinance.io/fair-lp-token-pricing/ , the fair price is determined by the external feed price(both token_0 and token_1) and the invariant(pool_0 * pool_1) of the liquidity pool to prevent the manipulatable attack .

### DIFF
--- a/modules/prices/src/lib.rs
+++ b/modules/prices/src/lib.rs
@@ -219,6 +219,7 @@ impl<T: Config> PriceProvider<CurrencyId> for Pallet<T> {
 }
 
 /// https://en.wikipedia.org/wiki/Integer_square_root
+/// TODO: use https://github.com/paritytech/parity-common/pull/554 after it released
 fn integer_sqrt(n: U256) -> U256 {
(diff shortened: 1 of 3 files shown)


modules/prices/src/lib.rs
```diff
@@ -219,3 +217,40 @@ impl<T: Config> PriceProvider<CurrencyId> for Pallet<T> {
 		<Pallet<T>>::deposit_event(Event::UnlockPrice(currency_id));
 	}
 }
+
+/// https://en.wikipedia.org/wiki/Integer_square_root
+fn integer_sqrt(n: U256) -> U256 {
```
xlc, line 222: add TODO to use https://github.com/paritytech/parity-common/pull/554 once it is available

xlc (approved):

wangjj9219:
@qwer951123 @RomeroYang @StrawberryFlavor  need to update sdk and front end

PR #1152, opened 2021-07-01
manually convert provisioning trading pair to enable

wangjj9219:


modules/dex/src/lib.rs
```diff
@@ -366,28 +375,84 @@ pub mod module {
...
+						.saturating_add(exchange_rate_1.saturating_mul_int(contribution_1));
+					T::Currency::transfer(
+						trading_pair.get_dex_share_currency_id().expect("shouldn't be invalid!"),
+						&Self::account_id(),
+						&who,
+						share_to_claim,
+					)?;
+
+					// decrease ref count
+					frame_system::Pallet::<T>::dec_consumers(&who);
```
shaunxw, line 451: Maybe I missed something, but cannot find a pairing increasing ref count.

wangjj9219: here: https://github.com/AcalaNetwork/Acala/blob/abfc14d474c038e8a58eb806a74c6fdd7e392a17/modules/dex/src/lib.rs#L763

modules/dex/src/lib.rs
```diff
@@ -494,69 +559,159 @@ pub mod module {
...
-						T::Currency::transfer(trading_pair.1, &module_account_id, &who, contribution.1)?;
+				TradingPairStatus::<_, _>::Provisioning(provisioning_parameters) => {
+					let (total_provision_0, total_provision_1) = provisioning_parameters.accumulated_provision;
+
+					if total_provision_0.is_zero() && total_provision_1.is_zero() {
+						// if these's no provision, enable trading pair directly
+						TradingPairStatuses::<T>::insert(trading_pair, TradingPairStatus::Enabled);
+						Self::deposit_event(Event::EnableTradingPair(trading_pair));
+					} else if frame_system::Pallet::<T>::block_number() >= provisioning_parameters.not_before
+						&& (total_provision_0 >= provisioning_parameters.target_provision.0
```
xlc, line 637: I think we should add another condition `total_provision_0 > 0 && total_provision_1 > 0`

xlc (commented):
all the `trading_pair.get_dex_share_currency_id().expect("shouldn't be invalid!");` looks strange. why should `TradingPair::get_dex_share_currency_id` ever return `None`?

I think some refactoring can be done on `TradingPair` to improve the types

We need two version of enable trading pair

- Only callable by ListingOrigin, can enable trading pair unconditionally
- Callable by any signed user, can only enable if meet the condition

wangjj9219:
I will refactor TradingPair in next PR

wangjj9219:
> all the `trading_pair.get_dex_share_currency_id().expect("shouldn't be invalid!");` looks strange. why should `TradingPair::get_dex_share_currency_id` ever return `None`?
> 
> I think some refactoring can be done on `TradingPair` to improve the types
> 
> We need two version of enable trading pair
> 
> * Only callable by ListingOrigin, can enable trading pair unconditionally
> * Callable by any signed user, can only enable if meet the condition

I add `end_provisioning` for signed user to enable provisioning trading pair if meet the condition

xlc (approved):

xlc:
This changes extrinsic index but we are not lived and don't support any hardware wallets yet so it is fine. But remember later we have to avoid reorder the extrinsic index of that in published runtime.

shaunxw (approved):

Issue #1000, opened 2021-05-13
Generate precompile function selector with macro


xlc:
IMO actions variant values should be generated automatically in the future improvement. Manual adding them could be error-prune.

_Originally posted by @shaunxw in https://github.com/AcalaNetwork/Acala/pull/982#discussion_r631730511_

PR #1165, opened 2021-07-03
Add create_function_selector macro

zjb0807:
Closes: #1000


primitives/src/evm.rs
```diff
@@ -85,3 +85,50 @@ pub struct EstimateResourcesRequest {
...
+				// write input message
+				hasher.update(s);
+				// read hash digest
+				let result = hasher.finalize();
+				u32::from_be_bytes(result[..4].try_into().unwrap())
+			}
+
+			#[test]
+			fn method_hash_works() {
+				$(if get_function_selector($method) != $val {
```
xlc, line 117:
```
assert_eq!(get_function_selector($method), $val, "method: {:?}", $method)
```

xlc (approved):
If will be even better if can generate sector using procedure macro but that can happen later

shaunxw (approved):

PR #1187, opened 2021-07-08
enable more pallets & update parameters

xlc:


runtime/karura/src/lib.rs
```diff
@@ -210,7 +210,11 @@ impl Filter<Call> for BaseCallFilter {
 			// Collactor Selection
 			Call::CollatorSelection(_) | Call::Session(_) |
 			// Vesting
-			Call::Vesting(_)
+			Call::Vesting(_) |
+			// TransactionPayment
+			Call::TransactionPayment(_) |
+			// XCM
+			Call::XcmpQueue(_) | Call::PolkadotXcm(_) | Call::DmpQueue(_) | Call::XTokens(_)
```
xlc, line 217: Is it OK to enable those? We want to make sure they cannot be used to transfer tokens within Karura

shaunxw (approved):

Issue #1192, opened 2021-07-09
Review the initialization of type FixedU128


wangjj9219:
Because `FixedU128` can only represent multiple between 3.4*10^20 and 0 * 10^-18 , if use constructions like `checked_from_rational` or `saturating_from_rational` to generate a multiple that beyond safe range，will cause the relevant calculations to be unable to get accurate results.

Using `FixedU128` will restrict some business, we need to wait for Substrate to support `FixedU256` or larger Fixed. For now
, review all initialization for type `Price`/`ExchangeRate`/`Rate` (`FixedU128`),  ensure that throw an error when `checked_from_rational` cannot get the result, and avoid using `saturating_from_rational`

PR #1195, opened 2021-07-09
refactor some calculation from saturating to checked

wangjj9219:
related to #1192 

If DEX opens trading pair registration in the future, especially for these ERC20 tokens, many saturation calculations which use `FixedU128` as multiper are actually easy to overflow, which will bring malicious attacks. So it's necessary to modify those saturation calculations to checked calculations.

### DIFF
--- a/modules/cdp-treasury/src/lib.rs
+++ b/modules/cdp-treasury/src/lib.rs
@@ -33,7 +33,7 @@ use frame_system::pallet_prelude::*;
 use orml_traits::{MultiCurrency, MultiCurrencyExtended};
 use primitives::{Balance, CurrencyId};
 use sp_runtime::{
-	traits::{AccountIdConversion, One, Zero},
+	traits::{AccountIdConversion, Bounded, One, Zero},
 	ArithmeticError, DispatchError, DispatchResult, FixedPointNumber,
 };
 use support::{AuctionManager, CDPTreasury, CDPTreasuryExtended, DEXManager, Ratio};
@@ -285,7 +285,7 @@ impl<T: Config> CDPTreasury<T::AccountId> for Pallet<T> {
 
 	fn get_debit_proportion(amount: Self::Balance) -> Ratio {
 		let stable_total_supply = T::Currency::total_issuance(T::GetStableCurrencyId::get());
-		Ratio::checked_from_rational(amount, stable_total_supply).unwrap_or_default()
+		Ratio::checked_from_rational(amount, stable_total_supply).unwrap_or_else(Ratio::max_value)
 	}
 
 	fn on_system_debit(amount: Self::Balance) -> DispatchResult {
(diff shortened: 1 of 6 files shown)


modules/cdp-treasury/src/lib.rs
```diff
@@ -285,7 +285,7 @@ impl<T: Config> CDPTreasury<T::AccountId> for Pallet<T> {
 
 	fn get_debit_proportion(amount: Self::Balance) -> Ratio {
 		let stable_total_supply = T::Currency::total_issuance(T::GetStableCurrencyId::get());
-		Ratio::checked_from_rational(amount, stable_total_supply).unwrap_or_default()
+		Ratio::checked_from_rational(amount, stable_total_supply).unwrap_or_else(Ratio::max_value)
```
xlc, line 288: return zero makes more sense here

modules/staking-pool/src/lib.rs
```diff
@@ -113,13 +113,15 @@ impl Ledger {
 
 	/// The ratio of `free_pool` in `total_belong_to_liquid_holders`.
 	fn free_pool_ratio(&self) -> Ratio {
-		Ratio::checked_from_rational(self.free_pool, self.total_belong_to_liquid_holders()).unwrap_or_default()
+		Ratio::checked_from_rational(self.free_pool, self.total_belong_to_liquid_holders())
```
xlc, line 116: I think it is ok to return zero here?

modules/staking-pool/src/lib.rs
```diff
@@ -113,13 +113,15 @@ impl Ledger {
...
+			.unwrap_or_else(Ratio::max_value)
 	}
 
 	/// The ratio of `unbonding_to_free` in
 	/// `total_belong_to_liquid_holders`.
 	fn unbonding_to_free_ratio(&self) -> Ratio {
-		Ratio::checked_from_rational(self.unbonding_to_free, self.total_belong_to_liquid_holders()).unwrap_or_default()
+		Ratio::checked_from_rational(self.unbonding_to_free, self.total_belong_to_liquid_holders())
```
xlc, line 123: I think it is ok to return zero here?

modules/cdp-treasury/src/lib.rs
wangjj9219: yup, i think too much

modules/dex/src/lib.rs
```diff
@@ -835,6 +852,11 @@ impl<T: Config> Pallet<T> {
 			Error::<T>::MustBeEnabled,
 		);
 
+		ensure!(
+			!max_amount_a.is_zero() && !max_amount_b.is_zero(),
+			Error::<T>::InvalidLiquidityIncrement
+		);
```
shaunxw, line 858: Do we need new unit tests to catch this error?

shaunxw (approved):
LGTM

PR #1204, opened 2021-07-13
ensure account of native token keeps alive after charge fee

wangjj9219:
charge fee maybe fail even if the account has sufficient tokens for these reasons:

1. https://github.com/AcalaNetwork/Acala/blob/ae1dba93636e7536ca8f49cfc80539460383751a/modules/transaction-payment/src/lib.rs#L676 the account must keep alive after charge fee.

2. https://github.com/AcalaNetwork/Acala/blob/ae1dba93636e7536ca8f49cfc80539460383751a/modules/transaction-payment/src/lib.rs#L593  The `balances::transfer` will executed when swap native token and `transfer` requires the receiver is alive, otherwise the `swap` will fail.

So fix `ensure_can_charge_fee` to guarantee account will keep alive.

### DIFF
--- a/modules/transaction-payment/src/lib.rs
+++ b/modules/transaction-payment/src/lib.rs
@@ -567,17 +567,24 @@ where
 		charge_fee_order.dedup();
 
 		let price_impact_limit = Some(T::MaxSlippageSwapWithDEX::get());
+		let native_existential_deposit = <T as Config>::Currency::minimum_balance();
+		let total_native = <T as Config>::Currency::total_balance(who);
+		// add the gap amount to keep account alive and have enough fee
+		let fee_and_alive_gap = if total_native < native_existential_deposit {
+			fee.saturating_add(native_existential_deposit.saturating_sub(total_native))
+		} else {
+			fee
+		};
 
 		// iterator charge fee order to get enough fee
 		for currency_id in charge_fee_order {
 			if currency_id == native_currency_id {
 				// check native balance if is enough
-				let native_is_enough =
-					<T as Config>::Currency::free_balance(who)
-						.checked_sub(&fee)
-						.map_or(false, |new_free_balance| {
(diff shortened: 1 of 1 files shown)


xlc (approved):
better to have some tests

modules/transaction-payment/src/lib.rs
```diff
@@ -567,17 +567,24 @@ where
...
-				let native_is_enough =
-					<T as Config>::Currency::free_balance(who)
-						.checked_sub(&fee)
-						.map_or(false, |new_free_balance| {
-							<T as Config>::Currency::ensure_can_withdraw(who, fee, reason, new_free_balance).is_ok()
-						});
+				let native_is_enough = <T as Config>::Currency::free_balance(who)
+					.checked_sub(&fee_and_alive_gap)
+					.map_or(false, |new_free_balance| {
+						<T as Config>::Currency::ensure_can_withdraw(who, fee, reason, new_free_balance).is_ok()
```
zjb0807, line 586:
```suggestion
						<T as Config>::Currency::ensure_can_withdraw(who, fee, reason, fee_and_alive_gap).is_ok()
```

zjb0807: @xlc The old code seems to be wrong.

xlc: https://github.com/paritytech/substrate/blob/993907f9deb180795b291ba752b9c204165d74dc/frame/support/src/traits/tokens/currency.rs#L103

the last parameter is expected new balance after withdraw

zjb0807: Ok, It's different from tokens.
https://github.com/open-web3-stack/open-runtime-module-library/blob/3d64077e5ac5539bde336a794308d85f6f5f2b82/tokens/src/lib.rs#L642

xlc: Yes we removed the last parameter

PR #1212, opened 2021-07-15
add coverage actions

zjb0807:
Closes: #811

### DIFF
--- a/.github/workflows/coverage.yml
+++ b/.github/workflows/coverage.yml
@@ -5,12 +5,19 @@ on:
   pull_request:
     branches:
       - master
+  push:
+    branches:
+      - master
 
 jobs:
   test:
     name: coverage
     runs-on: ubuntu-latest
     steps:
+      - name: Cancel Previous Runs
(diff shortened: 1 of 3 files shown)


zjb0807:
Need add `codecov/codecov-action@v1` and `actions/cache@v2`

.github/workflows/coverage.yml
```diff
@@ -0,0 +1,51 @@
+name: coverage
+
+on:
+  workflow_dispatch:
+  schedule:
+    - cron: "00 06 * * 1" # 6AM weekly on monday
```
xlc, line 6: we want to run code coverage for every pr & master, just like unit tests

zjb0807: This needs to use self-host.  Running the code coverage for more than 1 hour.

xlc: ok. I can setup another self-host server

xlc: also we don't need to wait for coverage result before merge PR

xlc (approved):

PR #1221, opened 2021-07-16
test coverage in self-hosted

zjb0807:


zjb0807:
It cost more than 30min.

PR #1222, opened 2021-07-16
Rename GH actions.

shaunxw:
To make it more explicit.


xlc (approved):

.github/workflows/test.yml
```diff
@@ -9,7 +9,7 @@ on:
     - master
 
 jobs:
-  cancel:
+  clean-up-actions:
```
zjb0807, line 12:
```
25:    needs: clean-up-actions
```

zjb0807: line 25

zjb0807 (approved):

PR #1229, opened 2021-07-19
deduplicate benchmarks

ermalkaleci:
Move benchmarks into separate package. Benchmarks are wrapped into macros that gets expanded into different runtime benchmarks


modules/nft/Cargo.toml
```diff
@@ -17,6 +17,7 @@ pallet-proxy = { git = "https://github.com/paritytech/substrate", branch = "polk
 primitives = { package = "acala-primitives", path = "../../primitives", default-features = false }
 sp-runtime = { git = "https://github.com/paritytech/substrate", branch = "polkadot-v0.9.8", default-features = false }
 sp-std = { git = "https://github.com/paritytech/substrate", branch = "polkadot-v0.9.8", default-features = false }
+chainbridge = { git = "https://github.com/AcalaNetwork/chainbridge-substrate", default-features = false, optional = true }
```
xlc, line 20: why nft need chainbridge?

runtime/benchmarks/Cargo.toml
```diff
@@ -0,0 +1,34 @@
+[package]
+name = "benchmarks"
+version = "1.1.3"
```
xlc, line 3:
```suggestion
version = "1.2.0"
```

runtime/benchmarks/Cargo.toml
```diff
@@ -0,0 +1,34 @@
...
+authors = ["Acala Developers"]
+edition = "2018"
+
+[dependencies]
+sp-io = { git = "https://github.com/paritytech/substrate", branch = "polkadot-v0.9.7", default-features = false }
```
xlc, line 8:
```suggestion
sp-io = { git = "https://github.com/paritytech/substrate", branch = "polkadot-v0.9.8", default-features = false }
```

shaunxw (commented):
Wrapping all benchmarks in macro would make them difficult to impl and debug.

As the only difference is the imports, we can use different features to control which runtime to import from, similar to how `sp-std` handles imports. In this way the benchmarks don't have to be in macro.

modules/nft/Cargo.toml
ermalkaleci: It's only if you run benchmark tests inside directory that's why it's optional.

ermalkaleci:
> Wrapping all benchmarks in macro would make them difficult to impl and debug.
> 
> 
> 
> As the only difference is the imports, we can use different features to control which runtime to import from, similar to how `sp-std` handles imports. In this way the benchmarks don't have to be in macro.

@shaunxw Not sure how to do that without ending up importing each other

xlc:
so it will be:

karura-runtime -> common-benchmark (features = with-karura-runtime)
mandala-runtime -> common-benchmark (features = with-mandala-runtime)

modules/nft/Cargo.toml
xlc: but that's wrong, it is something else need chainbridge, not nft

ermalkaleci:
> so it will be:
> 
> karura-runtime -> common-benchmark (features = with-karura-runtime)
> mandala-runtime -> common-benchmark (features = with-mandala-runtime)

@xlc how do I import karura-runtime into common-benchmark?

modules/nft/Cargo.toml
ermalkaleci: removed

xlc:
You can use `include!("./common/benchmark.rs")` in karura-runtime to import the file

https://doc.rust-lang.org/std/macro.include.html

ermalkaleci:
I don't see how this will make thing easier. It will be just a directory with rust code which have no purpose unless been included into runtime. Although we will get rid of the wrapper macro

xlc:
You can have the code in mandala-runtime, and include it from karura-runtime.
And get rid of the wrapper macro is already a benefit. It helps IDE.

xlc (approved):

PR #1245, opened 2021-07-22
Update orml

zjb0807:
## ⚠️ Breaking Change  ⚠️ 

Refactor orml-nft to use StorageNMap.

Not compatible with mandala network.

### DIFF
--- a/modules/nft/src/lib.rs
+++ b/modules/nft/src/lib.rs
@@ -395,7 +395,7 @@ impl<T: Config> NFT<T::AccountId> for Pallet<T> {
 	type Balance = NFTBalance;
 
 	fn balance(who: &T::AccountId) -> Self::Balance {
-		orml_nft::TokensByOwner::<T>::iter_prefix(who).count() as u128
+		orml_nft::TokensByOwner::<T>::iter_prefix((who,)).count() as u128
 	}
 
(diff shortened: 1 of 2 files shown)


xlc (approved):

modules/nft/src/lib.rs
```diff
@@ -395,7 +395,7 @@ impl<T: Config> NFT<T::AccountId> for Pallet<T> {
 	type Balance = NFTBalance;
 
 	fn balance(who: &T::AccountId) -> Self::Balance {
-		orml_nft::TokensByOwner::<T>::iter_prefix(who).count() as u128
+		orml_nft::TokensByOwner::<T>::iter_prefix((who,)).count() as u128
```
yrong, line 398: so what is the benifit of nmap? or could we just use child trie instead?

zjb0807: Any prefix key can be iterated
```
orml_nft::TokensByOwner::<Runtime>::iter_prefix((AccountId,))
orml_nft::TokensByOwner::<Runtime>::iter_prefix((AccountId, ClassId))
```

yrong: ok, my understanding is that no matter child trie or double map 2 levels  interation at most while with nmap  no limitation at all

PR #1251, opened 2021-07-23
use fee swap path

wangjj9219:


modules/transaction-payment/src/lib.rs
```diff
@@ -355,19 +364,33 @@ pub mod module {
 				);
 			})
 		}
+
+		// remove after runtime upgraded
+		fn on_runtime_upgrade() -> Weight {
+			DefaultFeeCurrencyId::<T>::remove_all(None);
+			T::DbWeight::get().reads_writes(0, 1)
```
xlc, line 371: we need to check how many items we have currently and change writes.

xlc (approved):

modules/transaction-payment/src/lib.rs
wangjj9219: There are currently 34 items in total, which is an easy mission for upgrades.

modules/transaction-payment/src/lib.rs
```diff
@@ -573,38 +587,46 @@ where
...
-				let trading_path = if currency_id == stable_currency_id {
-					vec![stable_currency_id, native_currency_id]
+		// check native balance if is enough
+		let native_is_enough = <T as Config>::Currency::free_balance(who)
+			.checked_sub(&fee_and_alive_gap)
+			.map_or(false, |new_free_balance| {
+				<T as Config>::Currency::ensure_can_withdraw(who, fee, reason, new_free_balance).is_ok()
+			});
+
+		// native is enough, try swap native to pay fee
```
xlc, line 597:
```suggestion
		// native is not enough, try swap native to pay fee
```

modules/transaction-payment/src/lib.rs
```diff
@@ -573,38 +587,46 @@ where
...
-					&trading_path,
-					fee_and_alive_gap.unique_saturated_into(),
-					<T as Config>::MultiCurrency::free_balance(currency_id, who),
-					price_impact_limit,
-				)
-				.is_ok()
-				{
-					// successfully swap, break iteration
-					break;
+			fee_swap_path_list.dedup();
```
xlc, line 608: dedup only works for sorted vec

modules/transaction-payment/src/lib.rs
```diff
@@ -284,20 +280,33 @@ pub mod module {
...
 	/// DefaultFeeCurrencyId: AccountId => Option<CurrencyId>
 	#[pallet::storage]
 	#[pallet::getter(fn default_fee_currency_id)]
 	pub type DefaultFeeCurrencyId<T: Config> = StorageMap<_, Twox64Concat, T::AccountId, CurrencyId, OptionQuery>;
 
+	/// The alternative fee swap path of accounts.
+	#[pallet::storage]
+	#[pallet::getter(fn alternative_fee_swap_path)]
+	pub type AlternativeFeeSwapPath<T: Config> =
+		StorageMap<_, Twox64Concat, T::AccountId, Vec<CurrencyId>, OptionQuery>;
```
shaunxw, line 308: `Twox64Concat` might not be safe enough for `T::AccountId`?

For swap path, we might need bounded vec instead.

xlc: People can only set swap path for themself so it should be safe.
Can hardcode the limit to 3 elements for swap path.

shaunxw:
How about the `Twox64Concat` hasher used for `T::AccountId`, it should be changed to blake2 ...no?

xlc:
It is safe. Only account owner can set that.

PR #1268, opened 2021-07-28
Fix the check for native token of ensure_can_charge_fee

wangjj9219:


modules/transaction-payment/src/lib.rs
```diff
@@ -587,22 +587,22 @@ where
...
 		let total_native = <T as Config>::Currency::total_balance(who);
-		// add the gap amount to keep account alive and have enough fee
-		let fee_and_alive_gap = if total_native < native_existential_deposit {
-			fee.saturating_add(native_existential_deposit.saturating_sub(total_native))
-		} else {
-			fee
-		};
+		let is_alive_after_charge = fee
+			.saturating_add(native_existential_deposit)
+			.saturating_sub(total_native)
```
xlc, line 592: why not just compare the number? this is a bit confusing

xlc (approved):

PR #1285, opened 2021-08-04
use the slippage for the price from oracle to limit internal swap

wangjj9219:


modules/cdp-engine/src/lib.rs
```diff
@@ -842,16 +840,26 @@ impl<T: Config> Pallet<T> {
 
 		let bad_debt_value = Self::get_debit_value(currency_id, debit);
 		let target_stable_amount = Self::get_liquidation_penalty(currency_id).saturating_mul_acc_int(bad_debt_value);
-
-		// try use collateral to swap enough stable token in DEX when the price impact
-		// is below the limit, otherwise create collateral auctions.
 		let liquidation_strategy = (|| -> Result<LiquidationStrategy, DispatchError> {
-			// swap exact stable with DEX in limit of price impact
+			// calculate the supply limit by slippage limit for the price of oracle,
+			// if oracle price is not avalible, do not limit
```
xlc, line 845: oracle price should always available, because liquidation are triggered by oracle price update
but otherwise I think it is better to fallback to auction instead of dumping into dex.

xlc (approved):

shaunxw (approved):
LGTM

Issue #1281, opened 2021-08-02
New ProxyType


xlc:
- Move ProxyType to common runtime, impl InstanceFilter can stay in each individual runtime
- Governance: similar to Kusama / Polkadot https://github.com/AcalaNetwork/polkadot/blob/4cdcd731d40e592449836e5d958cf1e8ab1ba60a/runtime/kusama/src/lib.rs#L1035-L1044
- Auction: only `auctions.bid`
- Swap: `dex.swap_with_exact_supply`, `dex.swap_with_exact_target`
- Loan: `honzon.adjust_loan`, `honzon.close_loan_has_debit_by_dex`

PR #1288, opened 2021-08-04
Improve ProxyType

zjb0807:
Closes: #1281

### DIFF
--- a/runtime/common/src/lib.rs
+++ b/runtime/common/src/lib.rs
@@ -324,6 +324,9 @@ pub enum ProxyType {
 	Any,
 	CancelProxy,
 	Governance,
+	Auction,
+	Swap,
+	Loan,
 }
 impl Default for ProxyType {
 	fn default() -> Self {
--- a/runtime/common/src/lib.rs
+++ b/runtime/common/src/lib.rs
@@ -20,13 +20,15 @@
 
 #![cfg_attr(not(feature = "std"), no_std)]
 
+use codec::{Decode, Encode};
 use frame_support::{
 	parameter_types,
-	traits::Contains,
+	traits::{Contains, MaxEncodedLen},
 	weights::{
 		constants::{BlockExecutionWeight, ExtrinsicBaseWeight, WEIGHT_PER_MILLIS},
 		DispatchClass, Weight,
 	},
+	RuntimeDebug,
 };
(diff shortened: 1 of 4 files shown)


runtime/karura/src/lib.rs
```diff
@@ -1165,26 +1165,48 @@ parameter_types! {
...
 			ProxyType::CancelProxy => matches!(c, Call::Proxy(pallet_proxy::Call::reject_announcement(..))),
+			ProxyType::Governance => {
+				matches!(
+					c,
+					Call::Authority(..)
+						| Call::Democracy(..) | Call::GeneralCouncil(..)
+						| Call::FinancialCouncil(..)
+						| Call::HomaCouncil(..) | Call::TechnicalCommittee(..)
+						| Call::Treasury(..) | Call::Bounties(..)
+						| Call::Tips(..) | Call::Utility(..)
```
xlc, line 1181: I think we can make `Call::Utility(..)` always allowed no matter type?

xlc (approved):

runtime/common/src/lib.rs
```diff
@@ -316,6 +318,22 @@ pub type EnsureRootOrThreeFourthsTechnicalCommittee = EnsureOneOf<
 	pallet_collective::EnsureProportionAtLeast<_3, _4, AccountId, TechnicalCommitteeInstance>,
 >;
 
+/// The type used to represent the kinds of proxying allowed.
+#[derive(Copy, Clone, Eq, PartialEq, Ord, PartialOrd, Encode, Decode, RuntimeDebug, MaxEncodedLen)]
+pub enum ProxyType {
+	Any,
+	CancelProxy,
+	Governance,
+	Auction,
```
xlc, line 327: acala.js needs update

xlc (approved):

runtime/karura/src/lib.rs
zjb0807:
```
fn filter(&self, c: &Call) -> bool {
    match self {
        _ if matches!(c, Call::Utility(..)) => true,
        ProxyType::Any => true,
        ...
```

It doesn't make sense to do. This will only make `Call::Utility(..)` skip the check, but the execution still fails.(e.g. addProxy `ProxyType::Governance` then use `Utility` to transfer. It will fail in `utility.BatchInterrupted`)

xlc (approved):

Issue #1271, opened 2021-07-28
Ensure tx priority order


xlc:
This reduces some front-running possibility.

Inherent -> Operational tx -> Unsigned tx -> Signed normal tx

We can modify this to adjust priority. Note that final priority is sum of all priority from all SignedExtensions

https://github.com/AcalaNetwork/Acala/blob/e95a7d2848da829a9533129a337e13cb88849072/modules/transaction-payment/src/lib.rs#L723-L737

Inherent are always first and so we don't need to worry about it.

Needs some integration tests to check priority.

PR #1304, opened 2021-08-06
Tx priority order

zjb0807:
Closes: #1271


xlc (approved):
should have more tests but that can happen later

shaunxw (approved):
LGTM.

PR #1307, opened 2021-08-08
Support crosschain transaction between Acala and Manta.

Dengjianping:
1. Add `MA` to `CurrencyId`.
2. Add manta `MultiLocation` to fix `TooExpense` error.
Originally, karura runtime only supports `X1(Parent)` multiLocation, but the multilocation is `X3` while sending transaction to manta parachain. So it will get nothing because this id is `X1`(https://github.com/paritytech/polkadot/blob/release-v0.9.8/xcm/xcm-executor/src/assets.rs#L250).
And this field `fungible` just saves data like:
```rust
// amount is how many tokens you send to parachain.
X3(Parent, Parachain(2084), GeneralKey(CurrencyId::Token(TokenSymbol::MA).encode())) => amount,
// others multilocation => amount
// ...
```

### DIFF
--- a/primitives/src/currency.rs
+++ b/primitives/src/currency.rs
@@ -189,6 +189,9 @@ create_currency_id! {
 
 		// External Ecosystem
 		CASH("Compound CASH", 8) = 140,
+
+		// Manta Parachain
+		MA("Manta", 12) = 5,
 	}
 }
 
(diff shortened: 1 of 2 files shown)


stechu:
@shaunxw @xlc 
The integration is pretty easy to do:
https://hackmd.io/ENLbw3RzRHu6NoIoDSX-LQ

Thanks a lot of the technical help!

xlc:
Stale

PR #1308, opened 2021-08-08
Feature/homa lite improvement

syan095:
User is now able to claim Liquid currency immediately from the Mint action.
Amount minted is: Liquid = (staking - MintFee) * (LiquidTotal / StakingTotal) * (1 - MaxRewards)

Removed all "batch" related code

"issue" is now "set_staking_total_issuance", setting the Total issuance for Staking currency, used to calculate the exchange rate.
Re-benchmarked

WIP: Update the documentation (readme) to keep it up to date


modules/homa-lite/src/lib.rs
```diff
@@ -167,21 +153,65 @@ pub mod module {
...
+
+		/// Updates the cap for how much Staking currency can be used to Mint liquid currency.
+		/// Requires `T::GovernanceOrigin`
+		///
+		/// Parameters:
+		/// - `new_cap`: The new cap for staking currency.
+		#[pallet::weight(< T as Config >::WeightInfo::set_minting_cap())]
+		#[transactional]
+		pub fn set_minting_cap(origin: OriginFor<T>, new_cap: Balance) -> DispatchResult {
+			// This can only be called by Governance or ROOT.
```
shaunxw, line 201:
```suggestion
```

modules/homa-lite/src/lib.rs
```diff
@@ -194,13 +224,31 @@ pub mod module {
...
+				false => {
+					let liquid_total = T::Currency::total_issuance(T::LiquidCurrencyId::get());
+					Ratio::checked_from_rational(liquid_total, staking_total).ok_or(ArithmeticError::Overflow)
+				}
+			}?;
+
+			// liquid_to_mint = ( (staked_amount - MintFee) * liquid_total / staked_total ) * (1 -
+			// MaxRewardPerEra)
+			let mut liquid_to_mint = exchange_rate
+				.checked_mul_int(amount - T::MintFee::get())
```
shaunxw, line 246: Use `checked_sub` with `.expect("proof of won't panic")`

modules/homa-lite/src/lib.rs
```diff
@@ -194,13 +224,31 @@ pub mod module {
...
+				.ok_or(ArithmeticError::Overflow)?;
+
+			liquid_to_mint = liquid_to_mint - T::MaxRewardPerEra::get().mul(liquid_to_mint);
```
shaunxw, line 249: Use `checked_sub`

modules/homa-lite/src/lib.rs
```diff
@@ -194,13 +224,31 @@ pub mod module {
...
 			let staking_currency = T::StakingCurrencyId::get();
 
 			// ensure the user has enough funds on their account.
 			T::Currency::ensure_can_withdraw(staking_currency, &who, amount)?;
 
-			// Cross-chain transfers the staking assets.
+			// Calculate how much Liquid currency is to be minted.
+			// Gets the current exchange rate
+			let staking_total = Self::staking_total_issuance();
+			let exchange_rate = match staking_total.is_zero() {
```
xlc, line 235: `Ratio::checked_from_rational(liquid_total, staking_total).unwrap_or_else(T::DefaultExchangeRate::get)`

modules/homa-lite/src/lib.rs
```diff
@@ -194,13 +224,31 @@ pub mod module {
...
+				true => Ok(T::DefaultExchangeRate::get()),
+				false => {
+					let liquid_total = T::Currency::total_issuance(T::LiquidCurrencyId::get());
+					Ratio::checked_from_rational(liquid_total, staking_total).ok_or(ArithmeticError::Overflow)
+				}
+			}?;
```
shaunxw, line 241: Use `if` instead of matching `true`

xlc (approved):

xlc (changes requested):
Still need some changes

modules/homa-lite/src/lib.rs
syan095: i was wondering in the lines above I've checked to ensure amount > (mintfee + minimum), so this subtraction is guaranteed to succeed - do I still need the checked_sub? or is this still needed for good coding style?

modules/homa-lite/src/lib.rs
syan095: this is brilliant

shaunxw (changes requested):
The new `StakingTotalIssuance` storage shouldn't be added. The duplication with `TotalStakedAmount` storage would lead to fund loss in the cases that `StakingTotalIssuance` was not updated in time.

Using `TotalStakedAmount` alone instead, and change `set_staking_total_issuance` to `set_total_staked_amount`.

modules/homa-lite/src/lib.rs
```diff
@@ -117,90 +102,89 @@ pub mod module {
...
 		#[transactional]
-		pub fn request_mint(origin: OriginFor<T>, amount: Balance, xcm_dest_weight: Weight) -> DispatchResult {
+		pub fn mint(origin: OriginFor<T>, amount: Balance, xcm_dest_weight: Weight) -> DispatchResult {
 			let who = ensure_signed(origin)?;
-
-			// Ensure the amount is above the minimum
+			// Ensure the amount is above the minimum, after the MintFee is deducted.
 			ensure!(
-				amount >= T::MinimumMintThreshold::get(),
+				amount > T::MinimumMintThreshold::get() + T::MintFee::get(),
```
shaunxw, line 148: `.checked_add`

modules/homa-lite/src/lib.rs
```diff
@@ -117,90 +102,89 @@ pub mod module {
...
+			let exchange_rate =
+				Ratio::checked_from_rational(liquid_total, staking_total).unwrap_or_else(T::DefaultExchangeRate::get);
+
+			// liquid_to_mint = ( (staked_amount - MintFee) * liquid_total / staked_total ) * (1 -
+			// MaxRewardPerEra)
+			let mut liquid_to_mint = exchange_rate
+				.checked_mul_int(
+					amount
+						.checked_sub(T::MintFee::get())
+						.expect("Arithmatic should not panic"),
```
shaunxw, line 179: This is *not a proof*. It should like "Amount is ensured greater than T::MintFee; qed"

modules/homa-lite/src/lib.rs
```diff
@@ -117,90 +102,89 @@ pub mod module {
...
+				)
+				.ok_or(ArithmeticError::Overflow)?;
+
+			liquid_to_mint = liquid_to_mint
+				.checked_sub(T::MaxRewardPerEra::get().mul(liquid_to_mint))
+				.expect("Arithmatic should not panic");
```
shaunxw, line 185: Same here.

modules/homa-lite/src/tests.rs
```diff
@@ -42,74 +42,82 @@ fn mock_initialize_token_works() {
...
-			HomaLite::request_mint(Origin::signed(ROOT), amount, 0),
+			HomaLite::mint(Origin::signed(ROOT), amount, 0),
 			orml_tokens::Error::<Runtime>::BalanceTooLow
 		);
 
-		assert_ok!(HomaLite::request_mint(Origin::signed(ALICE), amount, 0));
-		assert_eq!(PendingAmount::<Runtime>::get(&current_batch, &ALICE), amount);
+		// Since the exchange rate is not set, use the default 1:10 ratio
+		// liquid = (amount - MintFee) * 10 * (1 - MaxRewardPerEra)
+		let mut liquid = Permill::from_percent(90).mul((amount - MintFee::get()) * 10);
```
shaunxw, line 61: Unit tests should use numbers to verify the result. Here it's just repeating the same formula.

shaunxw (commented):
Unit tests for `mint` could be improved to verify that:
- The new arithmetic is able to cancel the era rewards with instant LKSM minting.
- After minting LKSM, the exchange ratio maintains within a reasonable range.
- After total staking currency amount updated, the exchange ratio change will make users get staking rewards.

xlc (approved):

modules/homa-lite/src/lib.rs
```diff
@@ -176,13 +176,13 @@ pub mod module {
 				.checked_mul_int(
 					amount
 						.checked_sub(T::MintFee::get())
-						.expect("Arithmatic should not panic"),
+						.expect("Mint amount is ensured to be greater than T::MintFee; qed"),
 				)
 				.ok_or(ArithmeticError::Overflow)?;
 
 			liquid_to_mint = liquid_to_mint
 				.checked_sub(T::MaxRewardPerEra::get().mul(liquid_to_mint))
```
shaunxw, line 184: Not introduced in this PR but would be good to change it to `checked_mul`

shaunxw (approved):

modules/homa-lite/src/lib.rs
syan095: I think checked_mul_int is correct here - it takes in an u128 and produces an u128. CheckedMul gives FixedU128 and I had to convert it back to u128.

shaunxw: Yes, should use `checked_mul_int`.

syan095: Sorry my mistake - i was looking at wrong line of code. For T::MaxRewardPerEra::get().mul(liquid_to_mint), checked_mul is correct. Mainly because checked_mul_int is implemented only for FixedU128, and not for U128.

syan095: Update: Refer to https://github.com/AcalaNetwork/Acala/issues/1321

PR #1311, opened 2021-08-09
check required collateral ratio when withdraw collateral or issue more debit

wangjj9219:
If a CDP is safe but below the required collateral ratio, user wants to deposit collateral or repay debt but the collateral ratio will be still below the required collateral ratio,  the operation will fail. That's not reasonable, should only check the required collateral ratio if the collateral ratio may increase.


xlc (approved):

wangjj9219:
@StrawberryFlavor need test this change by manual

PR #1312, opened 2021-08-09
payout no longer transfer rewards to owner directly

wangjj9219:


modules/incentives/src/lib.rs
```diff
@@ -435,6 +450,49 @@ impl<T: Config> Pallet<T> {
 	pub fn account_id() -> T::AccountId {
 		T::PalletId::get().into_account()
 	}
+
+	#[transactional]
```
shaunxw, line 454:
```suggestion
	#[require_transactional]
```

modules/incentives/src/lib.rs
```diff
@@ -337,7 +352,7 @@ pub mod module {
 		#[transactional]
 		pub fn claim_rewards(origin: OriginFor<T>, pool_id: PoolId<T::RelaychainAccountId>) -> DispatchResult {
 			let who = ensure_signed(origin)?;
-			<orml_rewards::Pallet<T>>::claim_rewards(&who, &pool_id);
+			Self::do_claim_reward(&who, &pool_id)?;
 			Ok(())
```
shaunxw, line 356:
```suggestion
			Self::do_claim_reward(&who, &pool_id)
```

modules/incentives/src/lib.rs
```diff
@@ -337,7 +352,7 @@ pub mod module {
 		#[transactional]
 		pub fn claim_rewards(origin: OriginFor<T>, pool_id: PoolId<T::RelaychainAccountId>) -> DispatchResult {
 			let who = ensure_signed(origin)?;
-			<orml_rewards::Pallet<T>>::claim_rewards(&who, &pool_id);
+			Self::do_claim_reward(&who, &pool_id)?;
```
shaunxw, line 355: I think we don't have to move dispatchable call impl into a new fn prefixed with `do_` anymore. IDEs are able to do syntax analysis with the new pallet attributes macros.

shaunxw (approved):

xlc (approved):

PR #1327, opened 2021-08-12
Homa-lite: use 'Any' network id for sovereign sub-account.

### LINKED ISSUE #1324: Change homa-lite relay chain NetworkId in runtime config to Any
There is no point to set a specific network id in our case. `Any` works both in production and rococo testing env.

shaunxw:
Closes #1324

### DIFF
--- a/runtime/karura/src/lib.rs
+++ b/runtime/karura/src/lib.rs
@@ -1458,7 +1458,7 @@ pub fn create_x2_parachain_multilocation(index: u16) -> MultiLocation {
 	MultiLocation::X2(
 		Junction::Parent,
 		Junction::AccountId32 {
-			network: RelayNetwork::get(),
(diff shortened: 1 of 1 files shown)


syan095:
I think you may need to update the parachain_subaccounts_are_unique test as well

xlc:
Already confirmed this won't break anything?

shaunxw:
> I think you may need to update the parachain_subaccounts_are_unique test as well

This test is only for Mandala runtime and there in no change there.

shaunxw:
> Already confirmed this won't break anything?

Only network id of homa-lite sovereign subaccount location is changed. The relay chain network id for xcm related types and configs is not touched, so there is no breaks.

syan095 (approved):

xlc (approved):

PR #1335, opened 2021-08-16
run integration tests for karura

xlc:


runtime/integration-tests/src/integration_tests.rs
```diff
@@ -472,7 +479,10 @@ fn liquidate_cdp() {
 				AccountId::from(ALICE),
 				10 * dollar(RELAY_CHAIN_CURRENCY),
 				50_000 * dollar(USD_CURRENCY),
+				#[cfg(feature = "with-mandala-runtime")]
 				LiquidationStrategy::Auction,
+				#[cfg(feature = "with-karura-runtime")]
+				LiquidationStrategy::Exchange,
```
xlc, line 485: this make it testing two different thing. better to update whatever makes this different than updating assertions.
@syan095 have identified why this is different.

runtime/integration-tests/src/integration_tests.rs
```diff
@@ -137,24 +137,26 @@ fn run_to_block(n: u32) {
...
-				relay_parent_number: number,
-				relay_parent_storage_root: relay_storage_root,
-				max_pov_size: Default::default(),
-			},
-			relay_chain_state: proof,
-			downward_messages: Default::default(),
-			horizontal_messages: Default::default(),
-		}
-	));
+	{
```
xlc, line 140:
```suggestion
```

syan095: Done

syan095 (approved):

PR #1340, opened 2021-08-17
suport alternative path for some internal swap

### LINKED ISSUE #1303: Support alternative swap path when DEX liquidate unsafe CDPs
now the swap path is dead code: collateral -> stablecoin when DEX liquidate unsafe CDPs,
it's better to support alternative paths, for example:  collateral -> KSM/DOT -> stablecoin ， collateral -> native token -> stablecoin

wangjj9219:
close #1303

### DIFF
--- a/modules/auction-manager/src/lib.rs
+++ b/modules/auction-manager/src/lib.rs
@@ -263,6 +263,12 @@ pub mod module {
 				}
 			}
 		}
+
+		fn integrity_test() {
+			assert!(T::DefaultSwapParitalPathList::get()
+				.iter()
+				.all(|path| path.len() > 0 && path[path.len() - 1] == T::GetStableCurrencyId::get()));
+		}
 	}
 
 	#[pallet::call]
@@ -399,9 +405,9 @@ impl<T: Config> Pallet<T> {
 		}
 
 		// calculate how much collateral to offset target in settle price
-		let stable_currency_id = T::GetStableCurrencyId::get();
-		let settle_price = T::PriceSource::get_relative_price(stable_currency_id, collateral_auction.currency_id)
-			.ok_or(Error::<T>::InvalidFeedPrice)?;
+		let settle_price =
+			T::PriceSource::get_relative_price(T::GetStableCurrencyId::get(), collateral_auction.currency_id)
+				.ok_or(Error::<T>::InvalidFeedPrice)?;
 		let confiscate_collateral_amount = if collateral_auction.always_forward() {
 			collateral_auction.amount
 		} else {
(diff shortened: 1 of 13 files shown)


modules/auction-manager/src/lib.rs
```diff
@@ -565,69 +571,91 @@ impl<T: Config> Pallet<T> {
...
+			let stable_currency_id = T::GetStableCurrencyId::get();
+
+			// iterator default_swap_parital_path_list to try swap until swap succeed.
+			for partial_path in default_swap_parital_path_list {
+				let partial_path_len = partial_path.len();
+
+				// check collateral currency_id and partial_path can form a valid swap path.
+				if partial_path_len > 0
+					&& collateral_auction.currency_id != partial_path[0]
+					&& partial_path[partial_path_len - 1] == stable_currency_id
```
xlc, line 586: do we need this check?
we can have `integrity_test` that asset that all of items in `DefaultSwapParitalPathList` ends with stable currency id

example: 
https://github.com/paritytech/substrate/blob/74101dc21cfffb4c2d014fcc28edc166d5ca1b16/frame/staking/src/lib.rs#L1423

modules/auction-manager/src/lib.rs
```diff
@@ -565,69 +571,91 @@ impl<T: Config> Pallet<T> {
...
+				{
+					let mut swap_path = vec![collateral_auction.currency_id];
+					swap_path.extend(partial_path);
+
+					// DEX must take a higher bid.
+					if bid_price
+						< T::DEX::get_swap_target_amount(&swap_path, collateral_auction.amount).unwrap_or_default()
```
xlc, line 593: we need to do some benchmarking but I am not sure if it is faster to `get_swap_target_amount` and then `swap_exact_collateral_to_stable` or just `swap_exact_collateral_to_stable` in a `transactional` and revert if price too low

this can happen later

modules/auction-manager/src/lib.rs
```diff
@@ -565,69 +571,91 @@ impl<T: Config> Pallet<T> {
...
+							&swap_path,
+							true,
+						) {
+							// swap successfully, will not deal.
+							should_deal = false;
+
+							// refund stable currency to the last bidder, it shouldn't fail and affect the
+							// process. but even it failed, just the winner did not get the bid price. it
+							// can be fixed by treasury council.
+							if let Some(bidder) = maybe_bidder.clone() {
```
xlc, line 609: Looks like `maybe_bidder` is not used anymore so clone is needed? Otherwise can also use `.as_ref`

xlc (approved):

PR #1341, opened 2021-08-17
homa-lite: update cross-chain transfer

shaunxw:


modules/homa-lite/src/lib.rs
```diff
@@ -193,17 +193,14 @@ pub mod module {
...
 				amount,
 				T::SovereignSubAccountLocation::get(),
 				Self::xcm_dest_weight(),
-			)?;
-			ensure!(
-				matches!(xcm_result, Outcome::Complete(_)),
-				Error::<T>::XcmTransferFailed
-			);
+			)
+			.map_err(|_| Error::<T>::XcmTransferFailed)?;
```
xlc, line 203: do we need to map the error (and drop original error), or we can throw the original error?

shaunxw: Right, we can throw the original error now, as `transfer` now returns a dispatch error. Will update.

syan095 (approved):

xlc (approved):

PR #1347, opened 2021-08-19
remove native token support of EVM

### LINKED ISSUE #1342: Remove native token support
Transfer with value should fail. Query balance always response 0.

wangjj9219:
close #1342


zjb0807:
What are the reasons for restricting native tokens?

xlc:
ACA decimal is 12, The native token in EVM is 18. We either need to have some logic to handle the decimal or just disable this feature and require people use ERC20 to send ACA.

modules/evm/src/runner/handler.rs
```diff
@@ -179,16 +179,11 @@ impl<'vicinity, 'config, 'meter, T: Config> Handler<'vicinity, 'config, 'meter,
...
-			&source,
-			&target,
-			transfer.value.saturated_into::<u128>().unique_saturated_into(),
-			ExistenceRequirement::AllowDeath,
-		)
-		.map_err(|_| ExitError::OutOfGas)
+		if transfer.value.is_zero() {
+			Ok(())
+		} else {
+			Err(ExitError::Other("invalid action".into()))
```
xlc, line 185: Maybe should return `OutOfFund` instead. Basically no one has any EVM native token (which is not ACA)

xlc (changes requested):
https://github.com/AcalaNetwork/Acala/blob/958f6148038070415931089fdac389538024ee65/modules/evm/src/runner/mod.rs#L160-L168

This also needs update

modules/evm/src/runner/handler.rs
zjb0807: > https://github.com/AcalaNetwork/Acala/blob/958f6148038070415931089fdac389538024ee65/modules/evm/src/runner/mod.rs#L160-L168
> 
> This also needs update

This will not support `payable`. Is this the desired result？

xlc: Yes. People should just use ERC20. Later we may want to convert all our build-in tokens to ERC677 to support similar semantic of payable.

zjb0807:
1. We can’t disable the feature `Gwei`,`Ether`..., because this is done during solidity compile time.
https://github.com/ethereum/solidity/blob/dea1b9ec7995ff7f16cdee7925a5858c5ce33982/libsolidity/ast/Types.cpp#L886-L913
And there are many js libraries that support the conversion of these units.
2. If disable native transfer, the `payable` related features will be unavailable, which will have a great impact on the contract. It also affects the schedule's pre-compiled contract, and it will not be able to reserve fee.
https://github.com/AcalaNetwork/Acala/blob/71f070aa042b40cb31df22fc6e43ca905ac03ccb/runtime/common/src/precompile/schedule_call.rs#L176-L182
3. I think it is enough to prompt the contract user with the currency decimals, and it does not support `Gwei`,`Ether` key words.

xlc:
I think it is ok to break all `payable` contracts. They should all support ERC20 anyway. For scheduler, it should be possible to ERC20 transfer for fee and then schedule the call.

zjb0807:
use `#[cfg(not(feature = "with-ethereum-compatibility"))]` to handle it.

xlc:
I think we will drop ethereum compatibility mode at some stage if it become too much burden to maintain it

xlc:
let's keep native token support to provider better compatibility

PR #1349, opened 2021-08-22
Enable evm test for mandala and karura

zjb0807:
Enable evm test for mandala and karura.


modules/evm/src/runner/handler.rs
```diff
@@ -638,7 +666,12 @@ impl<'vicinity, 'config, 'meter, T: Config> HandlerT for Handler<'vicinity, 'con
 							// try_or_rollback!(self.storage_meter.record_cost(0));
 							TransactionOutcome::Commit(Capture::Exit((s.into(), out)))
 						}
-						Err(e) => TransactionOutcome::Rollback(Capture::Exit((e.into(), Vec::new()))),
+						Err(e) => {
+							let mut w = sp_std::Writer::default();
+							let _ = core::write!(&mut w, "{:?}", e);
+							substate.err_msg = Some(w.into_inner());
+							TransactionOutcome::Rollback(Capture::Exit((e.into(), Vec::new())))
```
xlc, line 673: can we just put the error message here?
```suggestion
							TransactionOutcome::Rollback(Capture::Exit((e.into(), w.into_inner())))
```

modules/evm/src/runner/handler.rs
```diff
@@ -170,7 +178,27 @@ impl<'vicinity, 'config, 'meter, T: Config> Handler<'vicinity, 'config, 'meter,
...
+					output.extend_from_slice(&message_len.to_vec());
+					output.extend_from_slice(&err_msg);
+
+					log::debug!(
+						target: "evm",
+						"err_msg: {:?}, output: {:?}",
+						err_msg,
+						output,
+					);
+					(e.into(), output.to_vec())
```
xlc, line 197: this ignores `runtime.machine().return_value()`, we need to make sure it is safe to do so. e.g. only override it if it is empty.

modules/evm/src/runner/handler.rs
```diff
@@ -49,6 +49,8 @@ pub struct Handler<'vicinity, 'config, 'meter, T: Config> {
 	pub gasometer: Gasometer<'config>,
 	pub storage_meter: StorageMeter<'meter>,
 	pub is_static: bool,
+	/// precompile return error message.
+	pub err_msg: Option<Vec<u8>>,
```
xlc, line 53: It needs to be cleared at some stage to ensure we don't accidentally return error message from previous error

modules/evm/src/runner/handler.rs
zjb0807: https://github.com/rust-blockchain/evm/blob/22fe78a32af8a790663ac7cd437da89d1cb54577/runtime/src/eval/system.rs#L404

The code of precompile is empty, so the `out_len = 0`. The error message can't return from precompile.
```
let target_len = min(out_len, U256::from(runtime.return_data_buffer.len()));
```
```
ExitReason::Revert(_) => {
	push_u256!(runtime, U256::zero());

	let _ = runtime.machine.memory_mut().copy_large(
		out_offset,
		U256::zero(),
		target_len,
		&runtime.return_data_buffer[..],
	);

	Control::Continue
}
```

runtime/karura/src/lib.rs
```diff
@@ -749,6 +749,9 @@ parameter_type_with_key! {
 				// use the ED of currency_id_0 as the ED of lp token.
 				if currency_id_0 == GetNativeCurrencyId::get() {
 					NativeTokenExistentialDeposit::get()
+				} else if let CurrencyId::Erc20(_) = currency_id_0 {
+					// LP token with erc20
+					1
```
zjb0807, line 754: Maybe `Zero::zero()` is better.

xlc: we may want to allow governance to set ed for lp token with erc20 involved.
@wangjj9219 what do you think?

xlc (approved):

runtime/karura/src/lib.rs
wangjj9219: > we may want to allow governance to set ed for lp token with erc20 involved.
> @wangjj9219 what do you think?

agree, these assets will be added frequently.

PR #1363, opened 2021-08-25
impl fungibles traits for currencies

wangjj9219:


syan095:
Work will be continued from a more updated branch here:
https://github.com/AcalaNetwork/Acala/pull/2030

PR #1376, opened 2021-08-30
Added new Redeem extrinsics

syan095:
Modified existing Mint extrinsics to match with Redeem requests first, and only do XCM transfer if no redeem requests can be matched.

### DIFF
--- a/modules/homa-lite/src/lib.rs
+++ b/modules/homa-lite/src/lib.rs
@@ -27,10 +27,14 @@ pub mod weights;
 use frame_support::{pallet_prelude::*, transactional};
 use frame_system::{ensure_signed, pallet_prelude::*};
 use module_support::{ExchangeRate, Ratio};
-use orml_traits::{MultiCurrency, XcmTransfer};
-use primitives::{Balance, CurrencyId};
-use sp_runtime::{traits::Zero, ArithmeticError, FixedPointNumber, Permill};
-use sp_std::{ops::Mul, prelude::*};
+use orml_traits::{BalanceStatus, MultiCurrency, MultiReservableCurrency, XcmTransfer};
+use primitives::{Balance, CurrencyId, ReserveIdentifier};
+use sp_runtime::{
+	offchain::storage_lock::BlockNumberProvider,
+	traits::{Saturating, Zero},
+	ArithmeticError, FixedPointNumber, Permill,
+};
+use sp_std::{cmp::min, ops::Mul, prelude::*};
 use xcm::opaque::v0::MultiLocation;
 
 pub use module::*;
@@ -40,6 +44,8 @@ pub use weights::WeightInfo;
 pub mod module {
 	use super::*;
 
+	pub type RelaychainBlockNumberOf<T> = <<T as Config>::RelaychainBlockNumber as BlockNumberProvider>::BlockNumber;
+
 	#[pallet::config]
 	pub trait Config: frame_system::Config {
 		type Event: From<Event<Self>> + IsType<<Self as frame_system::Config>::Event>;
@@ -48,7 +54,8 @@ pub mod module {
 		type WeightInfo: WeightInfo;
 
 		/// Multi-currency support for asset management
-		type Currency: MultiCurrency<Self::AccountId, CurrencyId = CurrencyId, Balance = Balance>;
+		type Currency: MultiCurrency<Self::AccountId, CurrencyId = CurrencyId, Balance = Balance>
+			+ MultiReservableCurrency<Self::AccountId, CurrencyId = CurrencyId, Balance = Balance>;
 
 		/// The Currency ID for the Staking asset
 		#[pallet::constant]
@@ -83,6 +90,20 @@ pub mod module {
 		/// The fixed cost of transaction fee for XCM transfers.
 		#[pallet::constant]
 		type MintFee: Get<Balance>;
+
+		/// The fixed cost of withdrawing Staking currency via redeem.
+		#[pallet::constant]
+		type BaseWithdrawFee: Get<Permill>;
+
+		/// TODO: ???
+		type Reserveldentifier: Get<ReserveIdentifier>;
+
+		/// Block number provider for the relaychain.
+		type RelaychainBlockNumber: BlockNumberProvider<BlockNumber = Self::BlockNumber>;
+
+		/// The account ID to redeem from on the relaychain.
+		#[pallet::constant]
+		type ParachainAccount: Get<Self::AccountId>;
 	}
 
 	#[pallet::error]
@@ -93,13 +114,19 @@ pub mod module {
 		MintAmountBelowMinimumThreshold,
 		/// The amount of Staking currency used has exceeded the cap allowed.
 		ExceededStakingCurrencyMintCap,
+		/// There isn't enough reserved currencies to cancel the redeem request.
+		InsufficientReservedBalances,
+		/// Amount redeemed is above total amount staked.
+		InsufficientTotalStakingCurrency,
+		/// The user does not have enough liquid balance to redeem.
+		InsufficientLiquidBalance,
 	}
 
 	#[pallet::event]
 	#[pallet::generate_deposit(pub(crate) fn deposit_event)]
 	#[pallet::metadata(T::AccountId = "AccountId")]
 	pub enum Event<T: Config> {
-		/// The user has requested some Staking currency to be used to mint Liquid Currency.
+		/// The user has Staked some currencies to mint Liquid Currency.
 		/// \[user, amount_staked, amount_minted\]
 		Minted(T::AccountId, Balance, Balance),
 
@@ -112,6 +139,18 @@ pub mod module {
 
 		/// A new weight for XCM transfers has been set.\[new_weight\]
 		XcmDestWeightSet(Weight),
+
+		/// The redeem request has been cancelled, and funds un-reserved.
+		/// \[who, liquid_amount_unreserved\]
+		RedeemRequestCancelled(T::AccountId, Balance),
+
+		/// A new Redeem request has been registered.
+		/// \[who, liquid_amount\]
+		RedeemRequested(T::AccountId, Balance),
+
+		/// The user has redeemed some Liquid currency back to Staking currency.
+		/// \[user, staking_amount_redeemed, liquid_amount_deducted\]
+		Redeemed(T::AccountId, Balance, Balance),
 	}
 
 	/// The total amount of the staking currency on the relaychain.
@@ -133,80 +172,61 @@ pub mod module {
 	#[pallet::getter(fn xcm_dest_weight)]
 	pub type XcmDestWeight<T: Config> = StorageValue<_, Weight, ValueQuery>;
 
+	/// Requests to redeem staked currencies.
+	/// RedeemRequests: Map: AccountId => (liquid_amount: Balance, addtional_fee: Permill)
+	#[pallet::storage]
+	#[pallet::getter(fn redeem_requests)]
+	pub type RedeemRequests<T: Config> = StorageMap<_, Twox64Concat, T::AccountId, (Balance, Permill), OptionQuery>;
+
+	/// The amount of staking currency that is available to be redeemed.
+	/// AvailableStakingBalance: value: Balance
+	#[pallet::storage]
+	#[pallet::getter(fn available_staking_balance)]
+	pub type AvailableStakingBalance<T: Config> = StorageValue<_, Balance, ValueQuery>;
+
+	/// Funds that will be unbounded in the future
+	/// ScheduledUnbound:
+	#[pallet::storage]
+	#[pallet::getter(fn scheduled_unbound)]
+	pub type ScheduledUnbound<T: Config> = StorageValue<_, Vec<(Balance, RelaychainBlockNumberOf<T>)>, ValueQuery>;
+
 	#[pallet::pallet]
 	pub struct Pallet<T>(_);
 
 	#[pallet::call]
 	impl<T: Config> Pallet<T> {
 		/// Mint some Liquid currency, by locking up the given amount of Staking currency.
+		/// Will try to match Redeem Requests if available. Remaining amount is minted via XCM.
+		///
 		/// The exchange rate is calculated using the ratio of the total amount of the staking and
-		/// liquid currency. A portion is reducted (defined as T::MaxRewardPerEra) to make up for
-		/// the fact that staking is only effective from the next era on (on the relaychain).
+		/// liquid currency.
+		///
+		/// If any amount is minted through XCM, a portion of that amount (T::MintFee and
+		/// T::MaxRewardPerEra) is reducted as fee.
 		///
 		/// Parameters:
 		/// - `amount`: The amount of Staking currency to be exchanged.
 		#[pallet::weight(< T as Config >::WeightInfo::mint())]
 		#[transactional]
 		pub fn mint(origin: OriginFor<T>, amount: Balance) -> DispatchResult {
-			let who = ensure_signed(origin)?;
-			// Ensure the amount is above the minimum, after the MintFee is deducted.
-			ensure!(
-				amount > T::MinimumMintThreshold::get().saturating_add(T::MintFee::get()),
-				Error::<T>::MintAmountBelowMinimumThreshold
-			);
+			let minter = ensure_signed(origin)?;
 
-			// Ensure the total amount staked doesn't exceed the cap.
-			let new_total_staked = Self::total_staking_currency()
-				.checked_add(amount)
-				.ok_or(ArithmeticError::Overflow)?;
-			ensure!(
-				new_total_staked <= Self::staking_currency_mint_cap(),
-				Error::<T>::ExceededStakingCurrencyMintCap
-			);
+			Self::do_mint_with_requests(&minter, amount, vec![])
+		}
 
-			let staking_currency = T::StakingCurrencyId::get();
+		/// Mint some Liquid currency, by locking up the given amount of Staking currency.
+		/// This is similar with the mint() extrinsic, except that the given Redeem Requests are
+		/// matched with priority.
+		///
+		/// Parameters:
+		/// - `amount`: The amount of Staking currency to be exchanged.
+		/// - `requests`: The redeem requests that are prioritized to match.
+		#[pallet::weight(< T as Config >::WeightInfo::mint_for_requests())]
+		#[transactional]
+		pub fn mint_for_requests(origin: OriginFor<T>, amount: Balance, requests: Vec<T::AccountId>) -> DispatchResult {
+			let minter = ensure_signed(origin)?;
 
-			// ensure the user has enough funds on their account.
-			T::Currency::ensure_can_withdraw(staking_currency, &who, amount)?;
-
-			// Calculate how much Liquid currency is to be minted.
-			// Gets the current exchange rate
-			let staking_total = Self::total_staking_currency();
-			let liquid_total = T::Currency::total_issuance(T::LiquidCurrencyId::get());
-			let exchange_rate =
-				Ratio::checked_from_rational(liquid_total, staking_total).unwrap_or_else(T::DefaultExchangeRate::get);
-
-			// liquid_to_mint = ( (staked_amount - MintFee) * liquid_total / staked_total ) * (1 -
-			// MaxRewardPerEra)
-			let mut liquid_to_mint = exchange_rate
-				.checked_mul_int(
-					amount
-						.checked_sub(T::MintFee::get())
-						.expect("Mint amount is ensured to be greater than T::MintFee; qed"),
-				)
-				.ok_or(ArithmeticError::Overflow)?;
-
-			liquid_to_mint = liquid_to_mint
-				.checked_sub(T::MaxRewardPerEra::get().mul(liquid_to_mint))
-				.expect("Max rewards cannot be above 100%; qed");
-
-			// All checks pass. Proceed with Xcm transfer.
-			T::XcmTransfer::transfer(
-				who.clone(),
-				staking_currency,
-				amount,
-				T::SovereignSubAccountLocation::get(),
-				Self::xcm_dest_weight(),
-			)?;
-
-			// Mint the liquid currency into the user's account.
-			T::Currency::deposit(T::LiquidCurrencyId::get(), &who, liquid_to_mint)?;
-
-			TotalStakingCurrency::<T>::put(new_total_staked);
-
-			Self::deposit_event(Event::<T>::Minted(who, amount, liquid_to_mint));
-
-			Ok(())
+			Self::do_mint_with_requests(&minter, amount, requests)
 		}
 
 		/// Sets the total amount of the Staking currency that are currently on the relaychain.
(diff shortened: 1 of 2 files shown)


modules/homa-lite/src/lib.rs
```diff
@@ -133,80 +172,61 @@ pub mod module {
 	#[pallet::getter(fn xcm_dest_weight)]
 	pub type XcmDestWeight<T: Config> = StorageValue<_, Weight, ValueQuery>;
 
+	/// Requests to redeem staked currencies.
+	/// RedeemRequests: Map: AccountId => (liquid_amount: Balance, addtional_fee: Permill)
```
xlc, line 176:
```suggestion
	/// RedeemRequests: Map: AccountId => Option<(liquid_amount: Balance, addtional_fee: Permill)>
```

modules/homa-lite/src/lib.rs
```diff
@@ -83,6 +90,20 @@ pub mod module {
 		/// The fixed cost of transaction fee for XCM transfers.
 		#[pallet::constant]
 		type MintFee: Get<Balance>;
+
+		/// The fixed cost of withdrawing Staking currency via redeem.
+		#[pallet::constant]
+		type BaseWithdrawFee: Get<Permill>;
+
+		/// TODO: ???
+		type Reserveldentifier: Get<ReserveIdentifier>;
```
xlc, line 99: This is only needed for named reserves, which is not yet currently supported for MultiCurrency

modules/homa-lite/src/lib.rs
```diff
@@ -256,5 +276,304 @@ pub mod module {
...
+		#[transactional]
+		pub fn request_redeem(origin: OriginFor<T>, liquid_amount: Balance, additional_fee: Permill) -> DispatchResult {
+			let who = ensure_signed(origin)?;
+
+			if liquid_amount.is_zero() {
+				// If the maount is zero, cancel previuos redeem request.
+				if let Some((_, _)) = Self::redeem_requests(&who) {
+					// Unreserve the liquid fee and remove the redeem request.
+					ensure!(
+						T::Currency::unreserve(T::LiquidCurrencyId::get(), &who, liquid_amount) == liquid_amount,
```
xlc, line 295:
```suggestion
						T::Currency::unreserve(T::LiquidCurrencyId::get(), &who, liquid_amount).is_zero(),
```
the return value is how much failed to unreserve

modules/homa-lite/src/lib.rs
```diff
@@ -256,5 +276,304 @@ pub mod module {
...
+					RedeemRequests::<T>::remove(&who);
+
+					Self::deposit_event(Event::<T>::RedeemRequestCancelled(who, liquid_amount));
+				}
+			} else {
+				// Put in an redeem request, or add additional amounts to be redeemed
+				let (redeemed_amount, _) = Self::redeem_requests(&who).unwrap_or((0, Permill::zero()));
+
+				ensure!(
+					liquid_amount < Self::convert_staking_to_liquid(Self::total_staking_currency())?,
```
xlc, line 307: why do we need this check? the reserve will just fail if value is too high

modules/homa-lite/src/lib.rs
```diff
@@ -256,5 +276,304 @@ pub mod module {
...
+						T::Currency::unreserve(T::LiquidCurrencyId::get(), &who, liquid_amount) == liquid_amount,
+						Error::<T>::InsufficientReservedBalances
+					);
+					RedeemRequests::<T>::remove(&who);
+
+					Self::deposit_event(Event::<T>::RedeemRequestCancelled(who, liquid_amount));
+				}
+			} else {
+				// Put in an redeem request, or add additional amounts to be redeemed
+				let (redeemed_amount, _) = Self::redeem_requests(&who).unwrap_or((0, Permill::zero()));
```
xlc, line 304: this way two transaction is required to reduce redeem amount.
should just replace the current request

modules/homa-lite/src/lib.rs
```diff
@@ -256,5 +276,304 @@ pub mod module {
...
+
+				ensure!(
+					liquid_amount < Self::convert_staking_to_liquid(Self::total_staking_currency())?,
+					Error::<T>::InsufficientTotalStakingCurrency
+				);
+
+				let actual_redeem_amount = min(
```
xlc, line 311: no need this check, should just fail when reserve

modules/homa-lite/src/lib.rs
```diff
@@ -256,5 +276,304 @@ pub mod module {
...
+					Self::convert_staking_to_liquid(Self::available_staking_balance())?,
+				);
+				if !actual_redeem_amount.is_zero() {
+					let new_redeemed_amount = redeemed_amount
+						.checked_add(actual_redeem_amount)
+						.ok_or(ArithmeticError::Overflow)?;
+					let new_fee = additional_fee;
+
+					// Reserve the liquid currencies to be redeem.
+					ensure!(
```
xlc, line 322: no need to have this check, reserve will just fail

xlc (changes requested):

modules/homa-lite/src/lib.rs
```diff
@@ -256,5 +276,304 @@ pub mod module {
...
+					RedeemRequests::<T>::remove(&redeemer);
+				} else {
+					RedeemRequests::<T>::insert(&redeemer, (new_balance, extra_fee));
+				}
+			}
+			// Redeem request storage has now been updated.
+			new_balances.clear();
+
+			// Iterate all remaining redeem requests now.
+			for (redeemer, (request_amount, extra_fee)) in RedeemRequests::<T>::iter() {
```
xlc, line 503: add a const to limit maximum iterations to 20, this also helps to provide an upper bound for benchmarking

modules/homa-lite/src/tests.rs
```diff
@@ -126,9 +126,13 @@ fn repeated_mints_have_similar_exchange_rate() {
 
 		// Second exchange
 		// liquid = (1000 - 0.01) * 1004949.9505 / 201000 * 0.99
-		let liquid_2 = 4_949_703_990_002_437;
+		let liquid_2 = 4_949_703_990_002_433; // Actual amount is lower due to rounding loss
 		assert_ok!(HomaLite::mint(Origin::signed(BOB), amount));
-		assert_eq!(Currencies::free_balance(LKSM, &BOB), 9_899_654_490_002_437);
+		assert_eq!(
```
xlc, line 131: use `System::assert_last_event`

modules/homa-lite/src/lib.rs
```diff
@@ -256,5 +281,331 @@ pub mod module {
...
+		pub fn schedule_unbound(
+			origin: OriginFor<T>,
+			staking_amount: Balance,
+			unbound_block: RelaychainBlockNumberOf<T>,
+		) -> DispatchResult {
+			T::GovernanceOrigin::ensure_origin(origin)?;
+
+			let mut current_scheduled_unbound = Self::scheduled_unbound();
+			current_scheduled_unbound.push((staking_amount, unbound_block));
+			ScheduledUnbound::<T>::put(current_scheduled_unbound);
```
xlc, line 341: use `append`

modules/kusama-call-builder/src/lib.rs
```diff
@@ -0,0 +1,95 @@
...
+use sp_runtime::MultiAddress;
+
+// mod mock;
+// mod tests;
+
+pub use module::*;
+use module_support::CallBuilder;
+use primitives::{AccountIndex, Balance};
+
+#[frame_support::pallet]
```
xlc, line 37: this doesn't need to be a pallet

modules/kusama-call-builder/Cargo.toml
```diff
@@ -0,0 +1,30 @@
+[package]
+name = "module-kusama-call-builder"
```
xlc, line 2: I don't want to have a module-polkadot-call-builder with basically the same code in the future.
Make this `module-relaychain` and all the relaychain related code can be put here.
Add new feature `kusama` so we can only include Kusama related code for Karura.

modules/support/src/lib.rs
```diff
@@ -562,3 +562,10 @@ impl CurrencyIdMapping for () {
 pub trait CompoundCashTrait<Balance, Moment> {
 	fn set_future_yield(next_cash_yield: Balance, yield_index: u128, timestamp_effective: Moment) -> DispatchResult;
 }
+
+pub trait CallBuilder<AccountId, Balance> {
+	type RelaychainCall: Encode + Decode;
```
xlc, line 567:
```suggestion
	type RelaychainCall: Codec;
```

modules/support/src/lib.rs
```diff
@@ -562,3 +562,10 @@ impl CurrencyIdMapping for () {
 pub trait CompoundCashTrait<Balance, Moment> {
 	fn set_future_yield(next_cash_yield: Balance, yield_index: u128, timestamp_effective: Moment) -> DispatchResult;
 }
+
+pub trait CallBuilder<AccountId, Balance> {
```
xlc, line 566:
```suggestion
pub trait CallBuilder {
    type AccountId: Codec;
    type Balance: Codec;
```

modules/homa-lite/src/lib.rs
```diff
@@ -283,36 +373,505 @@ pub mod module {
...
+
+		/// Put in an request to redeem Staking currencies used to mint Liquid currency.
+		/// The redemption will happen after the currencies are unbonded on the relaychain.
+		///
+		/// Parameters:
+		/// - `liquid_amount`: The amount of liquid currency to be redeemed into Staking currency.
+		/// - `additional_fee`: Percentage of the fee to be awarded to the minter.
+		#[pallet::weight(< T as Config >::WeightInfo::request_redeem())]
+		#[transactional]
+		pub fn request_redeem(origin: OriginFor<T>, liquid_amount: Balance, additional_fee: Permill) -> DispatchResult {
```
xlc, line 414: put this before `mint_for_requests` so later we may want to remove/update set_xxx methods and won't have effect on user facing calls.

modules/homa-lite/src/lib.rs
```diff
@@ -283,36 +373,505 @@ pub mod module {
...
+			let who = ensure_signed(origin)?;
+
+			if liquid_amount.is_zero() {
+				// If the maount is zero, cancel previuos redeem request.
+				if let Some((request_amount, _)) = Self::redeem_requests(&who) {
```
xlc, line 419:
```suggestion
				if let Some((request_amount, _)) = RedeemRequests::<T>::take(&who) {
```

modules/homa-lite/src/lib.rs
```diff
@@ -283,36 +373,505 @@ pub mod module {
...
+					// Immediately redeem from the available_staking_balances
+					let actual_staking_amount = Self::convert_liquid_to_staking(actual_liquid_amount)?;
+
+					// Redeem from the available_staking_balances costs no extra fee.
+					T::Currency::deposit(T::StakingCurrencyId::get(), &who, actual_staking_amount)?;
+					let slash_amount = T::Currency::slash(T::LiquidCurrencyId::get(), &who, actual_liquid_amount);
+					ensure!(slash_amount.is_zero(), Error::<T>::InsufficientLiquidBalance);
+
+					// Update the available_staking_balance
+					available_staking_balance = available_staking_balance
```
xlc, line 445:
```suggestion
					let available_staking_balance = available_staking_balance
```
and no need to make it `let mut` on line 429

modules/homa-lite/src/lib.rs
```diff
@@ -283,36 +373,505 @@ pub mod module {
...
+						actual_staking_amount,
+						actual_liquid_amount,
+					));
+				}
+
+				// Unredeemed requests are added to a queue.
+				let liquid_remaining = liquid_amount
+					.checked_sub(actual_liquid_amount)
+					.expect("min() ensures actual amount cannot be more than the original; qed");
+				if !liquid_remaining.is_zero() {
```
xlc, line 461: I think we need a MinRedeemAmount and skip making redeem request if remaining is less than that.
Should be an error if initial requested amount is less than `MinRedeemAmount` and greater than `available_staking_balance`

modules/homa-lite/src/lib.rs
```diff
@@ -283,36 +373,505 @@ pub mod module {
...
+					// Lock the liquid currency.
+					T::Currency::reserve(T::LiquidCurrencyId::get(), &who, liquid_remaining)?;
```
xlc, line 463: this should only reserve the diff between the new request and old request
and should unreserve if old request is more than new request

modules/relaychain/src/lib.rs
```diff
@@ -0,0 +1,156 @@
...
+	pub dest: <T::Lookup as StaticLookup>::Source,
+	/// transfer amount
+	#[codec(compact)]
+	pub value: Balance,
+}
+
+#[derive(Encode, Decode, RuntimeDebug)]
+pub enum UtilityCall<RelaychainCall> {
+	#[codec(index = 1)]
+	AsDerivative(UtilityAsDerivativeCall<RelaychainCall>),
```
xlc, line 57: Another way is doing like this to avoid having a struct for each parameter. But the current way is also fine by me.

```suggestion
	AsDerivative { index: u16, call: RelaychainCall },
```

xlc (commented):
Better to put XCM related code into another pallet. But it can happen in a later PR.

modules/relaychain/src/lib.rs
```diff
@@ -0,0 +1,190 @@
...
+	}
+
+	fn balances_transfer_keep_alive(to: Self::AccountId, amount: Self::Balance) -> Self::RelaychainCall {
+		RelaychainCall::Balances(BalancesCall::TransferKeepAlive(BalancesTransferKeepAliveCall {
+			dest: T::Lookup::unlookup(to),
+			value: amount,
+		}))
+	}
+
+	fn finalize_call(call: Self::RelaychainCall, extra_fee: Self::Balance, weight: Weight, debt: Weight) -> Xcm<()> {
```
xlc, line 165: the name is a bit confusing. maybe `to_xcm_transact`?

modules/support/src/lib.rs
```diff
@@ -562,3 +564,39 @@ impl CurrencyIdMapping for () {
 pub trait CompoundCashTrait<Balance, Moment> {
 	fn set_future_yield(next_cash_yield: Balance, yield_index: u128, timestamp_effective: Moment) -> DispatchResult;
 }
+
+pub trait CallBuilder {
```
xlc, line 568: better to break this into multiple smaller trait. this can happen in later PR

something like

```rust
pub trait CallBuilder {
 	type AccountId: FullCodec;
 	type Balance: FullCodec;
 	type RelaychainCall: FullCodec;
 	fn finalize_call(call: Self::RelaychainCall, extra_fee: Self::Balance, weight: Weight, debt: Weight) -> Xcm<()>;
}

trait UtilityCallBuilder: CallBuilder {
	fn utility_batch_call(calls: Vec<Self::RelaychainCall>) -> Self::RelaychainCall;
 	fn utility_as_derivative_call(call: Self::RelaychainCall, index: u16) -> Self::RelaychainCall;
}

trait StakingCallBuilder: CallBuilder {
 	fn staking_withdraw_unbonded(num_slashing_spans: u32) -> Self::RelaychainCall;
}

```

syan095: Ran into issues on the compatibility of `RelaychainCall` between the different traits.
will will not be included in this PR

modules/homa-lite/src/lib.rs
```diff
@@ -134,82 +204,93 @@ pub mod module {
...
+	/// AvailableStakingBalance: value: Balance
+	#[pallet::storage]
+	#[pallet::getter(fn available_staking_balance)]
+	pub type AvailableStakingBalance<T: Config> = StorageValue<_, Balance, ValueQuery>;
+
+	/// Funds that will be unbonded in the future
+	/// ScheduledUnbond:
 	#[pallet::storage]
-	#[pallet::getter(fn xcm_dest_weight)]
-	pub type XcmDestWeight<T: Config> = StorageValue<_, Weight, ValueQuery>;
```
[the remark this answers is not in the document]
syan095: reverted storage rename so migration is not needed

modules/homa-lite/src/lib.rs
```diff
@@ -139,77 +213,73 @@ pub mod module {
...
+						let res = Self::process_scheduled_unbond(staking_amount);
+						if res.is_ok() {
+							current_weight = required_weight;
+
+							scheduled_unbond.remove(0);
+							ScheduledUnbond::<T>::put(scheduled_unbond);
+						} else {
+							log::debug!("{:?}", res);
+							debug_assert!(res.is_ok());
+						}
```
zjb0807, line 256:
```suggestion
						log::debug!("{:?}", res);
						debug_assert!(res.is_ok());

						scheduled_unbond.remove(0);
						ScheduledUnbond::<T>::put(scheduled_unbond);
						current_weight = required_weight;
```

modules/homa-lite/src/lib.rs
```diff
@@ -220,10 +290,7 @@ pub mod module {
 		///   conversion rate.
 		#[pallet::weight(< T as Config >::WeightInfo::set_total_staking_currency())]
 		#[transactional]
-		pub fn set_total_staking_currency(
-			origin: OriginFor<T>,
-			#[pallet::compact] staking_total: Balance,
```
zjb0807, line 225: Why remove `#[pallet::compact]`?

modules/homa-lite/src/lib.rs
```diff
@@ -293,29 +360,504 @@ pub mod module {
...
+					// Immediately redeem from the available_staking_balances
+					let actual_staking_amount = Self::convert_liquid_to_staking(actual_liquid_amount)?;
+
+					// Redeem from the available_staking_balances costs no extra fee.
+					T::Currency::deposit(
+						T::StakingCurrencyId::get(),
+						&who,
+						actual_staking_amount
+							.checked_sub(T::XcmUnbondFee::get())
+							.expect("Ensured amount is larger than fee; qed"),
```
zjb0807, line 430: I think to use `saturating_sub` instead of `checked_sub().expect` is better.

modules/homa-lite/src/lib.rs
```diff
@@ -139,77 +213,73 @@ pub mod module {
...
+	/// AvailableStakingBalance: value: Balance
+	#[pallet::storage]
+	#[pallet::getter(fn available_staking_balance)]
+	pub type AvailableStakingBalance<T: Config> = StorageValue<_, Balance, ValueQuery>;
+
+	/// Funds that will be unbonded in the future
+	/// ScheduledUnbond:
+	#[pallet::storage]
+	#[pallet::getter(fn scheduled_unbond)]
+	pub type ScheduledUnbond<T: Config> = StorageValue<_, Vec<(Balance, RelaychainBlockNumberOf<T>)>, ValueQuery>;
```
zjb0807, line 232: use `BoundedVec` instead of `Vec`

modules/homa-lite/src/lib.rs
```diff
@@ -293,29 +360,504 @@ pub mod module {
...
+			Ok(())
+		}
+
+		#[transactional]
+		fn process_scheduled_unbond(staking_amount: Balance) -> DispatchResult {
+			let msg = Self::construct_xcm_unreserve_message(T::ParachainAccount::get(), staking_amount);
+
+			let res = pallet_xcm::Pallet::<T>::send_xcm(Here, Parent.into(), msg);
+			//(origin, MultiLocation::X1(Junction::Parent.into()), msg.into());
+			if res.is_ok() {
```
zjb0807, line 762: missed `else`.  Should check the error here, then the `if` block can be removed.

modules/homa-lite/src/lib.rs
```diff
@@ -293,29 +360,504 @@ pub mod module {
...
-			.reciprocal()
-			.unwrap_or_default()
+	impl<T: Config> Pallet<T> {
+		/// Calculate the exchange rate between the Staking and Liquid currency.
+		/// returns Ratio(liquid : staking) = liquid_total_issuance / total_staking_amount
+		/// If the exchange rate cannot be calculated, T::DefaultExchangeRate is used
+		fn get_exchange_rate() -> Ratio {
+			let staking_total = Self::total_staking_currency();
+			let liquid_total = T::Currency::total_issuance(T::LiquidCurrencyId::get());
+			Ratio::checked_from_rational(liquid_total, staking_total).unwrap_or_else(T::DefaultExchangeRate::get)
```
xlc, line 536: `.unwrap_or_else(T::DefaultExchangeRate::get)` isn't right. it needs to be `1 / DefaultExchangeRate`

should have more tests to make sure `get_exchange_rate` / `convert_liquid_to_staking`  / `convert_staking_to_liquid` give correct number on default case

modules/homa-lite/src/lib.rs
```diff
@@ -293,29 +360,504 @@ pub mod module {
...
+			request_extra_fee: Permill,
+			liquid_amount_remaining: &mut Balance,
+			new_balances: &mut Vec<(T::AccountId, Balance, Permill)>,
+		) -> DispatchResult {
+			let liquid_amount_can_be_redeemed = min(request_amount, *liquid_amount_remaining);
+
+			let new_amount = request_amount
+				.checked_sub(liquid_amount_can_be_redeemed)
+				.expect("min() guarantees that the amount deducted is less than the current balance; qed");
+			*liquid_amount_remaining = liquid_amount_remaining
```
xlc, line 591: we have a lot of code with this pattern. maybe have a helper method to make it

```
let (
  liquid_amount_can_be_redeemed,
  new_amount,
  liquid_amount_remaining
) = min_diff(request_amount, liquid_amount_remaining);
```

modules/homa-lite/src/lib.rs
```diff
@@ -293,29 +360,504 @@ pub mod module {
...
+		///   Only redeem up to this amount.
+		/// - `new_balances`: Stores the new `RedeemRequest` balances. This should be iterated after
+		///   to update the actual storage in bulk. Actual `RedeemRequest` storage is NOT modified
+		///   here.
+		fn match_mint_with_redeem_request(
+			minter: &T::AccountId,
+			redeemer: &T::AccountId,
+			request_amount: Balance,
+			request_extra_fee: Permill,
+			liquid_amount_remaining: &mut Balance,
```
xlc, line 583: why not return the value instead of passing mutable reference?

modules/homa-lite/src/lib.rs
```diff
@@ -293,29 +360,504 @@ pub mod module {
...
+						actual_liquid_amount,
+					));
+				}
+
+				// Update storage to the new balances. Remove Redeem requests that have been filled.
+				Self::update_redeem_requests(&new_balances);
+
+				AvailableStakingBalance::<T>::put(available_staking_balance);
+				Self::deposit_event(Event::<T>::ScheduledUnbondWithdrew(staking_amount));
+			}
```
xlc, line 807: `send_xcm` should never fail right? maybe a d warning log if it failed.

modules/relaychain/src/lib.rs
```diff
@@ -0,0 +1,159 @@
...
+pub enum UtilityCall<RelaychainCall> {
+	#[codec(index = 1)]
+	AsDerivative(u16, RelaychainCall),
+	#[codec(index = 2)]
+	BatchAll(Vec<RelaychainCall>),
+}
+
+#[derive(Encode, Decode, RuntimeDebug)]
+pub enum StakingCall {
+	#[codec(index = 3)]
```
xlc, line 55: is it possible to write some tests to check the index with the Call form Kusama runtime? this can happen in later PR

runtime/karura/src/lib.rs
```diff
@@ -1532,10 +1533,21 @@ parameter_types! {
...
 	pub const LKSMCurrencyId: CurrencyId = CurrencyId::Token(TokenSymbol::LKSM);
 	pub MinimumMintThreshold: Balance = 10 * cent(KSM);
+	pub MinimumRedeemThreshold: Balance = 100 * cent(LKSM);
 	pub RelaychainSovereignSubAccount: MultiLocation = create_x2_parachain_multilocation(RelaychainSubAccountId::HomaLite as u16);
-	pub MaxRewardPerEra: Permill = Permill::from_rational(500u32, 1_000_000u32); // 1.2 ^ (1/365) = 1.0004996359
+	pub RelaychainSovereignSubAccountId: AccountId = Utility::derivative_account_id(
+		ParachainInfo::get().into_account(),
+		RelaychainSubAccountId::HomaLite as u16
+	);
+	pub MaxRewardPerEra: Permill = Permill::from_rational(261u32, 1_000_000u32); // 1.1^(1/365) = 1.00026115788
```
xlc, line 1542: why change this?

runtime/karura/src/lib.rs
```diff
@@ -1532,10 +1533,21 @@ parameter_types! {
...
 	pub MintFee: Balance = 20 * millicent(KSM); // 2x XCM fee on Kusama
-	pub DefaultExchangeRate: ExchangeRate = ExchangeRate::saturating_from_rational(1, 10);
+	pub DefaultExchangeRate: ExchangeRate = ExchangeRate::saturating_from_rational(10, 1);
```
xlc, line 1544: better to not change this as it is used by some js code

runtime/karura/src/lib.rs
```diff
@@ -1532,10 +1533,21 @@ parameter_types! {
...
+	pub BaseWithdrawFee: Permill = Permill::from_rational(7338u32, 1_000_000); // 10% yield per year, unbounding period = 28 days. 1.1^(28/365) = 1.0073382
```
xlc, line 1545: 20% per year, unbonding period is 7 days fro Kusama

runtime/karura/src/lib.rs
```diff
@@ -1532,10 +1533,21 @@ parameter_types! {
...
+	pub MaximumRedeemRequestMatchesForMint: u32 = 20;
+	pub RelaychainUnbondingSlashingSpans: u32 = 5;
+	pub ParachainAccount: AccountId = ParachainInfo::get().into_account();
+	pub SubAccountIndex: u16 = RelaychainSubAccountId::HomaLite as u16;
+	pub const XcmUnbondFee: Balance = 600_000_000; // From homa-lite integration test.
```
xlc, line 1550: where is this number coming from?

PR #1378, opened 2021-08-30
do not use default when real origin not found

### LINKED ISSUE #1372: Should not use default() value.
https://github.com/AcalaNetwork/Acala/blob/9b672bd58252054fafe2b4e3f4c411c1a591852c/modules/currencies/src/lib.rs#L277

wangjj9219:
close #1372


zjb0807 (approved):
LGTM

xlc (approved):

PR #1385, opened 2021-09-01
provide price for LKSM

wangjj9219:


xlc (changes requested):
Better to have HomaLite implements this and `type LiquidStakingExchangeRateProvider = HomaLite;`

modules/homa-lite/src/lib.rs
```diff
@@ -258,3 +258,13 @@ pub mod module {
...
 	}
 }
+
+impl<T: Config> ExchangeRateProvider for Pallet<T> {
+	fn get_exchange_rate() -> ExchangeRate {
+		ExchangeRate::checked_from_rational(
+			T::Currency::total_issuance(T::LiquidCurrencyId::get()),
+			Pallet::<T>::total_staking_currency(),
+		)
+		.unwrap_or_default()
```
xlc, line 268: better to match https://github.com/AcalaNetwork/Acala/blob/8e22258292a962f8188b7bc4419d07146814b8f9/modules/homa-lite/src/lib.rs#L176-L177

and have it use this helper method

xlc (approved):

PR #1390, opened 2021-09-02
Use frame-support library instead of orml-utilities for Iterators

ferrell-code:
companion to https://github.com/open-web3-stack/open-runtime-module-library/pull/600

### DIFF
--- a/modules/auction-manager/src/lib.rs
+++ b/modules/auction-manager/src/lib.rs
@@ -35,7 +35,7 @@ use frame_system::{
 	pallet_prelude::*,
 };
 use orml_traits::{Auction, AuctionHandler, Change, MultiCurrency, OnNewBidResult};
-use orml_utilities::{IterableStorageMapExtended, OffchainErr};
+use orml_utilities::OffchainErr;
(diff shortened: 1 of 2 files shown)


modules/auction-manager/src/lib.rs
```diff
@@ -359,35 +359,34 @@ impl<T: Config> Pallet<T> {
...
 			}
-			Self::submit_cancel_auction_tx(collateral_auction_id);
-			guard.extend_lock().map_err(|_| OffchainErr::OffchainLock)?;
-		}
 
-		if iterator.finished {
 			to_be_continue.clear();
-		} else {
-			to_be_continue.set(&iterator.storage_map_iterator.previous_key);
-		}
```
ferrell-code, line 385: I believe that the substrate iterator next() method will only return `None` in the case of the iterator finishing, hence the for loop instead of while let Some. 

I assume this was put like this for some reason but I can't tell why a for loop wasn't used in the first place

xlc (approved):

wangjj9219 (approved):

PR #1396, opened 2021-09-04
multi currency incentive rewards

ermalkaleci:
Stash


modules/incentives/src/lib.rs
```diff
@@ -222,7 +237,7 @@ pub mod module {
 		PoolId<T::RelaychainAccountId>,
 		Twox64Concat,
 		T::AccountId,
-		Balance,
+		BTreeMap<CurrencyId, Balance>,
```
xlc, line 240: need migration

xlc (commented):
Also need to add the migration for orml-rewards to karura runtime

xlc (commented):
Need migration for orml-rewards

wangjj9219 (commented):
plan to use `HomaValidatorAllowance` to implement the multi-currency reward? But reward accumulation for `HomaValidatorAllowance` is not periodic.

The definition for `PoolId` is no longer appropriate. `PooId` implies the reward currency id(native and stable or other allowance) and accumulation method （periodly by fixed amount ，or the multiple of  the base amount and fixed rate, or transfer allowance non-periodicly)

Now, `orml_rewards` supports to record multi-currency rewards amount with single `PoolId`, 
And the pool share of `DexIncentives` is equal to `DexSaving` for the same trading pair. 
it's better to refactor `PoolId` to `DEX`, `Loans` and others, and need new storage 
`map Vec<(reward_currency_id, reward_amount_per_period)>` 
to record periodic fixed amount reward, and define new enum to imply base amount source, add another storage 
`map PoolId => Vec<(reward_currency_id, fixed_rate, BaseAmountSourceEnum)>` to record the fixed rate and the base amount source. Aggregate these configurations when accumulating periodic multi rewards.

@xlc What do you think? 
If need to do, more migrations are required, and I am afraid that testing and front-end modification cannot be completed before this Thursday’s runtime upgrade.



 with PoolId to support native incentives and other currency reward.

modules/incentives/src/lib.rs
```diff
@@ -328,10 +295,39 @@ pub mod module {
 				0
 			}
 		}
+
+		fn on_runtime_upgrade() -> Weight {
```
xlc, line 299: we don't want this anymore right

wangjj9219: But other storages still need to be migrated on `on_runtime_upgrade`.
Do you mean to add calls to trigger manual migration for all storage like `migrate_share_and_withdrawn_reward` and `migrate_pending_rewards`?

xlc: right. that's fine then.

xlc (approved):

PR #1409, opened 2021-09-11
Fix offchain workers and add tests

### LINKED ISSUE #1405: Unit tests for offchain workers
Unsafe CDPs should be liquidated by cdp-engine offchain worker

ferrell-code:
Closes #1405 

Fixes auction-manager and cdp-engine offchain workers and adds tests

### DIFF
--- a/modules/auction-manager/src/lib.rs
+++ b/modules/auction-manager/src/lib.rs
@@ -371,10 +371,6 @@ impl<T: Config> Pallet<T> {
 
 		#[allow(clippy::while_let_on_iterator)]
 		while let Some((collateral_auction_id, _)) = iterator.next() {
-			if iteration_count >= max_iterations {
-				finished = false;
-				break;
-			}
 			iteration_count += 1;
 
 			if let (Some(collateral_auction), Some((_, last_bid_price))) = (
@@ -384,17 +380,26 @@ impl<T: Config> Pallet<T> {
 				// if collateral auction has already been in reverse stage,
 				// should skip it.
 				if collateral_auction.in_reverse_stage(last_bid_price) {
+					if iteration_count == max_iterations {
+						finished = false;
+						break;
+					}
 					continue;
 				}
 			}
 			Self::submit_cancel_auction_tx(collateral_auction_id);
+
+			if iteration_count == max_iterations {
+				finished = false;
+				break;
+			}
 			guard.extend_lock().map_err(|_| OffchainErr::OffchainLock)?;
 		}
 
 		if finished {
 			to_be_continue.clear();
 		} else {
-			to_be_continue.set(&iterator.prefix());
+			to_be_continue.set(&iterator.last_raw_key());
 		}
 
 		// Consume the guard but **do not** unlock the underlying lock.
--- a/modules/auction-manager/src/lib.rs
+++ b/modules/auction-manager/src/lib.rs
@@ -351,8 +351,7 @@ impl<T: Config> Pallet<T> {
 		// get the max iterationns config
 		let max_iterations = StorageValueRef::persistent(OFFCHAIN_WORKER_MAX_ITERATIONS)
 			.get::<u32>()
-			.unwrap_or(Some(DEFAULT_MAX_ITERATIONS))
-			.ok_or(OffchainErr::OffchainStore)?;
+			.unwrap_or(Some(DEFAULT_MAX_ITERATIONS));
 
 		log::debug!(
 			target: "auction-manager",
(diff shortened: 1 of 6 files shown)


modules/cdp-engine/src/lib.rs
```diff
@@ -648,23 +648,29 @@ impl<T: Config> Pallet<T> {
...
 		let is_shutdown = T::EmergencyShutdown::is_shutdown();
-		let mut map_iterator =
-			<loans::Positions<T>>::iter_prefix_from(currency_id, start_key.clone().ok_or(OffchainErr::OffchainStore)?);
+
+		// If start key is Some(value) continue iterating from that point in storage otherwise start
+		// iterating from the beginning of <loans::Positons<T>>
+		let mut map_iterator = match start_key.clone() {
+			Some(key) => <loans::Positions<T>>::iter_prefix_from(currency_id, key),
+			None => <loans::Positions<T>>::iter_prefix(currency_id),
+		};
```
ferrell-code, line 661: Mistake 1: when start key is none that is not an error it just means start iterating from the beginning

modules/cdp-engine/src/lib.rs
```diff
@@ -648,23 +648,29 @@ impl<T: Config> Pallet<T> {
 		// get the max iterationns config
 		let max_iterations = StorageValueRef::persistent(OFFCHAIN_WORKER_MAX_ITERATIONS)
 			.get::<u32>()
-			.unwrap_or(Some(DEFAULT_MAX_ITERATIONS))
-			.ok_or(OffchainErr::OffchainStore)?;
+			.unwrap_or(Some(DEFAULT_MAX_ITERATIONS));
```
ferrell-code, line 651: Mistake 2: When max iterations is stored as None this is not an error, just means iterator will run to completion no matter the size of the iterator. This is probably not what collators would want to set but is still a valid state

modules/cdp-engine/src/lib.rs
```diff
@@ -708,7 +714,7 @@ impl<T: Config> Pallet<T> {
 				};
 			to_be_continue.set(&(next_collateral_position, Option::<Vec<u8>>::None));
 		} else {
-			to_be_continue.set(&(collateral_position, Some(map_iterator.last_raw_key())));
+			to_be_continue.set(&(collateral_position, Some(map_iterator.prefix())));
```
ferrell-code, line 717: Mistake 3: We should store the current raw key, Not the previous one

ferrell-code:
Now there are unit tests so no one else can break it :)

modules/cdp-engine/src/lib.rs
xlc: I think we want to make None to be `DEFAULT_MAX_ITERATIONS`

ferrell-code: Yeah that makes sense, It is kinda a foot-gun unnecessarily for node operators

modules/auction-manager/src/lib.rs
```diff
@@ -351,7 +351,8 @@ impl<T: Config> Pallet<T> {
 		// get the max iterationns config
 		let max_iterations = StorageValueRef::persistent(OFFCHAIN_WORKER_MAX_ITERATIONS)
 			.get::<u32>()
-			.unwrap_or(Some(DEFAULT_MAX_ITERATIONS));
+			.unwrap_or(Some(DEFAULT_MAX_ITERATIONS))
+			.unwrap_or(DEFAULT_MAX_ITERATIONS);
```
xlc, line 355: should have some tests for this change

xlc (dismissed):

ferrell-code:
Don't merge this yet, key still doesn't work to continue iterator from where it left off, I will make add to test so it fails in that case, when I figure out how to fix it

modules/cdp-engine/src/tests.rs
```diff
@@ -922,3 +941,192 @@ fn close_cdp_has_debit_by_swap_on_alternative_path() {
...
+			Change::NoChange,
+			Change::NoChange,
+			Change::NoChange,
+		));
+		run_to_block_offchain(2);
+		// should only run 1000 iterations stopping due to DEFAULT_MAX_ITERATIONS
+		assert_eq!(pool_state.write().transactions.len(), 1000);
+		// should only now run 1 iteration to finish off where it ended last block
+		run_to_block_offchain(3);
+		assert_eq!(pool_state.write().transactions.len(), 1001);
```
ferrell-code, line 1130: this tests that iterator correctly takes off where it was stopped, the only way for it to get 1001 tx is if the iterator is set to the 1000 on the second offchain worker

xlc (approved):

PR #1413, opened 2021-09-15
Purge-chain to work for --dev

### LINKED ISSUE #1373: purge-chain for dev is broken
```
target/debug/acala purge-chain --dev -y
Error: Input("Relay chain argument error: Invalid input: `rococo-local` only supported with `rococo-native` feature enabled.")
```

ferrell-code:
Closes #1373

Problem originates upstream in cumulus, this should be a useful work around


xlc (approved):
Should we have an issue for Cumulus?

ferrell-code:
Probably, perhaps we are supposed to use `sc_cli` and not use `cumulus_client_cli` at all, but its probably good for us to get clarification either way. I can fix it there if they want to support --dev

PR #1073, opened 2021-06-08
Integrate chainlink feed

wangjj9219:
close #1056

PR #1415, opened 2021-09-15
integrate PINT pallets to mandela

mattsse:
Hey @xlc this is an initial proposal to integrate some of [PINT's](https://github.com/ChainSafe/PINT) pallets into the mandela runtime.

## Changes

* add the [PINT repo](https://github.com/ChainSafe/PINT) as submodule
* implement some pallets for mandela
    * `pint_local_treasury`
    * `pint_committee`
    * `pint_asset_index`
    * `pint_saft_registry`
* No chainlink pallet, instead implemented `pint::PriceFeed` for `AggregatedDataProvider`.
* added `TokenSymbol::PINT` (we selected ID 169 randomly, perhaps an ID from the ecosystem range would be more suitable?)
* Set Alice as PINT council member in the `PintCommitteeConfig` of the mandela chain spec
* Reused the weights generated from the PINT runtime (wonder if we should copy them to the mandala weights folder?)
* patched the orml repository url against the local submodule
* The `ExistentialDeposits` for `TokenSymbol::PINT` was set to `MAX` for karura, but we weren't sure for mandala.

Also no XCM support at this point.

Integrating runtime-benchmarks needed some adapters to mitigate the missing chainlink pallet and support for acala's `CurrencyId` type (`TokenSymbol`), because PINT relies on them for runtime benchmarks.
Setting up chainlink feeds during runtime is a bit cumbersome if you have any feature suggestions I'm happy to get this included in the chainlink pallet directly.

If you have any questions, concerns, please reach out.
(CC @BenWhiteJam)

Also, FYI, there is a bug in prost that affects substrate's sc-consensus on polkadot-v0.9.9 (or rather libp2p 0.37.1) so nothing that depends on polkadot-v0.9.9 compiles currently with the latest rust nightly https://github.com/tokio-rs/prost/issues/526


xlc:
It is ok to use weights from PINT runtime for now. Before mainnet deployment we can do a re benchmarking.
The ED value could be something ~$0.01. If PITN starts with $1 evaluation, it can be 0.01 token.

We have a WIP PR with chainlink integration https://github.com/AcalaNetwork/Acala/pull/1073. It is a bit outdated and we will resume the work next week.

PR #1419, opened 2021-09-17
proxy integration test

### LINKED ISSUE #1298: integration tests for ProxyTypes
Followup of https://github.com/AcalaNetwork/Acala/pull/1288
After #899

ferrell-code:
Closes #1298

### DIFF
--- a/runtime/integration-tests/src/integration_tests.rs
+++ b/runtime/integration-tests/src/integration_tests.rs
@@ -2065,7 +2065,7 @@ fn proxy_permissions_correct() {
 			assert_ok!(Proxy::proxy(
 				Origin::signed(AccountId::from(BOB)),
 				AccountId::from(ALICE),
-				None,
+				Some(ProxyType::Governance),
 				gov_call.clone()
 			));
 			let hash = BlakeTwo256::hash_of(&(BlakeTwo256::hash(b"bob is awesome"), AccountId::from(BOB)));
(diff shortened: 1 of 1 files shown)


ferrell-code:
I find it a little weird that the proxy call returns a `Ok(())` when the inner call fails, I would think that should be an error. But that is a substrate thing, it works as expected

xlc (commented):
can we have more both positive and negative test for other proxy types?

runtime/integration-tests/src/integration_tests.rs
```diff
@@ -1882,3 +1882,276 @@ fn cdp_treasury_handles_honzon_surplus_correctly() {
...
+			// the transfer call fails as Bob only had governence permission for alice
+			assert!(Currencies::free_balance(NATIVE_CURRENCY, &AccountId::from(BOB)) < 100 * dollar(NATIVE_CURRENCY));
+
+			// Alice gives loan permissions to Bob
+			assert_ok!(Proxy::remove_proxy(
+				Origin::signed(AccountId::from(ALICE)),
+				AccountId::from(BOB),
+				ProxyType::Governance,
+				0
+			));
```
ferrell-code, line 2099: I'm a bit confused why removing the previous proxy is needed. The proxy calls in my tests aren't working when there are multiple ProxyType for an account. commenting out this line shouldn't effect the Loan proxy but it does. I feel like I am missing something really dumb

xlc: If there is multiple proxy, you need to specify the proxy type instead of passing None, which it will stop on the first one when finding proxy, which may not be the one you want.

https://github.com/paritytech/substrate/blob/master/frame/proxy/src/lib.rs#L772

ferrell-code: Ahhhhh, that was making me feel crazy thanks

runtime/integration-tests/src/integration_tests.rs
```diff
@@ -1882,3 +1882,276 @@ fn cdp_treasury_handles_honzon_surplus_correctly() {
...
+				None,
+				gov_call.clone()
+			));
+			let hash = BlakeTwo256::hash_of(&(BlakeTwo256::hash(b"bob is awesome"), AccountId::from(BOB)));
+			// last event was sucessful tip call
+			assert_eq!(
+				System::events()
+					.into_iter()
+					.map(|r| r.event)
+					.filter_map(|e| if let Event::Tips(inner) = e { Some(inner) } else { None })
```
xlc, line 2077: why not using `System::assert_last_event`?

ferrell-code: It doesn't really matter, but that is more explicit. The result of `System::assert_last_event` is just `pallet_proxy::Event::ProxyExecuted(Ok())` which is less specific, but I don't feel strongly either way

ferrell-code: So that verifies that the event from the tip pallet did indeed return what was expected

xlc (approved):

PR #1426, opened 2021-09-22
Allow atomic conversion of Staking Currency CDPs to Liquid Currency CDPs 

### LINKED ISSUE #1173: Ability to swap staking asset to liquid asset in cdp loan directly
This will allow people to swap KSM to LKSM directly without repaying KSM

ferrell-code:
Fixes #1173, 

Needs a lot more tests

Also this current PR makes the swap all or nothing perhaps we should implement ability to choose how much collateral/debit to move over to staking

Edit: This also currently doesn't follow Verify First, Write Last but my understanding is the `#[transactional]` attribute makes it acceptable

### DIFF
--- a/modules/cdp-engine/src/lib.rs
+++ b/modules/cdp-engine/src/lib.rs
@@ -224,6 +224,8 @@ pub mod module {
 		MustAfterShutdown,
 		/// Failed to swap debit by default path list
 		SwapDebitFailed,
+		/// CDP does not exist
+		NoOpenPosition,
 	}
 
 	#[pallet::event]
@@ -1043,11 +1045,12 @@ fn pick_u32<R: RngCore>(rng: &mut R, max: u32) -> u32 {
 	rng.next_u32() % max
 }
 
+// Allows for homa-lite module to access cdp functionality
 impl<T: Config> UpdateLoan<T::AccountId, Amount> for Pallet<T> {
 	fn get_position(who: &T::AccountId, staking_id: CurrencyId) -> Result<Position, DispatchError> {
 		match loans::Positions::<T>::try_get(staking_id, who) {
 			Ok(val) => Ok(val),
-			Err(_e) => Err(DispatchError::Other("No Loan Present")),
+			Err(_e) => Err(Error::<T>::NoOpenPosition.into()),
 		}
 	}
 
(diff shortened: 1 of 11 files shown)


modules/loans/src/lib.rs
```diff
@@ -268,7 +258,7 @@ impl<T: Config> Pallet<T> {
 	}
 
 	/// mutate records of collaterals and debits
-	fn update_loan(
+	pub fn update_loan(
 		who: &T::AccountId,
 		currency_id: CurrencyId,
 		collateral_adjustment: Amount,
```
ferrell-code, line 264: I don't like that I made this public, because I see why it was private as it is dangerous to use... but it was useful to use in the UpdateLoan trait

modules/cdp-engine/src/lib.rs
```diff
@@ -1042,3 +1044,31 @@ impl<T: Config> RiskManager<T::AccountId, CurrencyId, Balance, Balance> for Pall
...
+	fn swap_position_to_liquid(
+		who: &T::AccountId,
+		staking_id: CurrencyId,
+		liquid_id: CurrencyId,
+		liquid_adjustment: Amount,
+		staking_adjustment: Amount,
+		debit_adjustment: Amount,
+	) -> DispatchResult {
+		Self::adjust_position(who, liquid_id, liquid_adjustment, debit_adjustment)?;
+		loans::Pallet::<T>::update_loan(who, staking_id, -staking_adjustment, -debit_adjustment)
```
wangjj9219, line 1072: The parameter type is `Amount`, and our purpose is to convert the collateral KSM/DOT into LKSM/LDOT. 

Should ensure `liquid_adjustment`, `staking_adjustment` and `debit_adjustment` are positive numbers or zero here, because if there are any negative numbers among these, the operation will be reversed and some risk checks may be bypassed, which will be dangerous. It is recommended to use `Balance` as the parameter type although the type conversion will be troublesome

ferrell-code: Yeah I agree it makes a lot more sense to pass `Balance` in the arguments and do the type conversion there instead of the `checked_into` currently done in the `mint_from_cdp_loan`.

In general I don't love how both `transfer_collateral_from_loan` and `swap_position_to_liquid` are unsafe individually and only work together but I don't see a better way

wangjj9219 (dismissed):

ferrell-code:
Closing because Stale, still nice to have in future

PR #1437, opened 2021-09-28
ts tests

zjb0807:


runtime/karura/src/lib.rs
```diff
@@ -1299,6 +1303,20 @@ pub type ScheduleCallPrecompile = runtime_common::ScheduleCallPrecompile<
...
 
+pub struct FindAuthorTruncated<F>(PhantomData<F>);
+impl<F: FindAuthor<u32>> FindAuthor<H160> for FindAuthorTruncated<F> {
+	fn find_author<'a, I>(digests: I) -> Option<H160>
+	where
+		I: 'a + IntoIterator<Item = (ConsensusEngineId, &'a [u8])>,
+	{
+		if let Some(author_index) = F::find_author(digests) {
+			let authority_id = Aura::authorities()[author_index as usize].clone();
+			return Some(H160::from_slice(&authority_id.to_raw_vec()[4..24]));
```
xlc, line 1314: should just use our EVM account mapping to convert Substrate address to EVM address

xlc: we can use the standard `FindAuthor<AccoutnId>` and do the conversion in module_evm

ts-tests/build/IERC20Metadata.json
```diff
@@ -0,0 +1,244 @@
+{
```
xlc, line 1: do we want to include the build folder?

zjb0807: Prepare to delete the `modules/evm-bridge/src/erc20_demo_contract` file and reference it from the build directory, but there is a problem

xlc (approved):

Issue #1421, opened 2021-09-18
Make evm call compact


xlc:
For storage limit, gas limit, value, they can all be`#[pallet::compact]`

PR #1448, opened 2021-10-01
Add compact

zjb0807:
Closes: #1421

### DIFF
--- a/modules/cdp-engine/src/lib.rs
+++ b/modules/cdp-engine/src/lib.rs
@@ -81,6 +81,7 @@ pub struct RiskManagementParams {
 	/// Maximum total debit value generated from it, when reach the hard
 	/// cap, CDP's owner cannot issue more stablecoin under the collateral
 	/// type.
+	#[codec(compact)]
 	pub maximum_total_debit_value: Balance,
 
 	/// Extra interest rate per sec, `None` value means not set
--- a/modules/cdp-engine/src/lib.rs
+++ b/modules/cdp-engine/src/lib.rs
@@ -81,7 +81,6 @@ pub struct RiskManagementParams {
 	/// Maximum total debit value generated from it, when reach the hard
 	/// cap, CDP's owner cannot issue more stablecoin under the collateral
 	/// type.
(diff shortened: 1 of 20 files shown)


modules/staking-pool/src/lib.rs
```diff
@@ -80,10 +80,13 @@ impl Default for Phase {
 #[derive(PartialEq, Eq, Clone, Encode, Decode, RuntimeDebug, Default)]
 pub struct Ledger {
 	/// The amount of total bonded.
+	#[codec(compact)]
```
zjb0807, line 83: I'm not sure if it will break the storage

modules/evm/src/lib.rs
```diff
@@ -243,8 +246,10 @@ pub mod module {
 	/// Account definition used for genesis block construction.
 	pub struct GenesisAccount<Balance, Index> {
 		/// Account nonce.
+		#[codec(compact)]
```
xlc, line 249: no need compact for genesis related code

modules/cdp-engine/src/lib.rs
```diff
@@ -81,6 +81,7 @@ pub struct RiskManagementParams {
 	/// Maximum total debit value generated from it, when reach the hard
 	/// cap, CDP's owner cannot issue more stablecoin under the collateral
 	/// type.
+	#[codec(compact)]
```
zjb0807, line 84: I'm not sure if it will break the storage

modules/evm/src/lib.rs
```diff
@@ -234,7 +235,9 @@ pub mod module {
 
 	#[derive(Clone, Copy, Eq, PartialEq, RuntimeDebug, Encode, Decode, MaxEncodedLen)]
 	pub struct CodeInfo {
+		#[codec(compact)]
```
xlc, line 238: no need compact for storage related code, it will require a migration and the storage saving on on chain state isn't worth the additional computation

xlc (changes requested):

xlc (approved):

PR #1463, opened 2021-10-09
Fix eth call

zjb0807:
- set `config.estimate = true` when eth call, avoid calculation of gas and storage.
- Not support eth call without `to` address.


xlc:
How to estimate gas cost of contract deployment?

zjb0807:
> How to estimate gas cost of contract deployment?

by `estimate_resources` RPC.
The PR only modified `call` RPC

Issue #1464, opened 2021-10-09
Fix test fail on master


xlc:
https://github.com/AcalaNetwork/Acala/runs/3845104977?check_suite_focus=true

PR #1465, opened 2021-10-09
Fix ci on master

zjb0807:
Closes: #1464


xlc:
they those tests are slow? waiting for new block?

zjb0807:
Loading airdrop at startup is slow, and there are multiple rpc requests

PR #1477, opened 2021-10-12
add support of PHA transfer on karura network

tolak:


tolak:
Close this PR, would have a new one soon

PR #1514, opened 2021-10-22
Feature/add update available staking

### LINKED ISSUE #1509: Governance method to adjust availableStakingBalance
Useful when people mint KSM, and then someone else make a redeem request before the KSM is bonded on Kusama

### LINKED ISSUE #1545: HomaLite on_idle leave dust available_staking_balance due to rounding error
**Describe the bug**
Due to rounding error in exchange calculation, available_staking_balance remains during on_idle's unbond

**Expected Behavior**
available_staking_balance becomes 0 when it is all spent, and no further redeem requests are affected.

**Current Behavior**
after available_staking_balance is below a certain amount, on_idle iterates all redeem requests and burn a tiny amount of liquid currency, but does not mint Staking currency in exchange.

Bug is present in release-karura-1.5.3

syan095:
Added a new extrinsic to udpate the availalbe_staking_balances
Also automatically match any redeem requests after new Staking balances becomes available

closes #1509
fixes #1545


modules/homa-lite/src/benchmarking.rs
```diff
@@ -73,7 +73,9 @@ benchmarks! {
 
 	set_total_staking_currency {}: _(RawOrigin::Root, 1_000_000_000_000)
 
-	adjust_total_staking_currency {}: _(RawOrigin::Root, AmountOf::<T>::default())
+	adjust_total_staking_currency {}: _(RawOrigin::Root, AmountOf::<T>::max_value())
+
+	adjust_available_staking_balance {}: _(RawOrigin::Root, AmountOf::<T>::max_value())
```
xlc, line 78: the weight depends on number of matched requests, which needs to be reflected on benchmarks

modules/homa-lite/src/lib.rs
```diff
@@ -775,20 +815,35 @@ pub mod module {
...
+		/// This should be called when new available_staking_balance becomes available.
+		#[transactional]
+		fn process_redeem_requests_with_available_staking_balance() -> DispatchResult {
 			let mut new_balances: Vec<(T::AccountId, Balance, Permill)> = vec![];
-			let mut available_staking_balance = Self::available_staking_balance()
-				.checked_add(staking_amount)
-				.ok_or(ArithmeticError::Overflow)?;
+			let mut available_staking_balance = Self::available_staking_balance();
 			for (redeemer, (request_amount, extra_fee)) in RedeemRequests::<T>::iter() {
 				// If all the currencies are minted, return.
```
ferrell-code, line 848: related to Bryan's benchmarking comment, Probably need to bound this to a max number of iterations

modules/homa-lite/src/benchmarking.rs
syan095: Reworked how weights are calculated so `adjust_total_staking_currency` takes into account the number of matches this call "allows"

modules/homa-lite/src/lib.rs
syan095: Updated `adjust_total_staking_currency` to add a limit to the number of requests to match. Also calculate weight from it.

modules/homa-lite/src/lib.rs
```diff
@@ -257,14 +261,31 @@ pub mod module {
...
+			// With remaining weight, calculate max number of redeems that can be matched
+			let num_redeem_matches = remaining_weight
+				.saturating_sub(current_weight)
+				.checked_div(<T as Config>::WeightInfo::redeem_with_available_staking_balance())
+				.unwrap_or_default();
+
+			// Iterate through existing redeem_requests, and try to match them with `available_staking_balance`
+			let res = Self::process_redeem_requests_with_available_staking_balance(num_redeem_matches);
+			debug_assert!(res.is_ok());
+			if res.is_ok() {
```
xlc, line 282:
```suggestion
			if Ok(count) = res) {
```
and replace `res.unwrap_or_default()`

modules/homa-lite/src/lib.rs
```diff
@@ -821,20 +916,26 @@ pub mod module {
 					actual_staking_amount,
 					actual_liquid_amount,
 				));
+				num_matched += 1u64;
+
+				// If all the currencies are minted, return.
+				if available_staking_balance.is_zero() || num_matched >= max_num_matches {
```
xlc, line 922: to better handle rounding error, could stop when available_staking_balance <= minimal_balance

modules/homa-lite/src/lib.rs
```diff
@@ -567,6 +589,53 @@ pub mod module {
...
+		#[pallet::weight(
+			< T as Config >::WeightInfo::adjust_available_staking_balance_with_no_matches().saturating_add(
+			max_num_matches.saturating_mul(< T as Config >::WeightInfo::redeem_with_available_staking_balance())
+			)
+		)]
+		#[transactional]
+		pub fn adjust_available_staking_balance(
+			origin: OriginFor<T>,
+			by_amount: AmountOf<T>,
+			max_num_matches: u64,
```
xlc, line 612: why u64? u32 is already more than enough

ferrell-code:
I don't have anything else to add, looking good, will need a lot of testing :P

modules/homa-lite/src/lib.rs
syan095: This also resolves #1545 
Added test to test the bug is fixed

modules/homa-lite/src/lib.rs
syan095: I originally used u64 to avoid conversion to Weights, but I think you are right - it is more correct to use u32

modules/homa-lite/src/lib.rs
```diff
@@ -775,25 +843,51 @@ pub mod module {
...
+		///
+		/// params:
+		/// 	- `max_num_matches`: Maximum number of redeem requests to be matched.
+		///
+		/// return:
+		/// 	Result<u32, DispatchError>: The number of redeem reqeusts actually matched.
+		#[transactional]
+		pub fn process_redeem_requests_with_available_staking_balance(
+			max_num_matches: u32,
+		) -> Result<u32, sp_runtime::DispatchError> {
```
xlc, line 879: looks like this method doesn't fail? so should just return u32

syan095: There's a few point that can return error. This function is doing the redeem work with available KSMs unbonded.

xlc (approved):

PR #1538, opened 2021-10-27
Feature/homa lite periodic update

### LINKED ISSUE #1497: Add periodic TotalStakingCurrency update
To save governance from daily updates to TotalStakingCurrency, add an periodic update that inflate the amount by a configurable proportion.

This should automatically update the TotalStakingCurrency to keep the Liquid to Staking ratio more accurate.

syan095:
TotalStakingCurrency in the HomaLite module periodically updates to simulate staking rewards from the Relaychain.
Add Configuration  for update frequency and interest rate.

closes #1497


modules/homa-lite/src/lib.rs
```diff
@@ -267,6 +280,21 @@ pub mod module {
 			}
 			current_weight
 		}
+
+		fn on_initialize(n: T::BlockNumber) -> Weight {
+			// Update the total amount of Staking balance by acrueing the interest periodically.
+			if !Self::staking_interest_rate_per_update().is_zero()
```
xlc, line 286: better to cache the read result

runtime/karura/src/lib.rs
```diff
@@ -1556,6 +1556,7 @@ parameter_types! {
 	// Calculated from polkadot/xcm/xcm-builder: fn buy_weight
 	// We must charge higher than what Kusama required (533_333_300, obtained from integration test)
 	pub XcmUnbondFee: Balance = 60 * millicent(KSM);
+	pub const StakingUpdateFrequency: BlockNumber = 14_400; // Update staking total Once per day
```
xlc, line 1559: use DAYS

https://github.com/AcalaNetwork/Acala/blob/851064d1aad6245cac194abf50d66bd6ad9b5646/runtime/karura/src/constants.rs#L32

xlc (changes requested):

runtime/karura/src/lib.rs
syan095: Also updated integration test

modules/homa-lite/src/lib.rs
```diff
@@ -267,6 +281,23 @@ pub mod module {
 			}
 			current_weight
 		}
+
+		fn on_initialize(n: T::BlockNumber) -> Weight {
+			// Update the total amount of Staking balance by acrueing the interest periodically.
+			let interest_rate = Self::staking_interest_rate_per_update();
+			if !interest_rate.is_zero()
+				&& n.checked_rem(&T::StakingUpdateFrequency::get())
+					.unwrap_or_else(One::one)
```
xlc, line 290: this is a bit confusing

could be 

`n.checked_rem(&T::StakingUpdateFrequency::get()).map_or(false, Zero::is_zero)`

modules/homa-lite/src/lib.rs
```diff
@@ -267,6 +281,23 @@ pub mod module {
...
+					.is_zero()
+			{
+				let current = Self::total_staking_currency();
+				let new_total = current.saturating_add(interest_rate.mul(current));
+				let _ = Self::update_total_staking_currency_storage(new_total);
```
xlc, line 295: add comment explain why it is ok to ignore the result
in this case, it can only fail if current is zero, which is ok to fail

modules/homa-lite/src/lib.rs
```diff
@@ -868,6 +912,18 @@ pub mod module {
 				Self::xcm_dest_weight(),
 			)
 		}
+
+		/// Helper function that update the storage with the new
+		fn update_total_staking_currency_storage(new_total: Balance) -> DispatchResult {
```
xlc, line 917: most of the time we are modify the current value, so could make this takes `f: impl FnOnce(Balance) -> Balance` and `*current = f(current)`

modules/homa-lite/src/lib.rs
syan095: Added comment

modules/homa-lite/src/lib.rs
syan095: Updated function and funtion signature

modules/homa-lite/src/lib.rs
```diff
@@ -366,7 +367,7 @@ pub mod module {
 					.ok_or(ArithmeticError::Underflow)?;
 			}
 
-			Self::update_total_staking_currency_storage(current_staking_total)
+			Self::update_total_staking_currency_storage(|_n| current_staking_total)
```
xlc, line 370: the calculation logic should move inside

modules/homa-lite/src/lib.rs
```diff
@@ -914,14 +915,14 @@ pub mod module {
...
 
 		/// Helper function that update the storage with the new
-		fn update_total_staking_currency_storage(new_total: Balance) -> DispatchResult {
-			ensure!(!new_total.is_zero(), Error::<T>::InvalidTotalStakingCurrency);
-
-			// Update storage and emit event.
-			TotalStakingCurrency::<T>::mutate(|current| {
-				*current = new_total;
-				Self::deposit_event(Event::<T>::TotalStakingCurrencySet(*current));
+		fn update_total_staking_currency_storage(f: impl FnOnce(Balance) -> Balance) -> DispatchResult {
```
xlc, line 919:
```suggestion
		fn update_total_staking_currency_storage(f: impl FnOnce(Balance) -> Result<Balance, DispatchError>) -> DispatchResult {
```

and then use try_mutate so the closure can return error

modules/homa-lite/src/lib.rs
```diff
@@ -868,6 +914,19 @@ pub mod module {
 				Self::xcm_dest_weight(),
 			)
 		}
+
+		/// Helper function that update the storage of total_staking_currency and emit event.
+		#[transactional]
```
xlc, line 919:
```suggestion
```

modules/homa-lite/src/lib.rs
```diff
@@ -267,6 +281,24 @@ pub mod module {
...
+					.map_or(false, |n| n.is_zero())
+			{
+				// Inflate the staking total by the interest rate.
+				// This will only fail when current TotalStakingCurrency is 0. In this case it is OK to fail.
+				let _ = Self::update_total_staking_currency_storage(|current| {
+					Ok(current.saturating_add(interest_rate.mul(current)))
+				});
+				<T as Config>::WeightInfo::on_initialize()
+			} else {
+				0
```
xlc, line 299: this should be one storage read

syan095: Added a new benchmark function to make return weight more accurate

modules/homa-lite/src/lib.rs
```diff
@@ -894,7 +939,8 @@ pub mod module {
 				);
 				let actual_staking_amount = Self::convert_liquid_to_staking(actual_liquid_amount)?;
 
-				TotalStakingCurrency::<T>::mutate(|x| *x = x.saturating_sub(actual_staking_amount));
+				total_staking_currency = total_staking_currency.saturating_sub(actual_staking_amount);
+				Self::update_total_staking_currency_storage(|_n| Ok(total_staking_currency))?;
```
xlc, line 943: why not do the original way?
```suggestion
				Self::update_total_staking_currency_storage(|n| Ok(n.saturating_sub(total_staking_currency)))?;
```

syan095: Yeah i think at some point `total_staking_currency` was used somewhere else. But his is redundant now. so updated to do the modification in the closure.

xlc (approved):

Issue #1557, opened 2021-10-29
disable xcm methods


zqhxuyuan:
on Karura side when invoke xcm methods like teleport_assets/reserve_transfer_assets to relaychain, the outcome may failed or not expected(the testcase is in [PR-1556](https://github.com/AcalaNetwork/Acala/pull/1556)). so we need to disable such kind of methods.

current, if disable methods, in the polakdot.js app will get BadOrigin error. But if invoke method by Pallet(ex. in testcase, calling PalletXcm::teleport_assets() or PalletXcm::reserve_transfer_assets()), it still can invoke those methods successfully.

Ps: do we need to check if it effect our own [app](https://apps.karura.network/)?

zqhxuyuan:
as @shaunxw recommend, it's better to use whitelist, not blacklist.

zqhxuyuan:
duplicate to: https://github.com/AcalaNetwork/Acala/issues/1549

PR #1558, opened 2021-10-29
Disable xcm method

zqhxuyuan:
issue: https://github.com/AcalaNetwork/Acala/issues/1557

### DIFF
--- a/runtime/karura/src/lib.rs
+++ b/runtime/karura/src/lib.rs
@@ -206,6 +206,8 @@ impl Contains<Call> for BaseCallFilter {
 				| Call::PolkadotXcm(pallet_xcm::Call::execute { .. })
 				| Call::PolkadotXcm(pallet_xcm::Call::teleport_assets { .. })
 				| Call::PolkadotXcm(pallet_xcm::Call::reserve_transfer_assets { .. })
+				| Call::PolkadotXcm(pallet_xcm::Call::limited_reserve_transfer_assets { .. })
+				| Call::PolkadotXcm(pallet_xcm::Call::limited_teleport_assets { .. })
 		);
 		if is_xcm_call {
 			return false;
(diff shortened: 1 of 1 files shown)


shaunxw (commented):
I would recommend to use allow list on `pallet-xcm`, instead of block list. Polkadot xcm pallet has been adding dispatchable calls in it's new releases, and it's easy to miss new added calls with block list.

xlc:
best to do a exhaustive match listing all the calls so in future if new calls are introduced, it will be a compiling error

zqhxuyuan:
As the Call enum has __Ignore case, so here the default case we return unimplemented!()

runtime/karura/src/lib.rs
```diff
@@ -228,6 +215,28 @@ impl Contains<Call> for BaseCallFilter {
...
+					return false;
+				}
+				pallet_xcm::Call::force_xcm_version { .. }
+				| pallet_xcm::Call::force_default_xcm_version { .. }
+				| pallet_xcm::Call::force_subscribe_version_notify { .. }
+				| pallet_xcm::Call::force_unsubscribe_version_notify { .. } => {
+					return true;
+				}
+				_ => {
+					unimplemented!()
```
xlc, line 235:
```suggestion
					unreachable!()
```

runtime/karura/src/lib.rs
```diff
@@ -215,6 +215,28 @@ impl Contains<Call> for BaseCallFilter {
...
+				| pallet_xcm::Call::limited_teleport_assets { .. } => {
+					return false;
+				}
+				pallet_xcm::Call::force_xcm_version { .. }
+				| pallet_xcm::Call::force_default_xcm_version { .. }
+				| pallet_xcm::Call::force_subscribe_version_notify { .. }
+				| pallet_xcm::Call::force_unsubscribe_version_notify { .. } => {
+					return true;
+				}
+				_ => {
```
xlc, line 234: Can you do this?

```suggestion
				__Ignored => {
```

zqhxuyuan: yeh, I can fix this

xlc (approved):

zqhxuyuan:
@zjb0807 would you look at the failed e2e-tests?

zjb0807:
> @zjb0807 would you look at the failed e2e-tests?

Retry it works.

PR #1578, opened 2021-11-02
query_info and query_fee_details return no fee for unsigned extrinsics.

### LINKED ISSUE #1570: Update transaction payment RPC for unsigned extrinsics
https://github.com/paritytech/substrate/pull/10107

wangjj9219:
close #1570

### DIFF
--- a/modules/transaction-payment/src/lib.rs
+++ b/modules/transaction-payment/src/lib.rs
@@ -408,7 +408,7 @@ where
 	///
 	/// All dispatchables must be annotated with weight and will have some
 	/// fee info. This function always returns.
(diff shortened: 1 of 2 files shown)


xlc (approved):

zjb0807:
https://github.com/AcalaNetwork/Acala/blob/1317df598f50c990cbc15a15d85c6c0103ad67db/modules/evm/rpc/src/lib.rs#L396-L408

The PR break `estimate_resources`

xlc:
Right. This means we cannot estimate weight without signing a transaction. I think we won't do this. If we really want to identify all the weights & fees in a block, it can check for signed/unsigned first.

Issue #612, opened 2020-12-07
EVM: Lazy contract removal


xlc:
We will have similar issue with this
https://github.com/paritytech/substrate/issues/6594

After #1466 is implemented, should just create a remote prefix task when delete contract. Need to make sure we charge enough gas for this.

Issue #1459, opened 2021-10-07
EVM: Avoid reset storage on create contract


ermalkaleci:
We need to think a solution to not reset storage when creating new contract. This is not bounded and the maintainer shouldn't pay for someone's else garbage. Maybe we need to make it impossible to recreate the same address

https://github.com/AcalaNetwork/Acala/blob/4ba09696394648909e3daacdddf027c7dc335a31/modules/evm/src/runner/stack.rs#L583-L584

xlc:
we should just remove all the account storage when contract is removed

zjb0807:
Closed by #1583

PR #1583, opened 2021-11-03
Idle schedule evm task

zjb0807:
Closes: #612 #1459


modules/evm/src/lib.rs
```diff
@@ -890,18 +909,26 @@ impl<T: Config> Pallet<T> {
...
+			)
 		})?;
 
 		// this should happen after `Accounts` is updated because this could trigger another updates on
 		// `Accounts`
 		frame_system::Pallet::<T>::dec_providers(&address_account)?;
 
-		Ok(size)
+		// estimate weight
+		let weight = (ContractStorageSizes::<T>::get(address)
```
xlc, line 927: I don't think it is useful to charge additional weight here. The tx fee will be charged based on max gas and refund unused but it cannot charge more if more weights are returned.

The additional removal work is done on idle block time, so it is ok for them to be free. We may also only refund x% of the deposit.

modules/evm/src/lib.rs
```diff
@@ -1494,3 +1529,91 @@ impl<T: Config + Send + Sync> SignedExtension for SetEvmOrigin<T> {
 		Ok(())
 	}
 }
+
+#[derive(Clone, RuntimeDebug, PartialEq, Encode, Decode, TypeInfo)]
+pub enum EvmTask<T: Config> {
+	// TODO: update
```
xlc, line 1535: write an issue for those TODO

modules/evm/src/lib.rs
```diff
@@ -1494,3 +1529,91 @@ impl<T: Config + Send + Sync> SignedExtension for SetEvmOrigin<T> {
...
+							finished: false,
+						}
+					}
+				}
+			}
+		}
+	}
+}
+
+impl<T: Config> From<EvmTask<T>> for () {
```
xlc, line 1615: this is for test only right? should add cfg flag

modules/support/src/lib.rs
```diff
@@ -600,3 +601,25 @@ pub trait CallBuilder {
...
+pub trait DispatchableTask {
+	fn dispatch(self, weight: Weight) -> TaskResult;
+}
+
+/// Idle scheduler trait
+pub trait IdleScheduler<Task> {
+	fn schedule(task: Task) -> DispatchResult;
+}
+
+impl DispatchableTask for () {
```
xlc, line 615: should use cfg to mark those for test only

primitives/src/task.rs
```diff
@@ -16,18 +16,29 @@
 // You should have received a copy of the GNU General Public License
 // along with this program. If not, see <https://www.gnu.org/licenses/>.
 use frame_support::weights::Weight;
+use sp_runtime::DispatchResult;
+
+#[allow(dead_code)]
```
xlc, line 21: is this still a dead code?

xlc (commented):
now delete contract is not atomic action. is it possible to create a contract with same address of a deleting one? I think that's possible and we need to make sure that action will fail, instead of overriding storages.

ermalkaleci:
we need to make sure it's not possible to deploy at the same address until everything is wiped out

modules/evm/src/lib.rs
zjb0807: #1591

zjb0807:
> we need to make sure it's not possible to deploy at the same address until everything is wiped out

https://github.com/AcalaNetwork/Acala/blob/ba2e0cee5e6d78aa06ac40ef5dfa444f169efd4d/modules/evm/src/runner/state.rs#L580-L583

I checked the `storage_size` and remove it at
https://github.com/AcalaNetwork/Acala/blob/ba2e0cee5e6d78aa06ac40ef5dfa444f169efd4d/modules/evm/src/lib.rs#L1584 

I'll add a test case.

zjb0807:
In order to avoid adding extra `storage_size`, delay deleting account to check the same address deployment problem

modules/evm/src/lib.rs
```diff
@@ -1494,3 +1521,94 @@ impl<T: Config + Send + Sync> SignedExtension for SetEvmOrigin<T> {
...
+							"EvmTask::Remove: [from: {:?}, contract: {:?}, maintainer: {:?}, count: {:?}, result: {:?}]",
+							caller, contract, maintainer, count, res
+						);
+
+						// Remove account after all of the storages are cleared.
+						Accounts::<T>::take(contract);
+
+						TaskResult {
+							result: res,
+							used_weight: <T as frame_system::Config>::DbWeight::get()
```
xlc, line 1582: we should do some real benchmarking here to include the weights of refund and remove accounts

this can happen later

xlc (approved):

PR #1584, opened 2021-11-03
able to use KAR and LKSM for crosschain transfer fee

xlc:


runtime/karura/src/lib.rs
```diff
@@ -1430,6 +1430,21 @@ parameter_types! {
 		// kUSD:KSM = 400:1
 		ksm_per_second() * 400
 	);
+	pub KarPerSecond: (AssetId, u128) = (
+		MultiLocation::new(
+			1,
+			X2(Parachain(u32::from(ParachainInfo::get())), GeneralKey(KAR.encode())),
+		).into(),
+		kar_per_second() * 400
```
shaunxw, line 1438: I think it should be `kar_per_second()`

xlc:
```suggestion
		kar_per_second()
```

shaunxw (approved):

Issue #1506, opened 2021-10-20
Fix & Restore code coverage


xlc:
Code coverage is disabled on #1505 because it always fail. Need to fix & restore it at some stage.

zjb0807:
https://github.com/parallel-finance/parallel/blob/7aa8849629261e8d652fa032db581cc55ae873de/.github/workflows/coverage.yml#L71

PR #1590, opened 2021-11-04
Fix test coverage for acala

syan095:
Closes: #1506

Fixed test coverage script
Excluded modules that are not useful and are causing errors.
Modules excluded are:
/nodes/*
/primitives
/runtime/*


xlc:
can you delete the `coverage.yml.disabled` file

zjb0807 (approved):

Issue #1574, opened 2021-11-02
module-asset-registry


xlc:
Allow to support foreign asset without runtime upgrade

- Types
  - enum CurrencyId
    - Token(TokenSymbol)
    - DexShare(DexShare, DexShare)
      - Token(TokenSymbol)
      - Erc20(EvmAddress)
      - LiquidCroadloan(Lease)
      - ForeignAsset(ForeignAssetId)
    - Erc20(EvmAddress)
    - StableAssetPoolToken(PoolId)
    - LiquidCroadloan(Lease)
    - ForeignAsset(ForeignAssetId)
  - type ForeignAssetId = u16
- module-asset-registry
  - types
    - struct AssetMetadata
      - name: Vec\<u8>
      - symbol: Vec\<u8>
      - decimals: u8
      - minimal_balance: Balance
  - config
    - type RegisterOrigin: EnsureOrigin
      - root or half council
  - storages
    - AssetMetadatas: map MultiLocation => metadata
  - calls
    - fn register_foreign_asset(origin, loc: VersionedMultiLocation, metadata: AssetMetadata)
      - ensure RegisterOrigin
    - fn update_foreign_asset(origin, loc: VersionedMultiLocation, metadata: AssetMetadata)
      - ensure RegisterOrigin

    
Needs to be merged with evm-manager

zqhxuyuan:
does this feature something like https://github.com/open-web3-stack/orml-token-registry/issues/1 ?

xlc:
They are mostly unrelated. Token registry is a way for everyone to register & share token information. This is about a pallet that allow governance to specify token information via proposals.

PR #1598, opened 2021-11-09
Add module asset registry

zjb0807:
Closes: #1574 

TODO:
- [x] benchmarking
- [x] EvmAddress support `DexShare::LiquidCroadloan` and `DexShare::ForeignAsset`
- [x] Intergration tests


modules/asset-registry/src/lib.rs
```diff
@@ -51,17 +64,74 @@ pub mod module {
...
+	#[pallet::getter(fn multi_locations)]
+	pub type MultiLocations<T: Config> = StorageMap<_, Twox64Concat, ForeignAssetId, MultiLocation, OptionQuery>;
+
+	/// The storages for AssetMetadatas.
+	///
+	/// AssetMetadatas: map MultiLocation => Option<AssetMetadata>
+	#[pallet::storage]
+	#[pallet::getter(fn asset_metadatas)]
+	pub type AssetMetadatas<T: Config> =
+		StorageMap<_, Twox64Concat, MultiLocation, AssetMetadata<BalanceOf<T>>, OptionQuery>;
```
xlc, line 133: use `ForeignAssetId` as key can reduce migration required when upgrading MultiLocation in later xcm version

zjb0807:
Try to integrate `statemine` to test cross-chain transfer in another PR

shaunxw (commented):
Note `v1::MultiLocation` is used in storages, we would need to do migration if upgrade to `v2` in the future.

xlc:
We deliberately use v1:MultiLocation. v2 is just reexport of v1.
We want to avoid the VersionedMultiLocation because it will not be unique.

modules/asset-registry/src/lib.rs
```diff
@@ -0,0 +1,534 @@
...
+
+				let mut vec = Vec::new();
+				vec.extend_from_slice(&b"LP "[..]);
+				vec.extend_from_slice(&name_0);
+				vec.extend_from_slice(&b" - ".to_vec());
+				vec.extend_from_slice(&name_1);
+				Some(vec)
+			}
+			CurrencyId::Erc20(address) => Pallet::<T>::get_erc20_mapping(address).map(|v| v.name),
+			CurrencyId::StableAssetPoolToken(_) => None,
```
xlc, line 349: we should have an issue and TODO for this

xlc (approved):

shaunxw (dismissed):

shaunxw (approved):

PR #1608, opened 2021-11-12
Update CdpEngine Benches

ferrell-code:
`liquidate_by_dex` benchmark is updated to correctly use worst case scenario. The `liquidate_by_auction` is also updated for worse case scenario, but there is no way to cheaply  know the path the function will take, so we will continue to just use the heavier weight of the two for which is `liquidate_by_auction`


xlc:
Given the worse case scenario is significantly worse than the average scenario, we need to refund the extra weights so we can pack more liquidation transaction in a single block

ferrell-code:
move to #1609

PR #1609, opened 2021-11-12
Cdp Treasury Bench update

ferrell-code:


runtime/acala/src/weights/module_cdp_treasury.rs
```diff
@@ -48,16 +48,16 @@ use sp_std::marker::PhantomData;
 pub struct WeightInfo<T>(PhantomData<T>);
 impl<T: frame_system::Config> module_cdp_treasury::WeightInfo for WeightInfo<T> {
 	fn auction_collateral() -> Weight {
-		(66_209_000 as Weight)
+		(2_908_193_000 as Weight)
 			.saturating_add(T::DbWeight::get().reads(6 as Weight))
-			.saturating_add(T::DbWeight::get().writes(7 as Weight))
+			.saturating_add(T::DbWeight::get().writes(304 as Weight))
 	}
```
ferrell-code, line 54: We need to adjust `MaxAuctionsCount` so the worst case scenario has less writes, currently a value of 100 in runtimes. Alternatively we could pass max_auction_count as a argument for `auction_collateral` and the weight be a function of that.

modules/cdp-treasury/src/lib.rs
```diff
@@ -177,24 +177,24 @@ pub mod module {
 			Ok(())
 		}
 
-		#[pallet::weight(T::WeightInfo::auction_collateral())]
+		#[pallet::weight(T::WeightInfo::auction_collateral(10))]
```
xlc, line 180:
```suggestion
		#[pallet::weight(T::WeightInfo::auction_collateral(T::MaxAuctionsCount()))]
```

modules/cdp-engine/src/lib.rs
```diff
@@ -112,8 +112,9 @@ type ChangeBalance = Change<Balance>;
 /// Liquidation strategy available
 #[derive(Encode, Decode, Clone, RuntimeDebug, PartialEq, Eq, TypeInfo)]
 pub enum LiquidationStrategy {
-	/// Liquidation CDP's collateral by create collateral auction
-	Auction,
+	/// Liquidation CDP's collateral by create collateral auction, u32 represents number of auctions
+	/// created
+	Auction(u32),
 	/// Liquidation CDP's collateral by swap with DEX
```
ferrell-code, line 118: I believe this type would need to be updated in acala.js, if this is reaasonable

xlc: no need to update types with metadata v14

modules/cdp-engine/src/lib.rs
```diff
@@ -112,8 +112,9 @@ type ChangeBalance = Change<Balance>;
 /// Liquidation strategy available
 #[derive(Encode, Decode, Clone, RuntimeDebug, PartialEq, Eq, TypeInfo)]
 pub enum LiquidationStrategy {
-	/// Liquidation CDP's collateral by create collateral auction
-	Auction,
+	/// Liquidation CDP's collateral by create collateral auction, u32 represents number of auctions
+	/// created
+	Auction(u32),
```
xlc, line 117: maybe this is more self-documenting?
```suggestion
	Auction { count: u32 },
```

xlc (approved):

xlc (approved):

xlc:
can you do a follow up PR to reduce MaxAuctionsCount to 50? just to reduce the weight of worse case scenario
for 10M liquidation (which will means we have some other issue to deal with), each auction size will be 200k, which is still reasonably small.

ferrell-code:
At 10M liquidation I think a lot of people have issues to deal with :smile:, makes sense

PR #1619, opened 2021-11-17
Fixed a redeem bug and added more Homa lite tests

syan095:
Fixed a bug where when redeeming with available staking balance the staking total is not updated
Refactored the code so all redeem logic shares the same code
Added a new mock that have no fees to test economic model
Added tests to ensure that when no fees are charged, the exchange rate says the same throughout all operations.
Modified `RedeemRequested` event to include the fee being charged.

### DIFF
--- a/modules/homa-lite/src/lib.rs
+++ b/modules/homa-lite/src/lib.rs
@@ -707,7 +707,7 @@ pub mod module {
 			let liquid_amount_can_be_redeemed = min(request_amount, *liquid_amount_remaining);
 
 			// Ensure the redeemer have enough liquid currency in their account.
-			if T::Currency::reserved_balance(T::LiquidCurrencyId::get(), &redeemer) >= liquid_amount_can_be_redeemed {
+			if T::Currency::reserved_balance(T::LiquidCurrencyId::get(), redeemer) >= liquid_amount_can_be_redeemed {
 				let new_amount = request_amount.saturating_sub(liquid_amount_can_be_redeemed);
 				*liquid_amount_remaining = liquid_amount_remaining.saturating_sub(liquid_amount_can_be_redeemed);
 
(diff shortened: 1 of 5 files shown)


modules/homa-lite/src/lib.rs
```diff
@@ -488,74 +492,42 @@ pub mod module {
...
-				// Unredeemed requests are added to a queue.
-				if Self::liquid_amount_is_above_minimum_threshold(liquid_remaining) {
-					// Check if there's already a queued redeem request.
-					let (request_amount, _) = Self::redeem_requests(&who).unwrap_or((0, Permill::default()));
-
-					match liquid_remaining.cmp(&request_amount) {
-						// Lock more liquid currency.
-						Ordering::Greater => T::Currency::reserve(
+						liquid_amount.saturating_sub(old_amount),
+					),
```
xlc, line 502: why not just error return here?
```suggestion
					)?,
```

modules/homa-lite/src/lib.rs
```diff
@@ -473,7 +476,8 @@ pub mod module {
 				Error::<T>::AmountBelowMinimumThreshold
 			);
 
-			RedeemRequests::<T>::try_mutate(&who, |request| -> DispatchResult {
+			// Deduct base withdraw fee and add the redeem request to the queue.
+			let _ = RedeemRequests::<T>::try_mutate(&who, |request| -> DispatchResult {
```
xlc, line 480: why ignore error?

modules/homa-lite/src/lib.rs
```diff
@@ -488,74 +492,42 @@ pub mod module {
...
+					additional_fee,
+					base_withdraw_fee,
+				));
 
 				Ok(())
-			})
+			});
+
+			// With redeem request added to the queue, try to redeem it with available staking balance.
+			let _ = Self::process_redeem_requests_with_available_staking_balance(1)?;
```
xlc, line 529: why ignore result?

modules/homa-lite/src/lib.rs
syan095: true. I will add more test to ensure this will fail when the user has insufficent fund

wangjj9219 (approved):

Issue #1621, opened 2021-11-18
ForeignAsset needs to implement xcm_executor::Config::Trader


zjb0807:
https://github.com/AcalaNetwork/Acala/blob/19709245d3353c094820389c57f45c6d4ed94b0b/runtime/karura/src/lib.rs#L1502-L1509

xlc:
I think we will do something similar to statemine.

https://github.com/paritytech/substrate/blob/3d4f1866ebb3cc73cceb12548b15775cedd3e475/frame/assets/src/types.rs#L200-L235

Use min balance ratio to determine how much we charge for fee.

PR #1623, opened 2021-11-19
Implement FixedRateOfForeignAsset

zjb0807:
Closes: #1621


modules/asset-registry/src/lib.rs
```diff
@@ -275,6 +280,88 @@ impl<T: Config> ForeignAssetIdMapping<ForeignAssetId, MultiLocation, AssetMetada
 	}
 }
 
+pub struct FixedRateOfForeignAsset<T, F: Get<u128>, R: TakeRevenue>(
+	Weight,
+	u128,
+	Option<MultiLocation>,
+	PhantomData<(T, F, R)>,
+);
```
shaunxw, line 288: Use fields for readability, and we need more expressive generic params naming.

modules/asset-registry/src/lib.rs
```diff
@@ -275,6 +280,88 @@ impl<T: Config> ForeignAssetIdMapping<ForeignAssetId, MultiLocation, AssetMetada
 	}
 }
 
+pub struct FixedRateOfForeignAsset<T, F: Get<u128>, R: TakeRevenue>(
```
shaunxw, line 283: Need docstring.

modules/asset-registry/src/lib.rs
```diff
@@ -275,6 +280,88 @@ impl<T: Config> ForeignAssetIdMapping<ForeignAssetId, MultiLocation, AssetMetada
...
+
+impl<T: Config, F: Get<u128>, R: TakeRevenue> WeightTrader for FixedRateOfForeignAsset<T, F, R> {
+	fn new() -> Self {
+		Self(0, 0, None, PhantomData)
+	}
+
+	fn buy_weight(&mut self, weight: Weight, payment: Assets) -> Result<Assets, XcmError> {
+		log::trace!(target: "asset-registry::weight", "buy_weight weight: {:?}, payment: {:?}", weight, payment);
+
+		for (index, multi_asset) in payment.fungible.iter().enumerate() {
```
shaunxw, line 298: You can use `fungible.iter().next()` to get the first element.

modules/asset-registry/src/lib.rs
```diff
@@ -275,6 +280,88 @@ impl<T: Config> ForeignAssetIdMapping<ForeignAssetId, MultiLocation, AssetMetada
...
+						return Ok(unused);
+					}
+				}
+				AssetId::Abstract(_) => {
+					// do nothing
+				}
+			}
+		}
+
+		log::trace!(target: "asset-registry::weight", "Did not match");
```
shaunxw, line 335:
```suggestion
		log::trace!(target: "asset-registry::weight", "no concrete fungible asset");
```

modules/asset-registry/src/lib.rs
```diff
@@ -275,6 +280,88 @@ impl<T: Config> ForeignAssetIdMapping<ForeignAssetId, MultiLocation, AssetMetada
...
+		for (index, multi_asset) in payment.fungible.iter().enumerate() {
+			// TODO: only support first fungible assets now.
+			// Maybe be removed in the future.
+			if index > 0 {
+				log::trace!(target: "asset-registry::weight", "buy_weight only support one fungible assets");
+				return Err(XcmError::TooExpensive);
+			}
+
+			match multi_asset.0 {
+				AssetId::Concrete(ref multi_location) => {
```
shaunxw, line 307: `if let` would be more idiomatic

modules/asset-registry/src/lib.rs
```diff
@@ -275,6 +280,88 @@ impl<T: Config> ForeignAssetIdMapping<ForeignAssetId, MultiLocation, AssetMetada
...
+					log::debug!(target: "asset-registry::weight", "buy_weight multi_location: {:?}", multi_location);
+					if let Some(CurrencyId::ForeignAsset(_)) =
+						Pallet::<T>::location_to_currency_ids(multi_location.clone())
+					{
+						let amount = F::get() * (weight as u128) / (WEIGHT_PER_SECOND as u128);
```
shaunxw, line 312: use safe math

modules/asset-registry/src/lib.rs
```diff
@@ -275,6 +280,88 @@ impl<T: Config> ForeignAssetIdMapping<ForeignAssetId, MultiLocation, AssetMetada
...
+		}
+
+		log::trace!(target: "asset-registry::weight", "Did not match");
+		Err(XcmError::TooExpensive)
+	}
+
+	fn refund_weight(&mut self, weight: Weight) -> Option<MultiAsset> {
+		log::trace!(target: "asset-registry::weight", "refund_weight weight: {:?}, weight: {:?}, amount: {:?}, multi_location: {:?}", weight, self.0, self.1, self.2);
+		let weight = weight.min(self.0);
+		let amount = F::get() * (weight as u128) / (WEIGHT_PER_SECOND as u128);
```
shaunxw, line 342: Use same math.

modules/asset-registry/src/lib.rs
```diff
@@ -275,6 +280,88 @@ impl<T: Config> ForeignAssetIdMapping<ForeignAssetId, MultiLocation, AssetMetada
...
+
+		self.0 = self.0.saturating_sub(weight);
+		self.1 = self.1.saturating_sub(amount);
+
+		log::trace!(target: "asset-registry::weight", "refund_weight amount: {:?}", amount);
+		if amount > 0 && self.2.is_some() {
+			Some((self.2.as_ref().expect("checked; qed").clone(), amount).into())
```
shaunxw, line 349: `checked; qed` doesn't contain any information

modules/asset-registry/src/lib.rs
```diff
@@ -275,6 +280,88 @@ impl<T: Config> ForeignAssetIdMapping<ForeignAssetId, MultiLocation, AssetMetada
...
+			None
+		}
+	}
+}
+
+impl<T, F: Get<u128>, R: TakeRevenue> Drop for FixedRateOfForeignAsset<T, F, R> {
+	fn drop(&mut self) {
+		log::trace!(target: "asset-registry::weight", "take revenue, weight: {:?}, amount: {:?}, multi_location: {:?}", self.0, self.1, self.2);
+		if self.1 > 0 && self.2.is_some() {
+			R::take_revenue((self.2.as_ref().expect("checked; qed").clone(), self.1).into());
```
shaunxw, line 360: same here

modules/asset-registry/src/lib.rs
zjb0807: Will there be multiple Assets here? I want to support all, but for the time being only support the first one

shaunxw: We can always update the impl if multiple assets support is needed in the future.

modules/asset-registry/src/lib.rs
shaunxw: tuple struct is not readable with 3 elements like `self.0`, `self.1`, `self.2`

modules/asset-registry/src/lib.rs
```diff
@@ -275,6 +280,105 @@ impl<T: Config> ForeignAssetIdMapping<ForeignAssetId, MultiLocation, AssetMetada
 	}
 }
 
+/// Simple fee calculator that requires payment in a single fungible at a fixed rate.
+///
+/// The constant `FixedRate` type parameter should be the concrete fungible ID and the amount of it
+/// required for one second of weight.
+pub struct FixedRateOfForeignAsset<T, FixedRate: Get<u128>, R: TakeRevenue> {
```
shaunxw, line 287: Is the generic param `T` used somewhere?

runtime/integration-tests/src/relaychain/kusama_cross_chain_transfer.rs
```diff
@@ -154,3 +154,155 @@ fn transfer_to_sibling() {
 		assert_eq!(Tokens::free_balance(BNC, &AccountId::from(ALICE)), 94_989_760_000_000);
 	});
 }
+
+#[test]
+fn test_asset_registry_module() {
+	TestNet::reset();
+
+	fn sibling_2000_account() -> AccountId {
```
shaunxw, line 162:
```suggestion
	fn karura_reserve_account() -> AccountId {
```

shaunxw (approved):
LGTM. Some final nitpicks.

modules/asset-registry/src/lib.rs
zjb0807: https://github.com/AcalaNetwork/Acala/blob/21dfd9edca4012dd730147ca32227fb019709531/modules/asset-registry/src/lib.rs#L316

used here.

xlc: the rate will be different for each asset. it should read minimal balance of the asset and use ratio between native token ED to determine the weight per second

xlc (approved):

modules/asset-registry/src/lib.rs
```diff
@@ -275,6 +283,132 @@ impl<T: Config> ForeignAssetIdMapping<ForeignAssetId, MultiLocation, AssetMetada
...
+					let ed = T::Currency::minimum_balance();
+
+					log::trace!(
+						target: "asset-registry::weight",
+						"buy_weight FixedRate: {:?}, asset minimal_balance: {:?}, native token ed: {:?}",
+						FixedRate::get(), asset_metadatas.minimal_balance, ed
+					);
+
+					ensure!(!ed.is_zero(), XcmError::TooExpensive);
+					ensure!(!WEIGHT_PER_SECOND.is_zero(), XcmError::TooExpensive);
```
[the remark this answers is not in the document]
zjb0807: Ok. I'll remove the `ed` too. The integration tests can ensure it is non-zero.

xlc (dismissed):

xlc (approved):

Issue #1561, opened 2021-10-30
Test for XCM barrier


xlc:
Followup of #1560 

Needs to have some test to verify it works

zqhxuyuan:
As cumlus PR, different runtime have different AllowSubscriptionsFrom, for example: Statemine and Statemint use ParentOrSiblings, while Rococo/Westmint use Everything, we may need testcase to make sure we use the right configuration.

PR #1624, opened 2021-11-19
test barrier

zqhxuyuan:
closes https://github.com/AcalaNetwork/Acala/issues/1561


xlc (commented):
I think we should align the xcm config between 3 runtimes to ensure they have similar behaviour.

zqhxuyuan:
besides Trader config may different as different runtime has different FixedRateOfFungible asset, for now I just find `UnitWeightCost` base weight in Weigher is different between 3 runtimes. for align xcm config, I'll create another PR to do and find if there're other different poins.

zqhxuyuan:
Ah, I think most different point between 3 runtime is asset related config, e.g. `CurrentIdConvert`

shaunxw (commented):
These look like *unit tests* for barrier structs to me.

xlc:
This is testing https://github.com/AcalaNetwork/Acala/blob/0e3c5ed2d56c097e602f0b8667cda4127198c873/runtime/karura/src/lib.rs#L1457-L1464 is correctly configured.

zqhxuyuan:
I was thinking restrict `SubscribeVersion from` with whitelist list instead Everything. is it worth?

shaunxw:
> 



> This is testing
> 
> https://github.com/AcalaNetwork/Acala/blob/0e3c5ed2d56c097e602f0b8667cda4127198c873/runtime/karura/src/lib.rs#L1457-L1464
> is correctly configured.

Then we should test against this `Barrier` type, instead of structs like `TakeWeightCredit`, `AllowTopLevelPaidExecutionFrom` etc.

xlc:
> I was thinking restrict `SubscribeVersion from` with whitelist list instead Everything. is it worth?

That will be a useful thing to do. Can happen in later PR. Only allow parent and sibling parachains.

zqhxuyuan:
> 

there're some `XcmExecutor::<XcmConfig>::execute_xcm(...)` test which is like _cross-chain transfer tests_. I add test code here(runtime.rs) because it's runtime config.  I'll add real cross chain barrier_test into `kusama_cross_chain_transfer.rs`

shaunxw:
> I add test code here(runtime.rs) because it's runtime config.

Tests like

```rust
let r = TakeWeightCredit::should_execute(&Parent.into(), &mut message, expect_weight, &mut weight_credit);
assert_eq!(r, Err(()));
assert_eq!(weight_credit, 400_000_000);
```

doesn't look related to our runtime config. It's more like a unit test to check if `TakeWeightCredit` struct is implemented correctly.

> I'll add real cross chain barrier_test into `kusama_cross_chain_transfer.rs`

Then there will be duplicated tests for the same scenarios, why not only add the real cross chain barrier tests here? I don't think this would help with test coverage, but will bring maintenance burden to us instead.

shaunxw (dismissed):

xlc (dismissed):

zjb0807 (approved):

PR #1626, opened 2021-11-20
Add improved homa lite redeem match

### LINKED ISSUE #1582: Improve homa-lite request matching
Every time matching the `RedeemRequests`, remember the next unmatched request, and start iterating from this one instead from beginning.

This doesn't guarantee strict FIFO order but provide a good enough one so we won't have unlucky requests never got matched.

syan095:
After redeem requests are processed, the "next" item is stored. When the next mint/unbond/adjust_available_staking happens, redeem requests are iterated from the "next" item. This ensures all redeem requests are "processed" before wrapping around.

Closes #1582


modules/homa-lite/src/lib.rs
```diff
@@ -1008,6 +1042,54 @@ pub mod module {
...
+
+			let starting_element = match Self::redeem_requests(Self::next_redeem_request_to_match()) {
+				Some(request) => (Self::next_redeem_request_to_match(), request),
+				None => first_element,
+			};
+			let mut iterator = RedeemRequests::<T>::iter();
+			let mut current = iterator.next();
+
+			// Skip elements until `starting_element` is found.
+			while current.is_some() && current.clone().unwrap_or_default() != starting_element {
```
xlc, line 1065: No this is very inefficient.

Do something like this

https://github.com/AcalaNetwork/Acala/blob/7f67fb914b19347eccd03d042daaa4b902b816e3/modules/cdp-engine/src/lib.rs#L664-L669

xlc (changes requested):

modules/homa-lite/src/lib.rs
syan095: Replaced the first iter with iter_from instead

modules/homa-lite/src/lib.rs
```diff
@@ -260,6 +260,12 @@ pub mod module {
 	#[pallet::getter(fn staking_interest_rate_per_update)]
 	pub type StakingInterestRatePerUpdate<T: Config> = StorageValue<_, Permill, ValueQuery>;
 
+	/// Next redeem request to iterate from when matching redeem requests.
+	/// NextRedeemRequestToMatch: Value: T::AccountId
+	#[pallet::storage]
+	#[pallet::getter(fn next_redeem_request_to_match)]
+	pub type NextRedeemRequestToMatch<T: Config> = StorageValue<_, T::AccountId, ValueQuery>;
```
xlc, line 267: this should store the raw key instead of AccountId. otherwise if this account request is redeemed or cancelled, the match will start from beginning again.

xlc (changes requested):

modules/homa-lite/src/lib.rs
syan095: Changed the storage to use key instead.
This storage is change to store the last redeem request key, since iter_from exclude from the input key

modules/homa-lite/src/lib.rs
```diff
@@ -783,44 +784,59 @@ pub mod module {
...
+				// Define the function to be executed for each of the redeem request:
+				// Redeem the given requests with the current minting request.
+				//
+				// Params: redeemer: T::AccountId, request_amount: Balance, extra_fee: Permill
+				// Returns: Result<(new_redeem_amount, should_break), Error>
+				//
+				// Capturing: &mut liquid_remaining
+				//            &mut redeem_requests_limit_remaining
+				// If all the currencies are minted, return `should_break` as true.
+				let mut f = |redeemer, request_amount, extra_fee| -> Result<(bool, Balance), DispatchError> {
```
shaunxw, line 818: Capturing mutable variables in closure is not readable and prone to bugs. Please refactor it to be no closure and mutable params.

modules/homa-lite/src/lib.rs
```diff
@@ -908,16 +924,31 @@ pub mod module {
...
+			{
+				// Define the function to be executed for each of the redeem request:
+				// Redeem the given requests using available_staking_balance.
+				//
+				// Params: redeemer: T::AccountId, request_amount: Balance, extra_fee: Permill
+				// Returns: Result<(should_break, new_redeem_amount), Error>
+				//
+				// Capturing: &mut available_staking_balance
+				//            &mut num_matched
+				let mut f = |redeemer, request_amount, _extra_fee| -> Result<(bool, Balance), DispatchError> {
```
shaunxw, line 938: Same here.

wangjj9219 (dismissed):

modules/homa-lite/src/lib.rs
```diff
@@ -260,6 +260,12 @@ pub mod module {
 	#[pallet::getter(fn staking_interest_rate_per_update)]
 	pub type StakingInterestRatePerUpdate<T: Config> = StorageValue<_, Permill, ValueQuery>;
 
+	/// Next redeem request to iterate from when matching redeem requests.
+	/// RedeemRequestKeyToIterFrom: Value: T::AccountId
+	#[pallet::storage]
+	#[pallet::getter(fn redeem_request_key_to_iter_from)]
+	pub type RedeemRequestKeyToIterFrom<T: Config> = StorageValue<_, Vec<u8>, ValueQuery>;
```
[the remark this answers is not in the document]
syan095: Due to the way iterator work, the stored key is not the "next" item per see. I've renamed this to be more accurate.

modules/homa-lite/src/lib.rs
syan095: Agree. WIth the closure and captures the whole thing is a mess.
I've significantly refactored the code structure. Please have a look at the updated structure.
Al closures are removed.

modules/homa-lite/src/lib.rs
```diff
@@ -260,6 +266,12 @@ pub mod module {
 	#[pallet::getter(fn staking_interest_rate_per_update)]
 	pub type StakingInterestRatePerUpdate<T: Config> = StorageValue<_, Permill, ValueQuery>;
 
+	/// Next redeem request to iterate from when matching redeem requests.
+	/// LastRedeemRequestKeyIterated: Value: T::AccountId
```
xlc, line 270:
```suggestion
	/// LastRedeemRequestKeyIterated: Value: Vec<u8>
```

xlc (dismissed):

xlc (approved):

PR #1630, opened 2021-11-22
Increase the minimum mint and redeem for HomaLite

syan095:
Updated integration tests accordingly.

### DIFF
--- a/runtime/acala/src/lib.rs
+++ b/runtime/acala/src/lib.rs
@@ -1537,8 +1537,8 @@ pub fn create_x2_parachain_multilocation(index: u16) -> MultiLocation {
 parameter_types! {
 	pub const DOTCurrencyId: CurrencyId = CurrencyId::Token(TokenSymbol::DOT);
 	pub const LDOTCurrencyId: CurrencyId = CurrencyId::Token(TokenSymbol::LDOT);
-	pub MinimumMintThreshold: Balance = 2 * dollar(DOT);
-	pub MinimumRedeemThreshold: Balance = 20 * dollar(LDOT);
+	pub MinimumMintThreshold: Balance = 50 * cent(DOT);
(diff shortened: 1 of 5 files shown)


runtime/acala/src/lib.rs
```diff
@@ -1537,8 +1537,8 @@ pub fn create_x2_parachain_multilocation(index: u16) -> MultiLocation {
 parameter_types! {
 	pub const DOTCurrencyId: CurrencyId = CurrencyId::Token(TokenSymbol::DOT);
 	pub const LDOTCurrencyId: CurrencyId = CurrencyId::Token(TokenSymbol::LDOT);
-	pub MinimumMintThreshold: Balance = 2 * dollar(DOT);
-	pub MinimumRedeemThreshold: Balance = 20 * dollar(LDOT);
+	pub MinimumMintThreshold: Balance = 50 * cent(DOT);
+	pub MinimumRedeemThreshold: Balance = 5 * dollar(LDOT);
```
xlc, line 1541: why are you reducing value here. it should be 10x the KSM value as KSM is ~10x the price of DOT

runtime/mandala/src/lib.rs
```diff
@@ -1250,15 +1250,15 @@ pub fn create_x2_parachain_multilocation(index: u16) -> MultiLocation {
 }
 
 parameter_types! {
-	pub MinimumMintThreshold: Balance = 10 * cent(DOT);
-	pub MinimumRedeemThreshold: Balance = 100 * cent(LDOT);
+	pub MinimumMintThreshold: Balance = 50 * cent(DOT);
```
xlc, line 1253: the value should match to acala

xlc (changes requested):

xlc (approved):

PR #1631, opened 2021-11-22
new stable-asset

frankyin-factual:


zjb0807 (commented):
```
cargo fmt --all
```

frankyin-factual:
@zjb0807 should be fixed our our submodule.

zjb0807 (approved):

Issue #736, opened 2021-02-09
EVM: improve AllPrecompiles type


xlc:
https://github.com/AcalaNetwork/Acala/blob/56be4f1d887bf2d6f579069ab214824149608cd5/runtime/common/src/precompile/mod.rs#L38-L54

This is starting become unmanageable and we will have many more precompiles in future

PR #1636, opened 2021-11-24
refactor AllPrecompiles

zqhxuyuan:
closes https://github.com/AcalaNetwork/Acala/issues/736

`ScheduleCallPrecompile` needs special attention as it deals with `Call, Origin, PalletsOrigin` trait Bounds.

### DIFF
--- a/runtime/common/src/precompile/schedule_call.rs
+++ b/runtime/common/src/precompile/schedule_call.rs
@@ -29,7 +29,7 @@ use frame_support::{
 	},
 };
 use module_evm::{Context, ExitError, ExitSucceed, Precompile};
-use module_support::{AddressMapping as AddressMappingT, Erc20InfoMapping as Erc20InfoMappingT, TransactionPayment};
+use module_support::{AddressMapping, TransactionPayment};
 use primitives::{Balance, BlockNumber};
 use sp_core::H160;
 use sp_runtime::RuntimeDebug;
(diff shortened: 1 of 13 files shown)


zjb0807 (approved):
LGTM

runtime/common/src/precompile/schedule_call.rs
```diff
@@ -99,46 +77,36 @@ type PalletBalanceOf<T> =
...
+		+ Sync,
+	<Runtime as pallet_scheduler::Config>::Call: Dispatchable + Debug + From<module_evm::Call<Runtime>>,
+	<<Runtime as pallet_scheduler::Config>::Call as Dispatchable>::Origin: IsType<<Runtime as frame_system::Config>::Origin>
+		+ OriginTrait<
+			AccountId = Runtime::AccountId,
+			PalletsOrigin = <Runtime as pallet_scheduler::Config>::PalletsOrigin,
+		>,
+	pallet_scheduler::Pallet<Runtime>: ScheduleNamed<
+		BlockNumber,
+		<Runtime as pallet_scheduler::Config>::Call,
```
zqhxuyuan, line 99: both `<Runtime as pallet_scheduler::Config>::Call` and `<Runtime as frame_system::Config>::Call`  works here, which one is better, or just doesn't matter because they use the same `Call`.

runtime/common/src/precompile/schedule_call.rs
```diff
@@ -213,12 +185,18 @@ impl<
...
+					BlockNumber,
+					<Runtime as pallet_scheduler::Config>::Call,
+					<Runtime as pallet_scheduler::Config>::PalletsOrigin,
+				>>::schedule_named(
 					task_id.clone(),
 					DispatchTime::After(min_delay),
 					None,
 					0,
-					Origin::root().caller().clone(),
+					<<<Runtime as pallet_scheduler::Config>::Call as Dispatchable>::Origin>::root()
```
zqhxuyuan, line 197: do we need use `ScheduleOrigin` inside `pallet_scheduler::Config`, or just use `root()` here?

runtime/common/src/precompile/schedule_call.rs
xlc: this is fine

runtime/common/src/precompile/schedule_call.rs
xlc: this is fine. later we may want to create an EVM Scheduler origin if we want to avoid use root here.
but user have no control on the call so it is safe.

xlc (approved):

PR #1648, opened 2021-11-28
refactor homa

wangjj9219:


modules/homa/src/lib.rs
```diff
@@ -16,118 +16,962 @@
...
 		#[transactional]
 		pub fn mint(origin: OriginFor<T>, #[pallet::compact] amount: Balance) -> DispatchResult {
-			let who = ensure_signed(origin)?;
-			T::Homa::mint(&who, amount)?;
+			let minter = ensure_signed(origin)?;
+
+			// Ensure the amount is above the mint threshold.
+			ensure!(amount > T::MintThreshold::get(), Error::<T>::BelowMintThreshold);
+
+			// TODO: is that neccesary?
```
xlc, line 324: yeah I think we shuold still have a soft cap

modules/homa/src/lib.rs
```diff
@@ -16,118 +16,962 @@
...
+			ensure!(
+				Self::get_total_staking_currency().saturating_add(amount) <= Self::get_staking_currency_soft_cap(),
+				Error::<T>::ExceededStakingCurrencySoftCap
+			);
+
+			T::Currency::transfer(T::StakingCurrencyId::get(), &minter, &Self::account_id(), amount)?;
+			ToBondPool::<T>::mutate(|pool| *pool = pool.saturating_add(amount));
+
+			// calculate the liquid amount by the current exchange rate.
+			let liquid_amount = Self::convert_staking_to_liquid(amount)?;
```
xlc, line 334: this depends on `ToBondPool` right, so we should read it first before modify `ToBondPool`

should always read exchange rate before any actions to ensure the rate is not partially updated

wangjj9219:
still need to add some integration tests

modules/homa/src/lib.rs
```diff
@@ -16,118 +16,1065 @@
...
+	#[pallet::storage]
+	#[pallet::getter(fn soft_bonded_cap_per_sub_account)]
+	pub type SoftBondedCapPerSubAccount<T: Config> = StorageValue<_, Balance, ValueQuery>;
+
+	/// Th staking amount of threshold to mint.
+	///
+	/// MintThreshold: value: Balance
+	#[pallet::storage]
+	#[pallet::getter(fn mint_threshold)]
+	pub type MintThreshold<T: Config> = StorageValue<_, Balance, ValueQuery>;
```
xlc, line 287: This is something we unlikely need to change, so could make it a constant instead.

modules/homa/src/lib.rs
```diff
@@ -16,118 +16,1065 @@
...
+
+	/// Th liquid amount of threshold to redeem.
+	///
+	/// RedeemThreshold: value: Balance
+	#[pallet::storage]
+	#[pallet::getter(fn redeem_threshold)]
+	pub type RedeemThreshold<T: Config> = StorageValue<_, Balance, ValueQuery>;
```
xlc, line 294: This is something we unlikely need to change, so could make it a constant instead.

modules/homa/src/lib.rs
```diff
@@ -16,118 +16,1065 @@
...
+	#[pallet::storage]
+	#[pallet::getter(fn fast_match_fee_rate)]
+	pub type FastMatchFeeRate<T: Config> = StorageValue<_, Rate, ValueQuery>;
+
+	/// The prefix of block numbers to bump local current era.
+	///
+	/// BumpEraPrefix: value: BlockNumber
+	#[pallet::storage]
+	#[pallet::getter(fn bump_era_prefix)]
+	pub type BumpEraPrefix<T: Config> = StorageValue<_, T::BlockNumber, ValueQuery>;
```
xlc, line 316:
```suggestion
	pub type BumpEraOffset<T: Config> = StorageValue<_, T::BlockNumber, ValueQuery>;
```

modules/homa/src/lib.rs
```diff
@@ -16,118 +16,1065 @@
...
+	pub type BumpEraFrequency<T: Config> = StorageValue<_, T::BlockNumber, ValueQuery>;
+
 	#[pallet::pallet]
 	pub struct Pallet<T>(_);
 
 	#[pallet::hooks]
-	impl<T: Config> Hooks<T::BlockNumber> for Pallet<T> {}
+	impl<T: Config> Hooks<T::BlockNumber> for Pallet<T> {
+		fn on_initialize(n: T::BlockNumber) -> Weight {
+			if Self::should_bump_local_current_era(n) {
```
xlc, line 329: I think we should use relay chain block number for better accuracy.
The relay chain number in `on_initialize` is the one from previous block, but should be fine.

modules/homa/src/lib.rs
```diff
@@ -45,6 +47,8 @@ pub mod weights;
 pub mod module {
 	use super::*;
 
+	pub type RelayChainBlockNumberOf<T> = <<T as Config>::RelayChainBlockNumber as BlockNumberProvider>::BlockNumber;
```
xlc, line 50: I don't think this is needed? `T::BlockNumber` should be enough?

modules/homa/src/lib.rs
```diff
@@ -968,11 +956,11 @@ pub mod module {
...
-		pub fn should_bump_local_current_era(block_number: T::BlockNumber) -> bool {
-			block_number
-				.checked_add(&Self::bump_era_prefix())
-				.and_then(|n| n.checked_rem(&Self::bump_era_frequency()))
-				.map_or(false, |n| n.is_zero())
+		pub fn should_bump_local_current_era(relaychain_block_number: RelayChainBlockNumberOf<T>) -> bool {
+			relaychain_block_number
+				.checked_sub(&Self::last_era_bumped_block())
+				.and_then(|n| n.checked_div(&Self::bump_era_frequency()))
+				.map_or(false, |n| n >= One::one())
```
xlc, line 963: do we need to support bump multiple eras? it will happen when network stalled for more than one era

modules/homa/src/lib.rs
```diff
@@ -956,20 +954,21 @@ pub mod module {
...
 
 		/// Bump current era.
 		/// The rebalance will send XCM messages to relaychain. Once the XCM message is sent,
 		/// the execution result cannot be obtained and cannot be rolled back. So the process
 		/// of rebalance is not atomic.
-		pub fn bump_current_era() -> DispatchResult {
+		pub fn bump_current_era(amount: EraIndex) -> DispatchResult {
 			let previous_era = Self::relay_chain_current_era();
-			let new_era = previous_era + 1;
+			let new_era = previous_era + amount;
```
xlc, line 971: use saturated add

xlc (approved):

Issue #1653, opened 2021-11-30
implements own `DropAssets` that deposit assets too low to treasury


zqhxuyuan:
https://github.com/paritytech/polkadot/issues/4402

PR #1654, opened 2021-11-30
Xcm trap assets ed

zqhxuyuan:
closes https://github.com/AcalaNetwork/Acala/issues/1653

- [x] integration test
   - [x] relaychain to parachain
   - [x] slibling to parachain
- [x] other runtime add `AcalaDropAssets` too
   - [x] ~~maybe refactor common code to `common` package?~~
- [ ] ~~the kusama_cross_chain_transfer.rs getting more large, should we seperate to different file?~~

### DIFF
--- a/runtime/acala/src/lib.rs
+++ b/runtime/acala/src/lib.rs
@@ -58,7 +58,8 @@ use module_relaychain::RelayChainCallBuilder;
 use module_support::{DispatchableTask, ForeignAssetIdMapping};
 use module_transaction_payment::{Multiplier, TargetedFeeAdjustment};
 use orml_traits::{
-	create_median_value_data_provider, parameter_type_with_key, DataFeeder, DataProviderExtended, MultiCurrency,
+	create_median_value_data_provider, parameter_type_with_key, DataFeeder, DataProviderExtended, GetByKey,
+	MultiCurrency,
 };
 use pallet_transaction_payment::RuntimeDispatchInfo;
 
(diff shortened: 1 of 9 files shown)


runtime/acala/src/lib.rs
```diff
@@ -728,6 +729,31 @@ impl DataFeeder<CurrencyId, Price, AccountId> for AggregatedDataProvider {
 	}
 }
 
+pub struct ExistentialDepositsForDropAssets;
+impl ExistentialDepositsForDropAssets {
+	fn get(currency_id: &CurrencyId) -> Balance {
+		match currency_id {
+			CurrencyId::Token(symbol) => match symbol {
+				TokenSymbol::ACA => NativeTokenExistentialDeposit::get(),
```
xlc, line 737: can we reuse the values instead of duplicate them?
i.e. only handle ACA and have default arm send to the other ExistentialDeposits

zqhxuyuan: yeah, doing it now.

runtime/acala/src/lib.rs
```diff
@@ -1429,6 +1442,40 @@ impl TakeRevenue for ToTreasury {
...
+		let mut asset_traps: Vec<MultiAsset> = vec![];
+		for asset in multi_assets {
+			if let MultiAsset {
+				id: Concrete(location),
+				fun: Fungible(amount),
+			} = asset.clone()
+			{
+				let currency_id = CurrencyIdConvert::convert(location);
+				// burn asset(do nothing here) if convert result is None
+				if let Some(currency_id) = currency_id {
```
zqhxuyuan, line 1462: when convert result is None, do nothing has the meaning of burn, right?

xlc: yes. we just burn unknown assets for now. later maybe we can have them in the unknown-tokens pallet

runtime/acala/src/lib.rs
```diff
@@ -1429,6 +1442,40 @@ impl TakeRevenue for ToTreasury {
 	}
 }
 
+pub struct AcalaDropAssets<X, T>(PhantomData<(X, T)>);
+impl<X, T> DropAssets for AcalaDropAssets<X, T>
+where
+	X: DropAssets,
+	T: TakeRevenue,
+{
+	fn drop_assets(origin: &MultiLocation, assets: Assets) -> Weight {
```
xlc, line 1451: this is not a trivial method, so best to abstract it into the common runtime to avoid repeat this 3 times

runtime/common/src/lib.rs
```diff
@@ -331,6 +337,60 @@ pub enum RelayChainSubAccountId {
 	HomaLite = 0,
 }
 
+pub struct AcalaDropAssets<X, T, C, NC, NB, GK>(PhantomData<(X, T, C, NC, NB, GK)>);
```
xlc, line 340: add some docs explain what are NC, NB, GK

xlc (approved):

PR #1657, opened 2021-11-30
Feature/improve collector reward

### LINKED ISSUE #1571: collector-selection add MinRewardDistributeAmount
Currently on Mandala the half of the rewards in the collection selection pot send to currently block author. This can be very low if blocks are empty.

Should just avoid distribute reward if the amount of tokens in pot is too low.

Should be configured to tx fee of an empty remark / 10.

syan095:
Only share the pot with author if current pot's reward is above `MinRewardDistributeAmount`
Added unit tests and integration tests

Closes #1571

### DIFF
--- a/modules/collator-selection/src/lib.rs
+++ b/modules/collator-selection/src/lib.rs
@@ -450,8 +450,8 @@ pub mod pallet {
 				.unwrap_or_default()
 				.div(2u32.into());
 
-			// `reward` is half of pot account minus ED, this should never fail.
 			if reward >= T::MinRewardDistributeAmount::get() {
+				// `reward` is half of pot account minus ED, this should never fail.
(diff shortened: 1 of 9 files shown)


modules/collator-selection/src/lib.rs
```diff
@@ -445,9 +449,12 @@ pub mod pallet {
 				.checked_sub(&T::Currency::minimum_balance())
 				.unwrap_or_default()
 				.div(2u32.into());
+
 			// `reward` is half of pot account minus ED, this should never fail.
```
xlc, line 453: move this comment the line above the transfer otherwise it looks like the `>=` should never fail

xlc (approved):

xlc (approved):

zjb0807 (approved):

PR #1669, opened 2021-12-03
add KBTC and KINT cross-chain assets

sander2:
This adds support for the Kintsugi's KBTC and KINT tokens. Currently marked as draft since we still need to determine a reasonable value for the KINT:KSM ratio.

I tested this with a rococo-local instance.

### DIFF
--- a/runtime/karura/src/lib.rs
+++ b/runtime/karura/src/lib.rs
@@ -1502,7 +1502,7 @@ parameter_types! {
 	pub KintPerSecond: (AssetId, u128) = (
 		MultiLocation::new(
 			1,
-			X2(Parachain(parachains::kintsugi::ID), GeneralKey(parachains::kintsugi::KBTC_KEY.to_vec())),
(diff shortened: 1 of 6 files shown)


runtime/karura/src/lib.rs
```diff
@@ -751,6 +751,8 @@ parameter_type_with_key! {
 				TokenSymbol::BNC => 800 * millicent(*currency_id),  // 80BNC = 1KSM
 				TokenSymbol::VSKSM => 10 * millicent(*currency_id),  // 1VSKSM = 1KSM
 				TokenSymbol::PHA => 4000 * millicent(*currency_id), // 400PHA = 1KSM
+				TokenSymbol::KINT => 13333 * microcent(*currency_id), // 1.33 KINT = 1 KSM
```
shaunxw, line 754:
```suggestion
				TokenSymbol::KINT => 13333 * microcent(*currency_id), // 4/3 KINT = 1 KSM
```

shaunxw (approved):
LGTM

PR #1674, opened 2021-12-06
Benchmark evm

ermalkaleci:


ferrell-code:
fyi, changes to handlebars-rust upstream broke our handlebars templates #1688 will fix

zjb0807 (approved):

ferrell-code (approved):
It all looks reasonable to me. We may want to in the future setup an e2e test filling blocks to the maximum block weight on different evm calls and seeing if block production slows/stalls

Issue #1671, opened 2021-12-06
Enforce minimal tip per weight


xlc:
Add a constant `TipPerWeightStep` and update the effective tip per weight to be `tipPerWeight = tipPerWight / TipPerWeightStep * TipPerWeightStep`.

This will raise the min effective tip cost and increase the cost of front-running.

Issue #1544, opened 2021-10-28
Update transaction payment


xlc:
https://github.com/paritytech/substrate/pull/9834

PR #1682, opened 2021-12-08
Update transaction-payment

zjb0807:
Closes: #1544 
Closes: #1671


runtime/integration-tests/src/runtime.rs
```diff
@@ -276,3 +276,8 @@ mod mandala_only_tests {
 		});
 	}
 }
+
+#[test]
+fn check_max_tips_of_priority() {
+	assert_eq!(MaxTipsOfPriority::get(), 204_358_782_087); // 0.2 KAR/ACA
```
xlc, line 282: this is a bit too low. I was thinking something like 10000 KAR/ACA

modules/transaction-payment/src/mock.rs
```diff
@@ -179,6 +179,9 @@ impl module_dex::Config for Runtime {
 parameter_types! {
 	pub MaxSwapSlippageCompareToOracle: Ratio = Ratio::saturating_from_rational(1, 2);
 	pub static TransactionByteFee: u128 = 1;
+	pub OperationalFeeMultiplier: u64 = 5;
+	pub TipPerWeightStep: u128 = 1;
```
xlc, line 183: make it a higher value and have test to ensure tip value that is less than TipPerWeightStep results no priority change

modules/transaction-payment/src/lib.rs
```diff
@@ -248,6 +248,39 @@ pub mod module {
...
+		/// ```
+		///
+		/// Note that since we use `final_fee` the multiplier applies also to the regular `tip`
+		/// sent with the transaction. So, not only does the transaction get a priority bump based
+		/// on the `inclusion_fee`, but we also amplify the impact of tips applied to `Operational`
+		/// transactions.
+		#[pallet::constant]
+		type OperationalFeeMultiplier: Get<u64>;
+
+		/// The minimum value of tips per weight.
```
xlc, line 275:
```suggestion
		/// The step amount of tips required to effect transaction priority.
```

xlc (approved):

PR #1683, opened 2021-12-08
stable-asset-update

frankyin-factual:
update stable asset benchmarks


ferrell-code:
Well that is an ugly output, I will fix the handlebars file

runtime/mandala/src/lib.rs
```diff
@@ -2588,8 +2591,8 @@ mod tests {
 	#[test]
 	fn check_call_size() {
 		assert!(
-			core::mem::size_of::<Call>() <= 230,
-			"size of Call is more than 230 bytes: some calls have too big arguments, use Box to \
+			core::mem::size_of::<Call>() <= 280,
+			"size of Call is more than 280 bytes: some calls have too big arguments, use Box to \
 			reduce the size of Call.
 			If the limit is too strong, maybe consider increasing the limit",
```
ferrell-code, line 2597: You can Box the argument in a call that causes to the `Call` enum to be too large... In this case the limit is only slightly changed so I don't feel strongly but see https://github.com/open-web3-stack/open-runtime-module-library/pull/598

runtime/mandala/src/benchmarking/nutsfinance_stable_asset.rs
```diff
@@ -78,49 +92,45 @@ runtime_benchmarks! {
...
+		let intial_a = 10000u128;
 		let fee_recipient: AccountId = account("fee", 0, SEED);
-		let _ = StableAsset::create_pool(RawOrigin::Root.into(), pool_asset, assets, precisions, mint_fee, swap_fee, redeem_fee, intial_a, fee_recipient.clone());
+		let yield_recipient: AccountId = account("yield", 1, SEED);
+		let _ = StableAsset::create_pool(RawOrigin::Root.into(), pool_asset, assets, precisions, mint_fee, swap_fee, redeem_fee, intial_a, fee_recipient, yield_recipient, 1000000000000000000u128);
 		let pool_id = StableAsset::pool_count() - 1;
 		set_balance_fungibles(LDOT, &tester, 200000000000u128);
 		set_balance_fungibles(AUSD, &tester, 200000000000u128);
 		let _ = StableAsset::mint(RawOrigin::Signed(tester.clone()).into(), pool_id, vec![10000000000u128, 20000000000u128], 0u128);
 	}: _(RawOrigin::Signed(tester), pool_id, 0, 1, 5000000u128, 0u128)
```
ferrell-code, line 104: swap's weight is a function of the length of `pool_id.balances`. It calls into `update_balance` which reads into storage for each iteration of balances. 
``` rust
fn update_balance(
		pool_info: &mut StableAssetPoolInfo<
			Self::AssetId,
			Self::AtLeast64BitUnsigned,
			Self::Balance,
			Self::AccountId,
			Self::BlockNumber,
		>,
	) -> DispatchResult {
		for (i, balance) in pool_info.balances.iter_mut().enumerate() {
			let balance_of: Self::AtLeast64BitUnsigned =
				T::Assets::balance(pool_info.assets[i], &pool_info.account_id).into();
			*balance = balance_of
				.checked_mul(&pool_info.precisions[i])
				.ok_or(Error::<T>::Math)?
				.into();
		}
		Ok(())
	}
```

runtime/mandala/src/weights/nutsfinance_stable_asset.rs
```diff
@@ -48,43 +48,47 @@ use sp_std::marker::PhantomData;
...
+	fn redeem_proportion(u: u32) -> Weight {
+		(90_887_000 as Weight) // Standard Error: 51_000
+			.saturating_add((21_808_000 as Weight).saturating_mul(u as Weight))
+			.saturating_add(T::DbWeight::get().reads(6 as Weight))
+			.saturating_add(T::DbWeight::get().reads((2 as Weight).saturating_mul(u as Weight)))
+			.saturating_add(T::DbWeight::get().writes(4 as Weight))
+			.saturating_add(T::DbWeight::get().writes((2 as Weight).saturating_mul(u as Weight)))
 	}
 	fn redeem_single() -> Weight {
-		(115_000_000 as Weight)
```
ferrell-code, line 81: redeem also is a function of pool_id.balances.len()

ferrell-code:
Just a few comments... overall looks good!

ferrell-code:
Nit, while I believe this is safe as `usize` and `u32` are the same in the wasm environment I still prefer something like `T::PoolAssetLimit::get().try_into().map_err(|_| ArithmeticError::Overflow)?;`

``` rust
ensure!(
	assets.len() <= T::PoolAssetLimit::get().try_into().unwrap(),
	Error::<T>::ArgumentsError
);
```

runtime/mandala/src/lib.rs
```diff
@@ -1863,8 +1863,9 @@ impl orml_xcm::Config for Runtime {
 }
 
 parameter_types! {
-	pub const Precision: u128 = 1000000000000000000u128; // 18 decimals
 	pub const FeePrecision: u128 = 10000000000u128; // 10 decimals
+	pub const APrecision: u128 = 100u128; // 10 decimals
+	pub const PoolAssetLimit: u32 = 5u32;
```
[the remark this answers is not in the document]
frankyin-factual: fixed

runtime/mandala/src/lib.rs
frankyin-factual: thanks. gonna keep the current approach.

runtime/mandala/src/benchmarking/nutsfinance_stable_asset.rs
frankyin-factual: fixed

ferrell-code (approved):

Issue #1233, opened 2021-07-20
XCM execution fee management


shaunxw:
### XCM execution fees

While receiving cross-chain transfer, the way to collect XCM execution fee depends on the kind of transferred asset:

- Karura/Acala native tokens, namely KAR/ACA: in this case use `WeightToFee` to figure out how much to charge.
- Foreign tokens which are listed in DEX:
  1. Firstly use `WeightToFee` to get the amount of native currency to charge.
  2. Use DEX convert to get required native currency.
- Foreign tokens which are not listed in DEX: we need to estimate a ratio between the foreign token and native tokens, and use it to get weight to foreign token amount.

### Chicken or the egg

Note that for charging foreign tokens via DEX, there is a *chicken or the egg* dilemma: the fee can't be converted in DEX until this foreign token listed, however it can't be listed until cross-chain transfer enabled and liquidity available. Thus for every foreign tokens to be listed in DEX:
1. Firstly we need to estimate a ratio between the foreign token and native token, and use it to charge XCM execution fees.
2. Then we could have cross-chain enabled, liquidity injected. And the foreign token TP could be listed.
3. Once the foreign token to native token trading path is available, replace the method to collect XCM execution fees as using DEX. The estimated ratio would be a back-up way then.

zqhxuyuan:
I think `Trader` config inside `xcm_executor` can do the first case, isn't it? as second and third case about DEX, can `FixedRateOfForeignAsset` deal with? @shaunxw @zjb0807

zqhxuyuan:
current transaction payment is used for sender to charge fee, and trader along with BuyExecution is used for receiver to execute xcm. after discuss with @xlc, the new design of fee charge will change to using a fee pool along with fix rate to not only support transaction payment, but also support xcm buy execution.
● the fix rate will be update once 20 blocks(need benchmark?)
● each token has its fee pool, this pool record the total balance of each token

the new workflow of transaction payment charge fee:
● if user has enough native token(e.x. KAR) for charge fee, then just use it's own balance. this part unchanged.
● else the user has enought foreign asset(e.x. KSM), but not enought of native token
  ○ calcutate fee by WeightToFee (x)
  ○ use fix rate to calculate the amount of foreign asset. 
      ■ e.x. the fix rate of KAR to KSM is m. then we get m*fee_amount_of_KAR amount of KSM(y)
      ■ the initial fix rate is determined by dex module or we could setup it an initial fixed value
[...]

PR #1687, opened 2021-12-08
rework fee payment

zqhxuyuan:
closes: https://github.com/AcalaNetwork/Acala/issues/1233

- [x] - init KAR pool from treasury
    - [x] each foreign asset has one sub account of TreasuryPallet
    - [x] on_runtime_upgrade, each sub account get initial_boot_strap_balance from treasury account
- [x] - how to trigger swap if needed
    - [x] ~~trigger by block number~~
    - [x] trigger by balance: each asset has native asset balance threshold,default to initial_boot_strap_balance div 5.
- [x] - change `ensure_can_charge_fee` to use fix rate
    - [x] - ~~use an switch to change to chage fee from pool~~ 
    - [x] - directly replace the swap_from_dex
- [x] add `WeightTrader` depend on the rate in the storage. 
    - [x] ~~move to other package(i.e. runtime?), so that no depend on xcm related module.~~
    - [x] more integration test when dynamic rate failed, then use fixed rate instead.
- [x] - testcase
    - [x] `transaction payment` 
    - [x] `PeriodUpdatedRateOfFungible`
    - [x] fix old testcase. need more fix when https://github.com/AcalaNetwork/Acala/pull/1682 is merged.

design work flow:
image
 
review point:
- the initial_bootstrap_balance value is suitable? and also the trigger_threshold=initial_bootstrap_balance/5?
- current each sub account has same initial_bootstrap_balance, do they need different setting?

discussion:
- do we need ED limitation for the treasury fee pool account? if trigger swap from dex, the `min_target_amount` of `DEX::swap_with_exact_supply` then should be ED + account.
- what if swap from dex failed, should we get native asset from treasury account like the initial bootstrap phase? current, we only have one-shot initial phase. after that if treasury account's native asset not enough, it swap out native asset with its foreign asset(which gets from user account), then the treasury account's native asset goes up.


modules/transaction-payment/src/lib.rs
```diff
@@ -272,6 +279,21 @@ pub mod module {
...
+		/// Initial native asset balance transfer from treasury to pool.
+		#[pallet::constant]
+		type InitialBootstrapBalanceForFeePool: Get<PalletBalanceOf<Self>>;
+
+		/// The rate of foreign_asset_per_second to native_asset_per_second.
+		#[pallet::constant]
+		type AssetFixRateAccountIds: Get<Vec<AssetFixRateAccountId<Self::AccountId>>>;
+
+		/// The root operator of admin operation(i.e. change trigger threshold).
+		type TreasuryOrigin: EnsureOrigin<Self::Origin, Success = Self::AccountId>;
```
xlc, line 296:
```suggestion
		type UpdateOrigin: EnsureOrigin<Self::Origin, Success = Self::AccountId>;
```

modules/transaction-payment/src/lib.rs
```diff
@@ -308,13 +334,43 @@ pub mod module {
...
+
+	/// The treasury fee pool sub account of foregign asset.
+	#[pallet::storage]
+	#[pallet::getter(fn treasury_accounts)]
+	pub type TreasuryAccounts<T: Config> = StorageMap<_, Twox64Concat, CurrencyId, T::AccountId, ValueQuery>;
+
+	/// Each asset has threshold to trigger swap, default to `InitialBootstrapBalanceForFeePool`/5.
+	#[pallet::storage]
+	#[pallet::getter(fn swap_threshold)]
+	pub type SwapBalanceThreshold<T: Config> = StorageMap<_, Twox64Concat, CurrencyId, u128, ValueQuery>;
```
xlc, line 351:
```suggestion
	pub type SwapBalanceThreshold<T: Config> = StorageMap<_, Twox64Concat, CurrencyId, Balance, ValueQuery>;
```

Why use u128?

modules/transaction-payment/src/lib.rs
```diff
@@ -308,13 +334,43 @@ pub mod module {
...
+	/// The ratio of foreign_asset_per_second to native_asset_per_second.
+	/// the reverse ratio means the value of foreign asset to native asset.
+	#[pallet::storage]
+	#[pallet::getter(fn charge_fee_pool)]
+	pub type FeeRateOfToken<T: Config> = StorageMap<_, Twox64Concat, CurrencyId, Ratio, OptionQuery>;
+
+	/// The treasury fee pool sub account of foregign asset.
+	#[pallet::storage]
+	#[pallet::getter(fn treasury_accounts)]
+	pub type TreasuryAccounts<T: Config> = StorageMap<_, Twox64Concat, CurrencyId, T::AccountId, ValueQuery>;
```
xlc, line 346: no need to store those. can just create sub account on the fly

https://github.com/paritytech/substrate/blob/a64f702d20fd0e979c69d11b60e0d6d6b81d6c6d/frame/bounties/src/lib.rs#L771-L775

modules/transaction-payment/src/lib.rs
```diff
@@ -391,6 +447,20 @@ pub mod module {
...
+
+		/// set trigger balance of native asset
+		#[pallet::weight(<T as Config>::WeightInfo::set_trigger_threshold())]
+		pub fn set_trigger_threshold(
+			origin: OriginFor<T>,
+			currency_id: CurrencyId,
+			threshold: PalletBalanceOf<T>,
+		) -> DispatchResult {
+			T::TreasuryOrigin::ensure_origin(origin)?;
+			if threshold < T::InitialBootstrapBalanceForFeePool::get() {
```
xlc, line 459: why compare this?  
should also have event for success update

modules/transaction-payment/src/lib.rs
```diff
@@ -272,6 +279,21 @@ pub mod module {
...
 		/// Weight information for the extrinsics in this module.
 		type WeightInfo: WeightInfo;
+
+		/// Treasury account used to initial balance bootstrap.
+		#[pallet::constant]
+		type TreasuryAccount: Get<Self::AccountId>;
+
+		/// Initial native asset balance transfer from treasury to pool.
+		#[pallet::constant]
+		type InitialBootstrapBalanceForFeePool: Get<PalletBalanceOf<Self>>;
```
xlc, line 289: this should just be `PoolSize`?

xlc (commented):
you can change to `type Currency: Currency<Self::AccountId, Balance = Balance>` to avoid u128 and conversion

zqhxuyuan:
> 

Great! I think we could also replace all `PalletBalanceOf<T>` to `Balance`.

modules/transaction-payment/src/lib.rs
```diff
@@ -293,6 +312,19 @@ pub mod module {
...
+		AssetNotInFeePool,
+	}
+
+	#[pallet::event]
+	#[pallet::generate_deposit(pub(crate) fn deposit_event)]
+	pub enum Event<T: Config> {
+		TriggerThresholdUpdated(Balance),
+		TriggerSwap(Balance, Balance, CurrencyId),
+		SwapFromDex(Balance, Balance, Ratio),
+		ChargeFee(Balance, Balance, Ratio),
```
shaunxw, line 327: 1. Add event dostrings like

```rust
/// Trigger threshold updated. \[Balance, Balance\]
TriggerThresholdUpdated(Balance),
```

2. Rename events. In general, events naming convention is like:
- DO: `FeeCharged`
- DO NOT: `ChargeFee`

modules/transaction-payment/src/lib.rs
```diff
@@ -302,19 +334,38 @@ pub mod module {
 	#[pallet::getter(fn next_fee_multiplier)]
 	pub type NextFeeMultiplier<T: Config> = StorageValue<_, Multiplier, ValueQuery, DefaultFeeMultiplier>;
 
+	#[pallet::storage]
+	#[pallet::getter(fn initial_balance)]
```
shaunxw, line 338:
```suggestion
	#[pallet::getter(fn pool_size)]
```

modules/transaction-payment/src/lib.rs
```diff
@@ -302,19 +334,38 @@ pub mod module {
...
 	#[pallet::storage]
 	#[pallet::getter(fn alternative_fee_swap_path)]
 	pub type AlternativeFeeSwapPath<T: Config> =
 		StorageMap<_, Twox64Concat, T::AccountId, BoundedVec<CurrencyId, T::TradingPathLimit>, OptionQuery>;
 
+	/// The ratio of foreign_asset_per_second to native_asset_per_second.
+	/// the reverse ratio means the value of foreign asset to native asset.
+	#[pallet::storage]
+	#[pallet::getter(fn fee_rate)]
+	pub type FeeRateOfToken<T: Config> = StorageMap<_, Twox64Concat, CurrencyId, Ratio, OptionQuery>;
```
shaunxw, line 351: Please make sure the storage type name and getter name are consistent.

```suggestion
	#[pallet::getter(fn token_fixed_rate)]
	pub type TokenFixedRate<T: Config> = StorageMap<_, Twox64Concat, CurrencyId, Ratio, OptionQuery>;
```

modules/transaction-payment/src/lib.rs
```diff
@@ -302,19 +334,38 @@ pub mod module {
...
+
+	/// Each asset has threshold to trigger swap, default to `PoolSize`/5.
+	#[pallet::storage]
+	#[pallet::getter(fn trigger_threshold)]
+	pub type SwapBalanceThreshold<T: Config> = StorageMap<_, Twox64Concat, CurrencyId, Balance, ValueQuery>;
```
shaunxw, line 356: Same here.

modules/transaction-payment/src/lib.rs
```diff
@@ -578,59 +642,199 @@ where
...
+		// make sure add extra gap to keep alive after swap.
+		let amount = fee.saturating_add(native_existential_deposit.saturating_sub(total_native));
+		Self::swap_native_asset(who, amount);
+	}
+
+	/// Iterate order list, if can swap out enough native asset amount with foreign asset, then stop
+	fn swap_native_asset(who: &T::AccountId, amount: Balance) {
+		let native_currency_id = T::NativeCurrencyId::get();
+		for trading_path in Self::get_trading_path(who) {
+			match trading_path.last() {
```
shaunxw, line 663: This `match` only has one valid branch, use `if let` instead.

modules/transaction-payment/src/lib.rs
```diff
@@ -578,59 +642,199 @@ where
...
+	fn swap_from_treasury(
+		who: &T::AccountId,
+		amount: Balance,
+		supply_currency_id: CurrencyId,
+	) -> sp_std::result::Result<(), DispatchError> {
+		let treasury_account = Self::treasury_sub_account_id(supply_currency_id);
+		let treasury_supply_balance = T::MultiCurrency::free_balance(supply_currency_id, &treasury_account);
+		let mut rate = FeeRateOfToken::<T>::get(supply_currency_id)
+			.ok_or(Error::<T>::AssetNotInFeePool)
+			.unwrap();
```
shaunxw, line 686:
```suggestion
			.unwrap();
```

`unwrap` is not allowed in runtime, return error or use `expect` with proofs.

modules/transaction-payment/src/lib.rs
```diff
@@ -578,59 +642,199 @@ where
...
+		// if treasury account has not enough native asset, trigger swap
+		let treasury_balance = T::Currency::free_balance(&treasury_account);
+		if treasury_balance < swap_balance_threshold {
+			let trading_path = Self::get_trading_path_by_currency(&treasury_account, supply_currency_id);
+			if let Some(trading_path) = trading_path {
+				Self::deposit_event(Event::TriggerSwap(
+					treasury_balance,
+					swap_balance_threshold,
+					supply_currency_id,
+				));
```
shaunxw, line 706: `TriggerSwap` here doesn't make sense to me, it's more like a debugging log.

modules/transaction-payment/src/lib.rs
```diff
@@ -578,59 +642,199 @@ where
...
+			vec![vec![trading_path.into_inner()], default_fee_swap_path_list].concat()
+		} else {
+			default_fee_swap_path_list
+		}
+	}
+
+	fn get_trading_path_by_currency(who: &T::AccountId, supply_currency_id: CurrencyId) -> Option<Vec<CurrencyId>> {
+		let fee_swap_path_list: Vec<Vec<CurrencyId>> = Self::get_trading_path(who);
+		for trading_path in fee_swap_path_list {
+			match trading_path.first() {
```
shaunxw, line 740: Replace with `if let`

modules/transaction-payment/src/lib.rs
```diff
@@ -578,59 +642,199 @@ where
...
+				Some(currency) if *currency == supply_currency_id => {
+					return Some(trading_path);
+				}
+				_ => {}
+			}
+		}
+		None
+	}
+
+	pub fn treasury_sub_account_id(id: CurrencyId) -> T::AccountId {
```
shaunxw, line 750: Add docstring to pub `fn`.

modules/transaction-payment/src/lib.rs
```diff
@@ -578,59 +642,199 @@ where
...
+		T::TreasuryPalletId::get().into_sub_account(id)
+	}
+
+	pub fn update_storage(balance: Balance, currency_id: CurrencyId, rate: Ratio) {
```
shaunxw, line 754: Same here.

modules/transaction-payment/src/lib.rs
```diff
@@ -578,59 +642,199 @@ where
...
+		let from = T::TreasuryAccount::get();
+		let account = Self::treasury_sub_account_id(currency_id);
+		let _ = T::Currency::transfer(&from, &account, balance, ExistenceRequirement::KeepAlive);
+		SwapBalanceThreshold::<T>::insert(currency_id, balance / 5);
+		FeeRateOfToken::<T>::insert(currency_id, rate);
+		PoolSize::<T>::set(balance);
+	}
+}
+
+pub struct PeriodUpdatedRateOfFungible<T, C, K: Get<u128>, R: TakeRevenue> {
```
shaunxw, line 764: Need docstring.

modules/transaction-payment/src/lib.rs
zqhxuyuan: ok, sounds too much events.

modules/transaction-payment/src/lib.rs
xlc: Actually we should just use named fields for new events. i.e. `TriggerThresholdUpdated { new_value: Balance }`

zqhxuyuan: As the `transfer` action already produce events. do we need `FeeCharged` event that show the supply asset balance transfer to and native asset got from treasury account, also the rate currently used?

runtime/karura/src/lib.rs
```diff
@@ -165,6 +165,7 @@ parameter_types! {
 	pub const NftPalletId: PalletId = PalletId(*b"aca/aNFT");
 	// Vault all unrleased native token.
 	pub UnreleasedNativeVaultAccountId: AccountId = PalletId(*b"aca/urls").into_account();
+	pub const TreasuryFeePoolPalletId: PalletId = PalletId(*b"aca/fees");
```
zjb0807, line 168: Add `TreasuryFeePoolPalletId` into `get_all_module_accounts`

modules/transaction-payment/src/lib.rs
```diff
@@ -611,60 +673,198 @@ where
...
+
+		// user's supply asset should have enough balance based on old fix rate.
+		ensure!(
+			T::MultiCurrency::free_balance(supply_currency_id, who) >= rate.saturating_mul_int(amount),
+			Error::<T>::SupplyBalanceTooLow
+		);
+
+		// if treasury account has not enough native asset, trigger swap
+		let treasury_balance = T::Currency::free_balance(&treasury_account);
+		if treasury_balance < SwapBalanceThreshold::<T>::get(supply_currency_id) {
```
zjb0807, line 725: `treasury_supply_balance` maybe is 0.
It seems we don’t need `SwapBalanceThreshold`. We need to swap all of the tokens(or the largest value) in the `on_initialize`.
@xlc How do you think about this?

modules/transaction-payment/src/lib.rs
```diff
@@ -611,60 +673,198 @@ where
...
 	}
+
+	/// swap from treasury fee pool.
+	#[transactional]
+	fn swap_from_treasury(
+		who: &T::AccountId,
+		amount: Balance,
+		supply_currency_id: CurrencyId,
+	) -> sp_std::result::Result<(), DispatchError> {
+		let treasury_account = Self::treasury_sub_account_id(supply_currency_id);
```
zjb0807, line 713: use `T::TreasuryPalletId::get()` to avoid ED

runtime/karura/src/lib.rs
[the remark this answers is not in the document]
zqhxuyuan: The sub account derivate by `TreasuryFeePoolPalletId` instead of `TreasuryFeePoolPalletId` is used in the charge fee pool, does it make sense here? or we should add all sub account to `get_all_module_accounts`?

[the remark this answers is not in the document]
zjb0807: I think just `TreasuryFeePoolPalletId` is enough, don't need sub account.

modules/transaction-payment/src/lib.rs
```diff
@@ -611,60 +673,198 @@ where
...
+		let account = Self::treasury_sub_account_id(currency_id);
+		let _ = T::Currency::transfer(
+			&T::TreasuryAccount::get(),
+			&account,
+			balance,
+			ExistenceRequirement::KeepAlive,
+		);
+		SwapBalanceThreshold::<T>::insert(currency_id, balance / 5);
+		TokenFixedRate::<T>::insert(currency_id, rate);
+		PoolSize::<T>::set(balance);
```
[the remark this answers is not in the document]
zqhxuyuan: if Acala treasury balance is 0, the `transfer` action will failed, and `TokenFixedRate` is none, when `swap_from_treasury_or_dex` it'll throw err causing no side effect.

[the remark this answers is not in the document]
zjb0807: The result of `transfer` is ignored.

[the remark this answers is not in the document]
zqhxuyuan: Ah, I should use `transfer()?` instead, does that make sense?

modules/transaction-payment/src/lib.rs
```diff
@@ -611,59 +673,206 @@ where
...
+	/// swap treasury account's native asset with user's foreign asset from treasury fee pool when
+	/// treasury account's native asset balance is gt threshold. or-else, swap treasury account's
+	/// native asset with treasury account's foreign asset from dex swap liquid pool.
+	#[transactional]
+	fn swap_from_treasury_or_dex(
+		who: &T::AccountId,
+		amount: Balance,
+		supply_currency_id: CurrencyId,
+	) -> sp_std::result::Result<(), DispatchError> {
+		match TokenFixedRate::<T>::get(supply_currency_id) {
```
xlc, line 716: replace the match with

```rs
let mut rate = TokenFixedRate::<T>::get(supply_currency_id).ok_or(Error::<T>::InvalidRate)?;
```

Also `DispatchError::CannotLookup` means account id look up error, which is not the right error to be used here.

modules/transaction-payment/src/lib.rs
```diff
@@ -611,59 +673,206 @@ where
...
+		None
+	}
+
+	/// The sub account derivated by `TreasuryPalletId`.
+	fn treasury_sub_account_id(id: CurrencyId) -> T::AccountId {
+		T::TreasuryPalletId::get().into_sub_account(id)
+	}
+
+	/// Setting some initial value when on_runtime_upgrade.
+	pub fn on_runtime_upgrade(balance: Balance, currency_id: CurrencyId, rate: Ratio) -> DispatchResult {
```
xlc, line 788:
```suggestion
	/// Initiate a treasury swap pool. Usually used in on_runtime_upgrade.
	pub fn initialize_pool(balance: Balance, currency_id: CurrencyId, rate: Ratio) -> DispatchResult {
```

modules/transaction-payment/src/lib.rs
```diff
@@ -611,59 +673,206 @@ where
...
+
+	/// swap treasury account's native asset with user's foreign asset from treasury fee pool when
+	/// treasury account's native asset balance is gt threshold. or-else, swap treasury account's
+	/// native asset with treasury account's foreign asset from dex swap liquid pool.
+	#[transactional]
+	fn swap_from_treasury_or_dex(
+		who: &T::AccountId,
+		amount: Balance,
+		supply_currency_id: CurrencyId,
+	) -> sp_std::result::Result<(), DispatchError> {
```
xlc, line 715:
```suggestion
	) -> DispatchResult {
```

modules/transaction-payment/src/lib.rs
```diff
@@ -611,59 +673,206 @@ where
...
+							rate = Ratio::saturating_from_rational(new_native_balance, PoolSize::<T>::get());
+							TokenFixedRate::<T>::insert(supply_currency_id, rate);
 						}
 					}
-					_ => {}
+				}
+
+				// use fix rate to calculate the amount of supply asset that equal to native asset.
+				// exchange user's supply asset and treasury's native asset.
+				let token_amount: u128 = rate.saturating_mul_int(amount);
```
xlc, line 751:
```suggestion
				let token_amount = rate.saturating_mul_int(amount);
```

modules/transaction-payment/src/lib.rs
```diff
@@ -611,59 +673,206 @@ where
...
+	) -> sp_std::result::Result<(), DispatchError> {
+		match TokenFixedRate::<T>::get(supply_currency_id) {
+			None => {
+				// if the token rate is none(i.e. disable by council or on purpose), return err.
+				Err(DispatchError::CannotLookup)
+			}
+			Some(mut rate) => {
+				// user's supply asset should have enough balance based on old fix rate.
+				ensure!(
+					T::MultiCurrency::free_balance(supply_currency_id, who) >= rate.saturating_mul_int(amount),
```
xlc, line 724: no need to check free balance, just do the transfer here, which will check the free balance.

modules/transaction-payment/src/lib.rs
```diff
@@ -611,59 +673,206 @@ where
...
+							T::MultiCurrency::free_balance(supply_currency_id, &treasury_account);
+						if let Ok(swap_native_balance) =
+							T::DEX::swap_with_exact_supply(&treasury_account, &trading_path, treasury_supply_balance, 0)
 						{
-							// successfully swap, break iteration
-							break;
+							// calculate and update new rate of supply asset relative to native asset
+							let new_native_balance =
+								rate.saturating_mul_int(swap_native_balance.saturating_add(treasury_balance));
+							rate = Ratio::saturating_from_rational(new_native_balance, PoolSize::<T>::get());
```
xlc, line 743: I think it is ok to make this transaction to use the old rate. makes frontend calculation easier.

This formula means we can't change pool size otherwise the rate will be wrong.

modules/transaction-payment/src/lib.rs
```diff
@@ -611,59 +673,206 @@ where
...
+		let account = Self::treasury_sub_account_id(currency_id);
+		T::Currency::transfer(
+			&T::TreasuryAccount::get(),
+			&account,
+			balance,
+			ExistenceRequirement::KeepAlive,
+		)?;
+		SwapBalanceThreshold::<T>::insert(currency_id, balance / 5);
+		TokenFixedRate::<T>::insert(currency_id, rate);
+		PoolSize::<T>::set(balance);
```
xlc, line 798: this needs to be different per currency id? otherwise it shouldn't be updated here

primitives/src/lib.rs
```diff
@@ -99,6 +99,12 @@ pub type Block = generic::Block<Header, UncheckedExtrinsic>;
 /// Block ID.
 pub type BlockId = generic::BlockId<Block>;
 
+/// The fixed rate of token.
+#[derive(Encode, Decode, Eq, PartialEq, Copy, Clone, RuntimeDebug, PartialOrd, Ord, TypeInfo)]
+#[cfg_attr(feature = "std", derive(Serialize, Deserialize))]
+#[cfg_attr(feature = "std", serde(rename_all = "camelCase"))]
+pub struct TokenFixedRate(pub CurrencyId, pub FixedU128);
```
xlc, line 106: this doesn't looks like useful

xlc (commented):
Another thing to keep in mind is that people can always send money to system accounts. So need to make sure nothing goes bad if that happens.

runtime/karura/src/lib.rs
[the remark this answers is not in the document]
zqhxuyuan: use sub account here, and add all sub account to whitelist.

modules/transaction-payment/src/lib.rs
zqhxuyuan: add some doc. we normaly set threshold >= ED to prevent exhausted treasury balance. but there maybe cases that treasury supply balance is <= ED(but it's not 0, because the treasury sub account is in our whitelist), it that happens, the swap may failed.

modules/transaction-payment/src/lib.rs
```diff
@@ -611,60 +675,203 @@ where
...
+		// fee in this transaction. so for fee charge as successful as possible, we reserve one block
+		// fee, that's SwapBalanceThreshold can set to ED + one block fee.
+		let treasury_balance = T::Currency::free_balance(&treasury_account);
+		if treasury_balance < SwapBalanceThreshold::<T>::get(supply_currency_id) {
+			let trading_path = Self::get_trading_path_by_currency(&treasury_account, supply_currency_id);
+			if let Some(trading_path) = trading_path {
+				// using all supply asset to swap out native asset, no need to consider ED limitation here.
+				let treasury_supply_balance = T::MultiCurrency::free_balance(supply_currency_id, &treasury_account);
+				if let Ok(swap_native_balance) =
+					T::DEX::swap_with_exact_supply(&treasury_account, &trading_path, treasury_supply_balance, 0)
```
zqhxuyuan, line 733: although we use whitelist treasury sub account to store foreign asset from user, but when swap happen, if the foreign asset balance of treasury sub account still less than foreign asset's ED, the swap will failed and finally cause native balance really exhausted then charge fee pool will not works any more. we may get balance from treasury account like initial_pool phase when swap failed in this scene. I'll calcuate some data, and if need, add this logic later.

xlc: the sub accounts are not whitelisted and I don't really want to add them to the ED whitelist
so I think the easy solution is ensure there is always ED in the sub accounts.
this means for all the transfer, query balance etc, it need to ensure there is always ED left
and in the initialization code, transfer ED amount from treasury account to sub account

xlc: we really need to make sure the swap don't fail, and add a `else { debug_assert!(false, "Swap tx fee pool should not fail"); } ` to check it

runtime/acala/src/lib.rs
```diff
@@ -178,7 +179,10 @@ pub fn get_all_module_accounts() -> Vec<AccountId> {
 		TreasuryPalletId::get().into_account(),
 		TreasuryReservePalletId::get().into_account(),
 		UnreleasedNativeVaultAccountId::get(),
-	]
+		TreasuryFeePoolPalletId::get().into_account(),
+	];
+	whitelist.append(&mut get_all_sub_account_of_fee_pool());
+	whitelist
```
zqhxuyuan, line 185: rollback to disable treasury aub account. make sure the sub account has ED of foreign asset in initial_pool phase.

runtime/karura/src/lib.rs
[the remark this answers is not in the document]
zqhxuyuan: don't add to whitelist, just make sure the account has ED balance limitation when initial_pool.

runtime/acala/src/lib.rs
```diff
@@ -164,6 +164,8 @@ parameter_types! {
 	pub const NftPalletId: PalletId = PalletId(*b"aca/aNFT");
 	// Vault all unrleased native token.
 	pub UnreleasedNativeVaultAccountId: AccountId = PalletId(*b"aca/urls").into_account();
+	// This Pallet is only used to payment fee pool, it's not added to whitelist by design.
```
xlc, line 167:
```suggestion
	// This Pallet is only used to payment fee pool, it's not added to whitelist by design because transaction payment pallet will ensure the accounts always have enough ED.
```

runtime/acala/src/lib.rs
```diff
@@ -1862,9 +1878,26 @@ pub type Executive = frame_executive::Executive<
...
+
+pub struct TransactionPaymentUpgrade;
+impl frame_support::traits::OnRuntimeUpgrade for TransactionPaymentUpgrade {
+	fn on_runtime_upgrade() -> Weight {
+		for asset in TokenFixedRates::get() {
+			let _ = <module_transaction_payment::Pallet<Runtime>>::initialize_pool(asset.0, asset.1, asset.2);
+		}
+		0
+	}
+}
```
ferrell-code, line 1899: we should probably use a non-zero value for the Weight. We could benchmark it or just use the `max_weight` of a block

zqhxuyuan: Oh I get it, when runtime upgrade happens, we dont't expect other extrinsics in this block

ferrell-code: `on_runtime_upgrade` uses computation time (weight) just like anything else so we don't want to go over the .5 seconds of allocated time for block production. Using max block weight is just a shortcut so we don't have to benchmark anything, because no other normal extrinsic will be added to the block and therefore the block will be produced in time. In practice the function `initialize_pool` is computationally light enough that using zero for its weight would be fine as we have some leeway.

modules/transaction-payment/src/lib.rs
```diff
@@ -305,6 +314,18 @@ pub mod module {
 
 		/// Weight information for the extrinsics in this module.
 		type WeightInfo: WeightInfo;
+
+		/// Treasury PalletId used to derivate sub account.
+		#[pallet::constant]
+		type TreasuryPalletId: Get<PalletId>;
```
xlc, line 320:
```suggestion
		type PalletId: Get<PalletId>;
```

This is pallet ID of this pallet, not the treasury pallet. We want to avoid any possible confusion.

modules/transaction-payment/src/lib.rs
```diff
@@ -326,6 +347,28 @@ pub mod module {
...
+		/// Dex swap pool is not available now
+		DexNotAvailable,
+		/// Charge fee pool is already exist
+		ChargeFeePoolAlreadyExisted,
+	}
+
+	#[pallet::event]
+	#[pallet::generate_deposit(pub(crate) fn deposit_event)]
+	pub enum Event<T: Config> {
+		/// The threshold balance that trigger swap from dex was updated. \[Balance\]
```
xlc, line 363:
```suggestion
		/// The threshold balance that trigger swap from dex was updated.
```

modules/transaction-payment/src/lib.rs
```diff
@@ -326,6 +347,28 @@ pub mod module {
...
+		SwapBalanceThresholdUpdated { new_threshold_value: Balance },
+		/// The charge fee pool is enabled \[CurrencyId, Ratio, Balance, Balance\]
```
xlc, line 365:
```suggestion
		/// The charge fee pool is enabled
```

modules/transaction-payment/src/lib.rs
```diff
@@ -341,6 +384,23 @@ pub mod module {
 	pub type AlternativeFeeSwapPath<T: Config> =
 		StorageMap<_, Twox64Concat, T::AccountId, BoundedVec<CurrencyId, T::TradingPathLimit>, OptionQuery>;
 
+	/// The bootstrap balance transfer from `TreasuryAccount` to sub account of `TreasuryPalletId`.
```
xlc, line 387:
```suggestion
	/// The size of fee pool in native token. During bootstrap this amount of native token will be transferred from `TreasuryAccount` to sub account of `PalletId`.
```

modules/transaction-payment/src/lib.rs
```diff
@@ -341,6 +384,23 @@ pub mod module {
...
+	#[pallet::storage]
+	#[pallet::getter(fn pool_size)]
+	pub type PoolSize<T: Config> = StorageMap<_, Twox64Concat, CurrencyId, Balance, ValueQuery>;
+
+	/// The ratio of `foreign_asset_per_second` relative to `native_asset_per_second`.
+	/// the reverse ratio means the value of foreign asset to native asset.
+	/// the rate is updated when executing swap from dex.
```
xlc, line 394:
```suggestion
	/// The exchange rate between the given currency and native token.
	/// This value is updated when upon swap from dex.
```

modules/transaction-payment/src/lib.rs
```diff
@@ -341,6 +384,23 @@ pub mod module {
...
+	#[pallet::storage]
+	#[pallet::getter(fn token_fixed_rate)]
+	pub type TokenFixedRate<T: Config> = StorageMap<_, Twox64Concat, CurrencyId, Ratio, OptionQuery>;
```
xlc, line 397:
```suggestion
	#[pallet::getter(fn token_exchange_rate)]
	pub type TokenExchangeRate<T: Config> = StorageMap<_, Twox64Concat, CurrencyId, Ratio, OptionQuery>;
```

modules/transaction-payment/src/lib.rs
```diff
@@ -424,6 +484,37 @@ pub mod module {
...
+		}
+
+		/// enable and initialize charge fee pool.
+		#[pallet::weight(<T as Config>::WeightInfo::enable_initial_pool())]
+		pub fn enable_initial_pool(
+			origin: OriginFor<T>,
+			currency_id: CurrencyId,
+			balance: Balance,
+			token_per_second: Balance,
+			native_per_second: Balance,
```
xlc, line 511:
```suggestion
			exchange_rate: Ratio,
```

modules/transaction-payment/src/lib.rs
```diff
@@ -424,6 +484,37 @@ pub mod module {
...
+			});
+			Ok(())
+		}
+
+		/// enable and initialize charge fee pool.
+		#[pallet::weight(<T as Config>::WeightInfo::enable_initial_pool())]
+		pub fn enable_initial_pool(
+			origin: OriginFor<T>,
+			currency_id: CurrencyId,
+			balance: Balance,
```
xlc, line 509:
```suggestion
			pool_size: Balance,
```

modules/transaction-payment/src/lib.rs
```diff
@@ -611,62 +702,240 @@ where
...
+	}
+
+	/// Get trading path by user.
+	fn get_trading_path(who: &T::AccountId) -> Vec<Vec<CurrencyId>> {
+		let default_fee_swap_path_list = T::DefaultFeeSwapPathList::get();
+		if let Some(trading_path) = AlternativeFeeSwapPath::<T>::get(who) {
+			vec![vec![trading_path.into_inner()], default_fee_swap_path_list].concat()
+		} else {
+			default_fee_swap_path_list
+		}
```
xlc, line 792:
```suggestion
		let mut default_fee_swap_path_list = T::DefaultFeeSwapPathList::get();
		if let Some(trading_path) = AlternativeFeeSwapPath::<T>::get(who) {
			default_fee_swap_path_list.insert(0, trading_path.into_inner())
		}
		default_fee_swap_path_list
```

modules/transaction-payment/src/lib.rs
```diff
@@ -611,62 +702,240 @@ where
...
+	/// action, if charge fee pool already existed, this will throw `ChargeFeePoolAlreadyExisted`
+	/// error. current we don't support change pool size as it's used as a factor of rate
+	/// calculation. normally the charge fee pool wouldn't exhausted. although user charge fee will
+	/// decrease the balance of charge fee pool, but when charge fee pool lt threshold, swap from
+	/// dex action will increase the balance of charge fee pool.
+	#[transactional]
+	pub fn initialize_pool(
+		currency_id: CurrencyId,
+		rate: Ratio,
+		balance: Balance,
```
xlc, line 823:
```suggestion
		pool_size: Balance,
```

modules/transaction-payment/src/lib.rs
```diff
@@ -611,62 +702,240 @@ where
...
+	/// Initiate a treasury swap pool. Usually used in on_runtime_upgrade. This is a one-shot
+	/// action, if charge fee pool already existed, this will throw `ChargeFeePoolAlreadyExisted`
+	/// error. current we don't support change pool size as it's used as a factor of rate
+	/// calculation. normally the charge fee pool wouldn't exhausted. although user charge fee will
+	/// decrease the balance of charge fee pool, but when charge fee pool lt threshold, swap from
+	/// dex action will increase the balance of charge fee pool.
+	#[transactional]
+	pub fn initialize_pool(
+		currency_id: CurrencyId,
+		rate: Ratio,
```
xlc, line 822: no need to pass rate, just use the real rate from dex

runtime/karura/src/lib.rs
```diff
@@ -1412,6 +1424,32 @@ pub type XcmOriginToCallOrigin = (
...
+pub struct ToTreasury;
+impl TakeRevenue for ToTreasury {
+	fn take_revenue(revenue: MultiAsset) {
+		if let MultiAsset {
+			id: Concrete(location),
+			fun: Fungible(amount),
+		} = revenue
+		{
+			if let Some(currency_id) = CurrencyIdConvert::convert(location) {
+				// ensure KaruraTreasuryAccount have ed for all of the cross-chain asset.
```
xlc, line 1445:
```suggestion
				// KaruraTreasuryAccount should not have ED requirement.
```

modules/transaction-payment/src/lib.rs
```diff
@@ -611,62 +702,240 @@ where
...
+				threshold,
+			});
+		}
+		Ok(())
+	}
+}
+
+/// `WeightTrader` implementation used for `Trader`, the `rate` is read from storage,
+/// and `token_per_second` is calculated by `rate` * `native_asset_per_second`.
+pub struct PeriodUpdatedRateOfFungible<T, C, K: Get<u128>, R: TakeRevenue> {
```
xlc, line 867: the name is a bit confusing. Just make it `TransactionFeePoolTrader`

xlc (changes requested):
the name treasury account is over used and confusing. we should only use name treasury account to the real treasury, and call it fee pool account for the pallet sub account.

runtime/karura/src/lib.rs
zqhxuyuan: just some code snippet reorder, but I think the ED limitation only aim to native asset, not to cross-chain asset, as for cross chain asset case, the `KaruraTreasuryAccount` is in whitelist.

xlc (approved):

wangjj9219 (approved):
LGTM

wangjj9219:
@StrawberryFlavor need test it on testnet

zqhxuyuan:
A few note about test, ping me if you have question:
- there'll be an transfer from treasury account to PalletId sub account to initialize charge fee pool
- when user have enough foreign tokens, but don't have enough native asset, exchange fee directly from charge fee pool
- when the charge fee pool is lt some threshold(which is gt ED and some reserve), trigger swap from dex
- each foreign token has an sub-account. that's each token has an charge fee pool
- after swap from dex, the charge fee pool get charge of native asset, and left only ED of foreign asset. 
- then user can continue exchange fee from charge fee pool again.

PR #1692, opened 2021-12-13
update bootnodes and verify script

xlc:


zjb0807 (approved):
It may fail at first, because there are fewer blocks, and it will be synchronized soon.

PR #1694, opened 2021-12-13
handle evm-bridge number overflow

ermalkaleci:


modules/evm-bridge/src/lib.rs
```diff
@@ -74,6 +74,8 @@ pub mod module {
 		ExecutionError,
 		/// Invalid return value
 		InvalidReturnValue,
+		/// Number overflow
+		NumberOverflow,
 	}
```
ferrell-code, line 79: There is a built in `ArithmeticError::Overflow` in `DispatchError`, in `sp_runtime`

https://github.com/paritytech/substrate/blob/3cdb30e1ecbafe8a866317d4550c921b4d686869/primitives/runtime/src/lib.rs#L589

ermalkaleci: @ferrell-code yeah but I wan't sure because we're not doing any arithmetic but just downgrading to u128 so thought this might make more sense. But I see know we're already using it. Updating...

modules/support/src/lib.rs
```diff
@@ -327,7 +327,13 @@ pub trait PrecompileCallerFilter {
 
 /// An abstraction of EVM for EVMBridge
 pub trait EVM<AccountId> {
-	type Balance: AtLeast32BitUnsigned + Copy + MaybeSerializeDeserialize + Default;
+	type Balance: AtLeast32BitUnsigned
+		+ TryFrom<u32>
+		+ TryFrom<u64>
+		+ TryFrom<u128>
```
xlc, line 333: `AtLeast32BitUnsigned` should already include those

ermalkaleci: @xlc yes you're right

xlc (approved):

PR #1707, opened 2021-12-16
handle nonce for eth tx

### LINKED ISSUE #1702: Support evm.nonce  for CheckNonce
Support  `evm.nonce` for `eth_call` to submit extrinsics.

ermalkaleci:
closes #1702


runtime/mandala/src/lib.rs
```diff
@@ -1959,21 +1959,37 @@ impl cumulus_pallet_aura_ext::Config for Runtime {}
...
 				gas_limit,
 				storage_limit,
+				nonce,
 				valid_until,
 			}) => {
 				if System::block_number() > valid_until {
 					return Err(InvalidTransaction::Stale);
 				}
 
+				let address = EvmAddressMapping::<Runtime>::get_default_evm_address(&who);
```
xlc, line 1982: should be using `get_evm_address` and fallback to default one if none

runtime/mandala/src/lib.rs
```diff
@@ -1959,21 +1959,37 @@ impl cumulus_pallet_aura_ext::Config for Runtime {}
...
+		(who, call, extra): (AccountId, Call, SignedExtra),
+	) -> Result<EthereumTransactionMessage, InvalidTransaction> {
 		match call {
 			Call::EVM(module_evm::Call::eth_call {
 				action,
 				input,
 				value,
 				gas_limit,
 				storage_limit,
+				nonce,
```
xlc, line 1975: why you added a new nonce instead of using the value from `CheckNonce`?

xlc (changes requested):

runtime/mandala/src/lib.rs
```diff
@@ -2602,4 +2615,104 @@ mod tests {
...
+				frame_system::CheckGenesis::<Runtime>::new(),
+				frame_system::CheckEra::<Runtime>::from(generic::Era::Immortal),
+				runtime_common::CheckNonce::<Runtime>::from(3),
+				frame_system::CheckWeight::<Runtime>::new(),
+				module_transaction_payment::ChargeTransactionPayment::<Runtime>::from(0),
+				module_evm::SetEvmOrigin::<Runtime>::new(),
+			);
+
+			let mut expected_extra = extra.clone();
+			expected_extra.4 .1 = true;
```
zjb0807, line 2658:
```suggestion
			expected_extra.4.1 = true;
```

runtime/common/src/check_nonce.rs
```diff
@@ -0,0 +1,285 @@
...
+		} else if self.0 != account.nonce {
+			return Err(if self.0 < account.nonce {
+				InvalidTransaction::Stale
+			} else {
+				InvalidTransaction::Future
+			}
+			.into());
+		}
+		account.nonce += T::Index::one();
+		frame_system::Account::<T>::insert(who, account);
```
zjb0807, line 110: EvmNonce shouldn't do this.

runtime/common/src/check_nonce.rs
```diff
@@ -0,0 +1,285 @@
...
+///
+/// This extension affects `requires` and `provides` tags of validity, but DOES NOT
+/// set the `priority` field. Make sure that AT LEAST one of the signed extension sets
+/// some kind of priority upon validating transactions.
+#[derive(Encode, Decode, Clone, Eq, PartialEq, TypeInfo)]
+#[scale_info(skip_type_params(T))]
+pub struct CheckNonce<T: frame_system::Config + module_evm::Config>(
+	#[codec(compact)] pub T::Index,
+	#[codec(skip)] pub bool, /* should check evm nonce */
+);
```
zjb0807, line 43: I think enum is better.
```
pub enum CheckNonce<T: frame_system::Config + module_evm::Config>(
	Nonce(#[codec(compact)] pub T::Index),
	EvmNonce(#[codec(skip)] pub T::Index),
);
```
Only change this:
https://github.com/AcalaNetwork/Acala/blob/78dfe2bb82/ts-tests/tests/test-sign-eth.ts#L122

zjb0807 (commented):
Please merge master.  `AcalaMultiSignature::Eip1559` was added.

runtime/common/src/check_nonce.rs
xlc: enum will break compatibility with existing extrinsic format. could have a custom Encode/Decode implementation to fix that

xlc: I think we should just do named fields

runtime/common/src/check_nonce.rs
```diff
@@ -0,0 +1,285 @@
...
+		_call: &Self::Call,
+		_info: &DispatchInfoOf<Self::Call>,
+		_len: usize,
+	) -> Result<(), TransactionValidityError> {
+		let mut account = frame_system::Account::<T>::get(who);
+		if self.1 {
+			// should check evm nonce
+			let address = <T as module_evm::Config>::AddressMapping::get_evm_address(who)
+				.unwrap_or_else(|| <T as module_evm::Config>::AddressMapping::get_default_evm_address(who));
+			let account = module_evm::Accounts::<T>::get(&address)
```
xlc, line 91: `let nonce = module_evm::Accounts::<T>::get(&address).map(|acc| acc.nonce).unwrap_or_default()`

runtime/common/src/check_nonce.rs
```diff
@@ -0,0 +1,285 @@
...
+				.unwrap_or_else(|| module_evm::AccountInfo::<T::Index>::new(Default::default(), None));
+			if self.0 < account.nonce {
+				return InvalidTransaction::Stale.into();
+			}
+
+			let provides = vec![Encode::encode(&(who, self.0, true /* evm */))];
```
xlc, line 131:
```suggestion
			let provides = vec![Encode::encode(&(who, self.0, b"evm"))];
```

runtime/mandala/src/lib.rs
```diff
@@ -1974,30 +1978,36 @@ impl Convert<(Call, SignedExtra), Result<EthereumTransactionMessage, InvalidTran
 					return Err(InvalidTransaction::Stale);
 				}
 
-				let era: frame_system::CheckEra<Runtime> = extra.3;
+				let era: frame_system::CheckEra<Runtime> = extra.clone().3;
```
xlc, line 1981:
```suggestion
				let era: frame_system::CheckEra<Runtime> = extra.3.clone();
```
Not 100% sure this works, but should have a way to only do one clone instead one per accessed field.

Otherwise could be 

`let (_, _, _, era, check_nonce, _, tip) = extra.clone();

runtime/common/src/check_nonce.rs
zjb0807: We must consider the increase of `evm.nonce`. We use `transactional` on `eth_call`, if the transaction fails, `evm.nonce` will not increase, then we will see the same nonce of 2 transactions.

ermalkaleci: @zjb0807 yes I'm thinking about that. There's also the case when tx isn't evm call but it still needs to increment nonce

xlc (commented):
I am not sure if this can handle recursive calls. i.e. when an evm call somehow triggers a dispatch of another evm call. this is not yet possible but could be in future.

ermalkaleci:
> I am not sure if this can handle recursive calls. i.e. when an evm call somehow triggers a dispatch of another evm call. this is not yet possible but could be in future.

@xlc maybe we can do take so skip_nonce_incremental is been read once

modules/evm/src/lib.rs
```diff
@@ -742,7 +737,6 @@ pub mod module {
 		/// - `gas_limit`: the maximum gas the call can use
 		/// - `storage_limit`: the total bytes the contract's storage can increase by
 		#[pallet::weight(T::GasToWeight::convert(*gas_limit))]
-		#[transactional]
```
xlc, line 745: why remove this?

modules/evm/src/runner/stack.rs
```diff
@@ -353,6 +360,8 @@ impl<T: Config> RunnerT<T> for Runner<T> {
 			)
 		})?;
 
+		Pallet::<T>::inc_nonce(source);
```
xlc, line 363: I don't think this is correct.

modules/evm/src/runner/stack.rs
```diff
@@ -317,6 +322,8 @@ impl<T: Config> RunnerT<T> for Runner<T> {
...
```
xlc, line 325: why increase nonce here?

modules/evm/src/runner/stack.rs
```diff
@@ -273,6 +276,8 @@ impl<T: Config> RunnerT<T> for Runner<T> {
...
```
xlc, line 279: why increase nonce after execute?

xlc (changes requested):
A DispatchError should still not resulting any state change. This means you will want to make it success even if the evm execution failed and emit another event to indicate the error

modules/evm/src/runner/stack.rs
ermalkaleci: I just rearranged the code but the nonce incremental order is the same, before call and after create

xlc: this doesn't make much sense. something isn't right.

zjb0807:
@ermalkaleci I think doing it in `post_dispatch` is better.  Avoid modifying the EVM.

ermalkaleci:
> @ermalkaleci I think doing it in `post_dispatch` is better. Avoid modifying the EVM.

@zjb0807 nonce is incremented before `call` and after `create`

zjb0807:
> @zjb0807 nonce is incremented before `call` and after `create`

But I think the `#[transactional]` is more important. The nonce will be increased in the evm module. If the tx return error before `inc_nonce` in evm, the nonce will be increased in `post_dispatch`.

zjb0807:
https://github.com/AcalaNetwork/Acala/blob/b8e0d02334984582fc09b11c2ecdae08105a8b95/modules/evm/src/runner/stack.rs#L226-L229

Such as this error, we also need to increase nonce.

zjb0807:
https://github.com/AcalaNetwork/Acala/blob/b8e0d02334984582fc09b11c2ecdae08105a8b95/modules/evm/src/runner/state.rs#L384
https://github.com/AcalaNetwork/Acala/blob/b8e0d02334984582fc09b11c2ecdae08105a8b95/modules/evm/src/runner/state.rs#L577


Or remove these `inc_nonce` in evm module,  increase the evm nonce  like `frame_system::Account nonce` in `pre_dispatch`.

zjb0807:
https://github.com/ethereum/go-ethereum/blob/12f971fb2d/core/state_transition.go#L321

It seems that there is no need to guarantee that the nonce must be increased. In ethereum, there are some checks before increasing nonce.

runtime/common/src/check_nonce.rs
```diff
@@ -0,0 +1,347 @@
...
+			let evm_nonce = module_evm::Accounts::<T>::get(&address)
+				.map(|x| x.nonce)
+				.unwrap_or_default();
+			if self.nonce < evm_nonce {
+				return InvalidTransaction::Stale.into();
+			}
+
+			let provides = vec![Encode::encode(&(who, self.nonce, b"eth_tx"))];
+			let requires = if evm_nonce < self.nonce {
+				vec![Encode::encode(&(who, self.nonce - One::one(), b"eth_tx"))]
```
zjb0807, line 165:
```suggestion
			let provides = vec![Encode::encode(&(address, self.nonce, b"eth_tx"))];
			let requires = if evm_nonce < self.nonce {
				vec![Encode::encode(&(address, self.nonce - One::one(), b"eth_tx"))]
```

@xlc Should I need to use `address` instead of `who`?

xlc: this code is only triggered with eth signature so best to use the eth address

shaunxw: we might not need `b"eth_tx"` if eth address is used?

shaunxw (approved):

xlc (approved):

PR #1708, opened 2021-12-17
Refactor/event to use named fields

syan095:
Refactored events in Modules folder and Ecosystem folder.
Refactored events in stable-asset module


runtime/integration-tests/src/authority.rs
```diff
@@ -266,78 +266,19 @@ fn test_authority_module() {
...
-				6,
-			)));
+			System::assert_last_event(Event::Authority(orml_authority::Event::Cancelled {
+				origin: OriginCaller::system(RawOrigin::Root),
+				index: 6,
+			}));
 		});
 }
-
-#[test]
```
zjb0807, line 286: Why remove this?

syan095: this must be a mistake (perhaps merge error).
I'll add this test back and double check agian

zjb0807 (approved):

PR #1721, opened 2021-12-22
Skip idle_scheduler Calls When Not Producing blocks 

### LINKED ISSUE #1467: Skip idle scheduler if block production took too long
Followup of #1466

Skip calling any tasks if time since last block is > 3 * expected block time, which indicates something is wrong and we want to avoid doing any extra work.

ferrell-code:
closes #1467


xlc (commented):
I don't fully get the formula. While we currently expecting one para block per two relay blocks, that will change in future and we shouldn't assume that.

Also `BlockNumberDifference` doesn't get updated on the `do_dispatch_tasks` path so it will get triggered every few blocks?

ferrell-code:
Yeah I could use `aura` slots instead but our consensus engine could change too. 

Under this solution it would get triggered every 3 slots skipped... so under perfect conditions it would never update `BlockNumberDifference`. It felt silly to constantly write to storage

modules/idle-scheduler/src/lib.rs
```diff
@@ -58,6 +59,15 @@ pub mod module {
...
 		type MinimumWeightRemainInBlock: Get<Weight>;
+
+		/// Gets RelayChain Block Number
+		type RelayChainBlockNumberProvider: BlockNumberProvider;
+
+		/// Number of Relay Chain blocks skipped to disable `on_idle` dispatching scheduled tasks
+		/// this shuts down idle-scheduler when block production is slower than this number of
+		/// relaychain blocks
+		#[pallet::constant]
+		type SkipRelayBlocks: Get<BlockNumber>;
```
xlc, line 70: This name isn't very clear, maybe something like `DisableBlockThreshold`?

modules/idle-scheduler/src/lib.rs
```diff
@@ -77,13 +87,51 @@ pub mod module {
...
+	pub type PreviousRelayBlockNumber<T: Config> = StorageValue<_, BlockNumber, ValueQuery>;
+
 	#[pallet::pallet]
 	pub struct Pallet<T>(_);
 
 	#[pallet::hooks]
 	impl<T: Config> Hooks<T::BlockNumber> for Pallet<T> {
+		fn on_initialize(_n: T::BlockNumber) -> Weight {
+			// Unwrap shouldn't fail if it does defaults to 0
+			let test: BlockNumber = T::RelayChainBlockNumberProvider::current_block_number()
```
xlc, line 102:
```suggestion
			let prev_relay: BlockNumber = T::RelayChainBlockNumberProvider::current_block_number()
```

And add a comment explain this is actually the previous relay chain block because `on_initialize` is executed before inherent are executed, which sets the new relay chain block number

modules/idle-scheduler/src/lib.rs
```diff
@@ -77,13 +87,51 @@ pub mod module {
...
+		}
+
 		fn on_idle(_n: T::BlockNumber, remaining_weight: Weight) -> Weight {
-			Self::do_dispatch_tasks(remaining_weight)
+			// Checks if we have skipped enough relay blocks without block production to skip dispatching
+			// scheduled tasks
+			let current_relay_block_number: BlockNumber = T::RelayChainBlockNumberProvider::current_block_number()
+				.try_into()
+				.unwrap_or_default();
+			let previous_relay_block_number = Self::previous_relay_block();
```
xlc, line 115:
```suggestion
			let previous_relay_block_number = PreviousRelayBlockNumber::<T>::take();
```

and no need to `kill`

ferrell-code:
We might want to wait for #1734 to merge. IdleScheduler needs to execute `on_initialize` in this PR before ParachainSystem, therefore currently we would need to change the order they are listed in `construct_runtime` and then change back if we want to implement the more logical order of hooks executing

ferrell-code:
Need to fix mock on `evm-tests` before merge

modules/idle-scheduler/src/lib.rs
```diff
@@ -76,14 +86,53 @@ pub mod module {
...
 
 	#[pallet::hooks]
 	impl<T: Config> Hooks<T::BlockNumber> for Pallet<T> {
+		fn on_initialize(_n: T::BlockNumber) -> Weight {
+			// Unwrap shouldn't fail if it does defaults to 0
+			// This is the previous relay block because `on_initialize` is executed before
+			// the inherent that sets the new relay chain block number
+			let previous_relay_block: BlockNumber = T::RelayChainBlockNumberProvider::current_block_number()
+				.try_into()
+				.unwrap_or_default();
```
shaunxw, line 106: In case it fails, it might be better to skip the update instead of using zero.

modules/idle-scheduler/src/lib.rs
```diff
@@ -76,14 +86,53 @@ pub mod module {
...
+			PreviousRelayBlockNumber::<T>::put(previous_relay_block);
+			// One write and one read
+			T::DbWeight::get().reads_writes(One::one(), One::one())
```
shaunxw, line 109: Should use benchmark. Weight could change if `on_initialize` updated in the future, and it's easy to miss weight update then.

modules/idle-scheduler/src/mock.rs
```diff
@@ -65,13 +66,16 @@ impl frame_system::Config for Runtime {
...
 	pub const MinimumWeightRemainInBlock: Weight = 100_000_000_000;
+	pub const DisableBlockThreshold: BlockNumber = 6;
 );
 
 impl module_idle_scheduler::Config for Runtime {
 	type Event = Event;
 	type WeightInfo = ();
 	type Task = ScheduledTasks;
 	type MinimumWeightRemainInBlock = MinimumWeightRemainInBlock;
+	type RelayChainBlockNumberProvider = System;
```
shaunxw, line 77: I would suggest to use a mock block number provider, to make it explicit in tests that the provider is from relay chain instead of system.

runtime/acala/src/lib.rs
```diff
@@ -1528,13 +1528,18 @@ define_combined_task! {
...
+	// once this number is reached idle scheduler is disabled as block production is slow
+	pub DisableBlockThreshold: BlockNumber = 6;
 );
 
 impl module_idle_scheduler::Config for Runtime {
 	type Event = Event;
 	type WeightInfo = ();
 	type Task = ScheduledTasks;
 	type MinimumWeightRemainInBlock = MinimumWeightRemainInBlock;
+	type RelayChainBlockNumberProvider = cumulus_pallet_parachain_system::RelaychainBlockNumberProvider<Runtime>;
```
shaunxw, line 1541: There a `RelaychainBlockNumberProvider` in `runtime_common` doing the same thing, can you remove it and replace imports in runtime with `cumulus_pallet_parachain_system`.

modules/idle-scheduler/src/lib.rs
ferrell-code: Fine by me :+1:. However, it really will never fail unless our `BlockNumber` and relaychain `BlockNumber` are different types (u32 right now) which seems unlikely, but no qed possible so no `.expect()`

ferrell-code:
Added more comprehensive benchmarks to `on_idle` too

shaunxw (approved):

modules/idle-scheduler/src/lib.rs
```diff
@@ -76,14 +86,56 @@ pub mod module {
...
 	#[pallet::without_storage_info]
 	pub struct Pallet<T>(_);
 
 	#[pallet::hooks]
 	impl<T: Config> Hooks<T::BlockNumber> for Pallet<T> {
+		fn on_initialize(_n: T::BlockNumber) -> Weight {
+			// `try_into()` shouldn't fail, if it does, storage will be left empty which will return a value of
+			// zero as default when read. This is the previous relay block because `on_initialize` is executed
+			// before the inherent that sets the new relay chain block number
+			let previous_relay_block: Option<BlockNumber> =
```
xlc, line 104: we should have a `type BlockNumberOf<T>` using the type from `BlockNumberProvider` so we don't need to deal with 3 different (but the same) block number types here.

xlc (approved):

shaunxw (approved):

Issue #1723, opened 2021-12-23
Update asset registry to support StableAssetPoolToken


xlc:
https://github.com/AcalaNetwork/Acala/blob/bbfa64f3bfee44753f4b2c0d7e05ebc4ef7150be/modules/asset-registry/src/lib.rs#L472

https://github.com/AcalaNetwork/Acala/blob/bbfa64f3bfee44753f4b2c0d7e05ebc4ef7150be/modules/asset-registry/src/lib.rs#L517

https://github.com/AcalaNetwork/Acala/blob/bbfa64f3bfee44753f4b2c0d7e05ebc4ef7150be/modules/asset-registry/src/lib.rs#L549

PR #1727, opened 2021-12-23
Add register_stable_asset and update_stable_asset

zjb0807:
Closes: #1723


xlc (approved):
I am not sure if we want to maintain `NextStableAssetId`  here. It could be in the stable asset pallet.

But I guess it will make everything easier to do this way so I am fine with that.

modules/asset-registry/src/lib.rs
```diff
@@ -244,41 +327,76 @@ impl<T: Config> Pallet<T> {
...
+			let old_multi_locations = maybe_multi_locations.as_mut().ok_or(Error::<T>::AssetIdNotExists)?;
+
+			AssetMetadatas::<T>::try_mutate(
+				AssetIds::ForeignAssetId(foreign_asset_id),
+				|maybe_asset_metadatas| -> DispatchResult {
+					ensure!(maybe_asset_metadatas.is_some(), Error::<T>::AssetIdNotExists);
+
+					// modify location
+					if location != old_multi_locations {
+						LocationToCurrencyIds::<T>::remove(old_multi_locations.clone());
```
xlc, line 339:
```suggestion
						LocationToCurrencyIds::<T>::remove(&old_multi_locations);
```

zjb0807: It was used here.
```
*old_multi_locations = location.clone();
```

zjb0807:
> I am not sure if we want to maintain `NextStableAssetId` here. It could be in the stable asset pallet.
> 

Maintaining NextStableAssetId here is really bad. But ed will be checked during `mint`, if it is not registered, `mint` will fail, so it will be forced to be registered.

PR #1730, opened 2021-12-24
Add Deposit for Setting Alternative Fee Swap Path

ferrell-code:
Deposit prevent bloated chain state


modules/transaction-payment/src/lib.rs
```diff
@@ -419,8 +424,10 @@ pub mod module {
 					Error::<T>::InvalidSwapPath
 				);
 				AlternativeFeeSwapPath::<T>::insert(&who, &path);
+				T::Currency::reserve_named(&RESERVE_ID, &who, T::AlternativeFeeSwapDeposit::get())?;
```
xlc, line 427: use `ensure_reserved_named` to avoid duplicated reserve when update path

modules/transaction-payment/src/lib.rs
```diff
@@ -419,8 +424,10 @@ pub mod module {
...
 			} else {
 				AlternativeFeeSwapPath::<T>::remove(&who);
+				T::Currency::unreserve_named(&RESERVE_ID, &who, T::AlternativeFeeSwapDeposit::get());
```
xlc, line 430: use `unreserve_all_named` to handle the case when deposit changes

xlc (changes requested):
need a different `RESERVE_ID` to avoid mixing them

modules/transaction-payment/src/lib.rs
```diff
@@ -64,6 +64,7 @@ pub use weights::WeightInfo;
 /// Fee multiplier.
 pub type Multiplier = FixedU128;
 
+pub type BalanceOf<T> = <<T as Config>::Currency as Currency<<T as frame_system::Config>::AccountId>>::Balance;
```
xlc, line 67: isn't this the same as `PalletBalanceOf` below?

ferrell-code: Yeah don't know how I missed that :skull:

xlc (approved):

PR #1739, opened 2021-12-30
fix remove contract

ermalkaleci:
current impl is not removing contract_info

### DIFF
--- a/modules/evm/src/lib.rs
+++ b/modules/evm/src/lib.rs
@@ -982,12 +982,12 @@ impl<T: Config> Pallet<T> {
 		// Deref code, and remove it if ref count is zero.
 		Accounts::<T>::mutate_exists(&address, |maybe_account| {
 			if let Some(account) = maybe_account {
-				if let Some(contract_info) = account.contract_info.clone() {
-					CodeInfos::<T>::mutate_exists(&contract_info.code_hash, |maybe_code_info| {
-						if let Some(code_info) = maybe_code_info.as_mut() {
+				if let Some(ContractInfo { code_hash, .. }) = account.contract_info {
+					CodeInfos::<T>::mutate_exists(&code_hash, |maybe_code_info| {
+						if let Some(code_info) = maybe_code_info {
 							code_info.ref_count = code_info.ref_count.saturating_sub(1);
 							if code_info.ref_count == 0 {
-								Codes::<T>::remove(&contract_info.code_hash);
+								Codes::<T>::remove(&code_hash);
 								account.contract_info = None;
 								*maybe_code_info = None;
 							}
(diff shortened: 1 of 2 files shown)


modules/evm/src/lib.rs
```diff
@@ -980,21 +980,22 @@ impl<T: Config> Pallet<T> {
...
-			CodeInfos::<T>::mutate_exists(&contract_info.code_hash, |maybe_code_info| {
-				if let Some(code_info) = maybe_code_info.as_mut() {
-					code_info.ref_count = code_info.ref_count.saturating_sub(1);
-					if code_info.ref_count == 0 {
-						Codes::<T>::remove(&contract_info.code_hash);
-						*maybe_code_info = None;
-					}
+		Accounts::<T>::mutate_exists(&address, |maybe_account| {
+			if let Some(account) = maybe_account {
+				if let Some(contract_info) = account.contract_info.clone() {
```
xlc, line 985:
```suggestion
				if let Some(ref contract_info) = account.contract_info {
```

I think this can avoid the clone, or `if let Some(contract_info) = account.contract_info.as_ref()`

xlc (approved):

zjb0807 (commented):
https://github.com/AcalaNetwork/Acala/blob/e373273079cbeef5a5ccebb6815b0d211c663ef2/modules/evm/src/lib.rs#L980-L1011

remove line L1000-L1008 and add `debug_assert!(false);` in `mutate_exists`.

ermalkaleci:
> https://github.com/AcalaNetwork/Acala/blob/e373273079cbeef5a5ccebb6815b0d211c663ef2/modules/evm/src/lib.rs#L980-L1011
> 
> remove line L1000-L1008 and add `debug_assert!(false);` in `mutate_exists`.

this will assert if remove_account is called while contract ref > 1

zjb0807:
> this will assert if remove_account is called while contract ref > 1

```
Accounts::<T>::mutate_exists(&address, |maybe_account| {
    if let Some(account) = maybe_account {
        if let Some(ContractInfo { code_hash, .. }) = account.contract_info {
            CodeInfos::<T>::mutate_exists(&code_hash, |maybe_code_info| {
                if let Some(code_info) = maybe_code_info {
                    code_info.ref_count = code_info.ref_count.saturating_sub(1);
                    if code_info.ref_count == 0 {
                        Codes::<T>::remove(&code_hash);
                        *maybe_code_info = None;
                    }
                }
            });

            // remove_account can only be called when account is killed. i.e. providers == 0
            // but contract_info should maintain a provider
            // so this should never happen
            debug_assert!(false, "removed account while is still linked to contract info");
        } else {
            maybe_account = None;
	}
});
```

`debug_assert` is only enabled in non-optimized builds by default. I think this is better.

xlc (approved):

PR #1747, opened 2022-01-04
off-by-one

frankyin-factual:
always transfer one more ksm to fix rounding issue

@shengda

### DIFF
--- a/runtime/mandala/src/lib.rs
+++ b/runtime/mandala/src/lib.rs
@@ -1864,7 +1864,7 @@ impl orml_tokens::ConvertBalance<Balance, Balance> for ConvertBalanceHoma {
 			CurrencyId::Token(TokenSymbol::LKSM) => Homa::get_exchange_rate()
 				.reciprocal()
 				.and_then(|x| x.checked_mul_int(balance))
-				.and_then(|x: Balance| -> Option<Balance> { x.checked_add(1) })
+				.and_then(|x| x.checked_add(1))
 				.unwrap_or_default(),
 			_ => balance,
 		}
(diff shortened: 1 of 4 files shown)


xlc (commented):
Better to have some integration test to cover this kind of errors.
Also should use saturated add

frankyin-factual:
added integration tests. I don't think saturated add is the right semantics here because we do want to ensure the converted back balance >= the original balance. This means if +1 is an overflow, it should fail.

xlc:
But you are using `unwrap_or_default` so overflow become 0, which is more wrong than max value.

frankyin-factual:
Yeah, but when it becomes `0`, our program definitely won't work on the current transaction instead of failing on possible later transactions. This gives us much more visibility.

xlc:
ok. in that case can you add some comments in the code to explain why return 0 on edge case is better? also should use `and_then` instead of multiple `unwrap_of_default`

runtime/mandala/src/lib.rs
```diff
@@ -1855,13 +1855,16 @@ impl orml_tokens::ConvertBalance<Balance, Balance> for ConvertBalanceHoma {
...
+		 * transactions have a possibility to fail, and this is undesirable.
+		 */
 		match asset_id {
 			CurrencyId::Token(TokenSymbol::LKSM) => Homa::get_exchange_rate()
 				.reciprocal()
-				.unwrap_or_default()
-				.checked_mul_int(balance)
-				.unwrap_or_default()
-				.checked_add(1)
+				.and_then(|x: Ratio| -> Option<Balance> { x.checked_mul_int(balance) })
```
xlc, line 1866:
```suggestion
				.and_then(|x| x.checked_mul_int(balance))
```

this should work

xlc (approved):

PR #1754, opened 2022-01-06
math

frankyin-factual:
Fix collect_fee

cc @shengda


frankyin-factual:
This is to fix total supply can go down temporarily because of rounding errors in balance conversion.

xlc (approved):

xlc:
It will be better if the stable assets side have more test

PR #1755, opened 2022-01-06
Claim Account to use Eip-712

### LINKED ISSUE #1579: Use EIP712 for EVM claim account signature
This should also prevent relay attack across between Karura and Acala

ferrell-code:
closes #1579 

I believe this signature would work with no collision problems (Obviously still need to update tests/benchmarks)


modules/evm-accounts/src/lib.rs
```diff
@@ -236,14 +232,43 @@ impl<T: Config> Pallet<T> {
...
 		let mut r = [0u8; 65];
 		r[0..64].copy_from_slice(&sig.serialize()[..]);
 		r[64] = recovery_id.serialize();
-		EcdsaSignature::from_slice(&r)
+		r
 	}
+
+	fn verify_eip712_signature(eth_address: EvmAddress, sig: &[u8; 65]) -> Option<H160> {
+		let domain_hash = keccak256!("EIP712Domain(string name,string version,uint256 ss58Id,bytes32 salt)");
+		let tx_type_hash = keccak256!("Transaction(address claim)");
```
xlc, line 246:
```suggestion
		let tx_type_hash = keccak256!("Transaction(string address)");
```
The message payload should be the Substrate account in hex format.

ferrell-code: the substrate address and the evm address being claimed right?

xlc: we don't need evm address in the signing payload because that's recovered from signature. it is already included from the claim extrinsic parameter so we just compare it to the recovered one to make sure it matches

modules/evm-accounts/src/lib.rs
```diff
@@ -236,14 +214,55 @@ impl<T: Config> Pallet<T> {
...
+
+	fn evm_account_payload_hash(who: &T::AccountId) -> [u8; 32] {
+		let tx_type_hash = keccak256!("Transaction(string address)");
+		let mut tx_msg = tx_type_hash.to_vec();
+		tx_msg.extend(who.using_encoded(to_ascii_hex));
+		keccak_256(tx_msg.as_slice())
+	}
+
+	fn evm_account_domain_separator() -> [u8; 32] {
+		let domain_hash = keccak256!("EIP712Domain(string name,string version,uint256 ChainId,bytes32 salt)");
```
xlc, line 252:
```suggestion
		let domain_hash = keccak256!("EIP712Domain(string name,string version,uint256 chainId,bytes32 salt)");
```

runtime/karura/src/weights/module_evm_accounts.rs
```diff
@@ -1,6 +1,6 @@
 // This file is part of Acala.
 
-// Copyright (C) 2020-2022 Acala Foundation.
+// Copyright (C) 2020-2021 Acala Foundation.
```
zjb0807, line 3: Merge master and rerun it.

ts-tests/tests/test-claim-account-eip712.ts
```diff
@@ -0,0 +1,76 @@
...
+
+        const tx = context.provider.api.tx.evmAccounts.claimAccount(await signer.getAddress(), signature);
+
+        await new Promise(async (resolve) => {
+			tx.signAndSend(await alice.getSubstrateAddress(), (result) => {
+				if (result.status.isFinalized || result.status.isInBlock) {
+					resolve(undefined);
+				}
+			});
+        });
```
zjb0807, line 67: format

zjb0807 (commented):
LGTM

ts-tests/tests/test-claim-account-eip712.ts
ferrell-code: Do we use a particular formatter, like `cargo fmt` for .ts files

zjb0807: I use `vscode` to format it. 
Maybe you can try https://prettier.io/docs/en/install.html

zjb0807 (dismissed):
Create an issue in https://github.com/AcalaNetwork/bodhi.js/issues

xlc (approved):

xlc (approved):

Issue #1762, opened 2022-01-10
new reanchor compatible change


zqhxuyuan:
relates to: https://github.com/paritytech/polkadot/pull/4470

orml: https://github.com/open-web3-stack/open-runtime-module-library/pull/672

zqhxuyuan:
change `CurrencyIdConvert` to support canonical reanchor. we only need adapt `to-reserve` and `non-reserve` cases. make karura and bifrost as example.
- Karura -> [BNC] -> Bifrost. From Bifrost's point of view, the reanchored location change from `(1, Parachain(2001), GeneralKey("BNC"))` now to `(0, GeneralKey("BNC"))`. so Bifrost make sure compatibility before Karura upgrade.
- Karura -> [KAR] -> Bifrost. no change, From Bifrost's point of view, the reanchored location keep (1, Parachain(2000), GeneralKey("KAR")).
- Bifrost -> [BNC] -> Karura. no change, From Karura's point of view, the reanchored location keep (1, Parachain(2001), GeneralKey("BNC")).
- Bifrost -> [KAR] -> Karura. From Karura's point of view, the reanchored location change from `(1, Parachain(2000), GeneralKey("KAR"))` now to `(0, GeneralKey("KAR"))`. so Karura make sure compatibility before Bifrost upgrade.
- Karura <-> [KSM] <-> Bifrost. no change.

Issue #1662, opened 2021-12-02
CheckNonZeroSender


xlc:
https://github.com/paritytech/substrate/pull/10413

ferrell-code:
Closed by #1765

PR #1765, opened 2022-01-10
upgrade to v0.9.16

zqhxuyuan:
**xcm related:**
- [x] **canonical reanchored location**: closes: https://github.com/AcalaNetwork/Acala/issues/1762
    - relates to: https://github.com/paritytech/polkadot/pull/4470 
    - I found it's difficult to setup config only to testcase, so I use `integration-tests` feature flag for now. 
- [x] **statemine asset prefix location**
    - relates to: https://github.com/paritytech/cumulus/pull/831
    - ~~also use `integration-tests` to testcase which make sure RMRK migration works.~~

**removal Default:**
- [x] MaxEncodedLen & without_storage_info
    - [x] stable-asset module should have this too, otherwise it'll compile error: https://github.com/nutsfinance/stable-asset/pull/14
- [x] ~~ParentIsPreset (not sure new cumulus release will include this yet)~~
- [x] CheckNonZeroSender: https://github.com/AcalaNetwork/Acala/issues/1662

client:
- [x] cumulus consensus refactor

runtime new pallet:
- [x] add pallet_preimage to pallet_scheduler
    - [x] migration: https://github.com/paritytech/substrate/pull/10356
- [x] ~~add pallet_child_bounties to pallet_bounties~~(do not add for now as kusama runtime not added)

tests:
- [x] ~~kusama native asset: https://github.com/AcalaNetwork/Acala/issues/1768~~
    - ~~current karura and sibling both use same runtime, so their native asset both is KAR, need `integration-tests` to manual adapt/fixed the para_id to 2000~~
- [x] test-service(version <= 0.9.15 has compile error of client type, so moved here)

---

upgrade guide:
- **canonical reanchored location**: before we upgrade, make sure bifrost already compatible. and current it's already done in bifrost side.
- **statemine asset prefix location** we don't need compatible because we use foreign_asset_id internal, but after statemine upgrade, we must change RMRK location on asset-registry from `(1, X2(1000, GenealIndex(8)))` to `(1, X3(1000, PalletInstance(50), GeneralIndex(8)))`


primitives/src/currency.rs
```diff
@@ -205,6 +205,8 @@ create_currency_id! {
 		PHA("Phala Native Token", 12) = 170,
 		KINT("Kintsugi Native Token", 12) = 171,
 		KBTC("Kintsugi Wrapped BTC", 8) = 172,
+		RMRK("Rmrk", 10) = 173,
```
xlc, line 208: why add those?

runtime/acala/src/lib.rs
```diff
@@ -921,6 +927,8 @@ impl orml_vesting::Config for Runtime {
 parameter_types! {
 	pub MaximumSchedulerWeight: Weight = Perbill::from_percent(10) * RuntimeBlockWeights::get().max_block;
 	pub const MaxScheduledPerBlock: u32 = 10;
+	// Retry a scheduled item every 10 blocks (1 minute) until the preimage exists.
+	pub const NoPreimagePostponement: Option<u32> = Some(10);
```
xlc, line 931: 10 blocks is 2 mins for us. Use `MINUTES` to ensure it is 1 minute.
But I think we can make it 5 minutes here.

runtime/acala/src/lib.rs
```diff
@@ -1912,8 +1959,32 @@ pub type SignedPayload = generic::SignedPayload<Call, SignedExtra>;
...
+	Runtime,
+	AllPalletsWithSystem,
+	SchedulerMigrationV3,
+>;
+
+// Migration for scheduler pallet to move from a plain Call to a CallOrHash.
+pub struct SchedulerMigrationV3;
+impl frame_support::traits::OnRuntimeUpgrade for SchedulerMigrationV3 {
+	fn on_runtime_upgrade() -> frame_support::weights::Weight {
+		Scheduler::migrate_v2_to_v3()
```
xlc, line 1975: I think the current version we have is still V1

runtime/karura/Cargo.toml
```diff
@@ -323,3 +326,5 @@ try-runtime = [
 	"module-homa-xcm/try-runtime",
 	"module-session-manager/try-runtime",
 ]
+
+integration-tests = []
```
xlc, line 330: why is this needed?

runtime/karura/src/lib.rs
```diff
@@ -920,6 +925,8 @@ impl orml_vesting::Config for Runtime {
 parameter_types! {
 	pub MaximumSchedulerWeight: Weight = Perbill::from_percent(10) * RuntimeBlockWeights::get().max_block;
 	pub const MaxScheduledPerBlock: u32 = 10;
+	// Retry a scheduled item every 10 blocks (1 minute) until the preimage exists.
+	pub const NoPreimagePostponement: Option<u32> = Some(10);
```
xlc, line 929: make it 5 minutes

runtime/karura/src/lib.rs
```diff
@@ -1545,7 +1571,46 @@ parameter_types! {
 	pub ForeignAssetUnitsPerSecond: u128 = kar_per_second();
 	pub KarPerSecondAsBased: u128 = kar_per_second();
 }
+#[cfg(feature = "integration-tests")]
+parameter_types! {
```
xlc, line 1575: I think better to move those into another file

runtime/karura/src/lib.rs
```diff
@@ -2011,8 +2131,32 @@ pub type SignedPayload = generic::SignedPayload<Call, SignedExtra>;
...
+	Runtime,
+	AllPalletsWithSystem,
+	SchedulerMigrationV3,
+>;
+
+// Migration for scheduler pallet to move from a plain Call to a CallOrHash.
+pub struct SchedulerMigrationV3;
+impl frame_support::traits::OnRuntimeUpgrade for SchedulerMigrationV3 {
+	fn on_runtime_upgrade() -> frame_support::weights::Weight {
+		Scheduler::migrate_v2_to_v3()
```
xlc, line 2147: should be from v1 to v3

runtime/acala/src/lib.rs
xlc: actually we are on v2 but the storage version in the pallet is still v1 for some reason, so we need to set the storage version before run the migration code

runtime/karura/Cargo.toml
zqhxuyuan: integration-tests use some special config only in testcase, but it's not required for production config. so I use `integration-tests` feature flag in runtime

runtime/acala/src/lib.rs
zqhxuyuan: change to `migrate_v1_to_v3 `

primitives/src/currency.rs
zqhxuyuan: deleted now. when I test statemine prefix location break change. I use RMRK and RMRK_V2 as location migration.

runtime/acala/src/lib.rs
zqhxuyuan: change to 25 blocks, 25 * 12 = 300s = 5 min

runtime/karura/src/lib.rs
zqhxuyuan: done

sander2:
Sorry to intrude on your PR, but I have a question. Do you know if the following change is required as well? I suspect that maybe polkadot/substrate will automatically canonicalize it, in which case it doesn't matter, but tbh I am not completely sure where the new `simplify` function gets called. I'm asking because we use some very similar runtime code.
```diff
fn native_currency_location(id: CurrencyId) -> MultiLocation {
-    MultiLocation::new(1, X2(Parachain(ParachainInfo::get().into()), GeneralKey(id.encode())))
+    MultiLocation::new(0, X1(GeneralKey(id.encode())))
}
```

zqhxuyuan:
> 

don't need change `native_currency_location()`. as the new canonical reanchored only change the meaning of `to-reserve`/`non-reserve` case. i.e ParaB transfer token A to ParaA, from the point view of ParaA, the location is changed. But for the case of ParaA transfer token A to ParaB. we do not need to change `native_currency_location()` as xtokens will invoke reanchor, the for the point view of ParaB, the location not change, still got `(1, 2000, A)`.

primitives/src/signature.rs
```diff
@@ -94,18 +94,24 @@ impl TryFrom<AcalaMultiSignature> for ecdsa::Signature {
...
-			(Self::Ed25519(ref sig), who) => sig.verify(msg, &ed25519::Public::from_slice(who.as_ref())),
-			(Self::Sr25519(ref sig), who) => sig.verify(msg, &sr25519::Public::from_slice(who.as_ref())),
+			(Self::Ed25519(ref sig), who) => match ed25519::Public::from_slice(who.as_ref()) {
+				Ok(signer) => sig.verify(msg, &signer),
+				Err(()) => false,
+			},
+			(Self::Sr25519(ref sig), who) => match sr25519::Public::from_slice(who.as_ref()) {
+				Ok(signer) => sig.verify(msg, &signer),
+				Err(()) => false,
+			},
```
shaunxw, line 114: use `map_or` to replace nested match

runtime/acala/src/lib.rs
```diff
@@ -1705,6 +1731,21 @@ impl Convert<MultiLocation, Option<CurrencyId>> for CurrencyIdConvert {
 					_ => None,
 				}
 			}
+			// adapt for reanchor canonical location: https://github.com/paritytech/polkadot/pull/4470
+			MultiLocation {
+				parents: 0,
+				interior: X1(GeneralKey(key)),
+			} => {
+				let key = &key[..];
+				if let Ok(currency_id) = CurrencyId::decode(&mut &*key) {
```
shaunxw, line 1740:
```rust
let currency_id = CurrencyId::decode(&mut &*key).ok()?;
```

runtime/acala/src/lib.rs
```diff
@@ -933,6 +939,24 @@ impl pallet_scheduler::Config for Runtime {
 	type MaxScheduledPerBlock = MaxScheduledPerBlock;
 	type WeightInfo = ();
 	type OriginPrivilegeCmp = EqualPrivilegeOnly;
+	type PreimageProvider = Preimage;
+	type NoPreimagePostponement = NoPreimagePostponement;
+}
+
+parameter_types! {
+	pub const PreimageMaxSize: u32 = 4096 * 1024;
```
shaunxw, line 947: Can we have comments about the actual size?

runtime/karura/src/lib.rs
```diff
@@ -1804,10 +1837,29 @@ impl Convert<MultiLocation, Option<CurrencyId>> for CurrencyIdConvert {
...
 			} => Some(Token(PHA)),
+			// adapt for reanchor canonical location: https://github.com/paritytech/polkadot/pull/4470
+			MultiLocation {
+				parents: 0,
+				interior: X1(GeneralKey(key)),
+			} => match &key[..] {
+				#[cfg(feature = "integration-tests")]
+				parachains::bifrost::BNC_KEY => Some(Token(BNC)),
+				key => {
+					if let Ok(currency_id) = CurrencyId::decode(&mut &*key) {
```
shaunxw, line 1848: let currency_id = CurrencyId::decode(&mut &*key).ok()?;

runtime/karura/src/lib.rs
```diff
@@ -1804,10 +1837,29 @@ impl Convert<MultiLocation, Option<CurrencyId>> for CurrencyIdConvert {
...
+						match currency_id {
+							Token(KAR) | Token(KUSD) | Token(LKSM) => Some(currency_id),
```
shaunxw, line 1850: do we need `TAI`

primitives/src/signature.rs
```diff
@@ -94,18 +94,24 @@ impl TryFrom<AcalaMultiSignature> for ecdsa::Signature {
...
-impl Default for AcalaMultiSignature {
-	fn default() -> Self {
-		Self::Ed25519(Default::default())
-	}
-}
+// impl Default for AcalaMultiSignature {
+// 	fn default() -> Self {
+// 		Self::Ed25519(Default::default())
+// 	}
+// }
```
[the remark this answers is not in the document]
zqhxuyuan: there'll be error: the trait bound `sp_core::ed25519::Signature: std::default::Default` is not satisfied. as Default is removed everywhere. I'll delete this code, or change the parameter to some fixed value.

runtime/karura/src/lib.rs
zqhxuyuan: good catch, I think `TAI` can be also consider as our parachain side token, then we should add it.

modules/evm/src/lib.rs
```diff
@@ -1295,8 +1296,11 @@ impl<T: Config> Pallet<T> {
 		let digest = <frame_system::Pallet<T>>::digest();
 		let pre_runtime_digests = digest.logs.iter().filter_map(|d| d.as_pre_runtime());
 
-		let author = T::FindAuthor::find_author(pre_runtime_digests).unwrap_or_default();
-		T::AddressMapping::get_default_evm_address(&author)
+		if let Some(author) = T::FindAuthor::find_author(pre_runtime_digests) {
+			T::AddressMapping::get_default_evm_address(&author)
+		} else {
+			H160::default()
+		}
```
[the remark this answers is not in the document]
zqhxuyuan: the `find_author()` return `Option(author)` value now, previous the return value is just `author`.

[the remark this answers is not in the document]
zjb0807: LGTM.
`H160::default()` is better than `T::AddressMapping::get_default_evm_address(AccountId::default())`

ts-tests/tests/test-contract-methods.ts
```diff
@@ -63,9 +63,9 @@ describeWithAcala("Acala RPC (Contract Methods)", (context) => {
 		expect((await contract.chainId()).toString()).to.eq('595');
 	});
 
-	it("should get correct environmental coinbase", async function () {
-		expect((await contract.coinbase()).toString()).to.not.eq('0x0000000000000000000000000000000000000000');
-	});
+	// it("should get correct environmental coinbase", async function () {
+	// 	expect((await contract.coinbase()).toString()).to.not.eq('0x0000000000000000000000000000000000000000');
```
zjb0807, line 67:
```suggestion
		expect((await contract.coinbase()).toString()).to.eq('0x0000000000000000000000000000000000000000');
```

modules/homa-lite/src/mock.rs
```diff
@@ -105,6 +105,10 @@ impl XcmTransfer<AccountId, Balance, CurrencyId> for MockXcm {
 	}
 }
 impl InvertLocation for MockXcm {
+	fn ancestry() -> MultiLocation {
+		todo!()
```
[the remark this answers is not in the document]
zqhxuyuan: add Parachain(2000) as Ancestry as this is a mock.

runtime/karura/src/lib.rs
zqhxuyuan: I checked mandala and acala runtime, `TAI` is only used in `ExistentialDeposits`, but not in `CurrencyIdConvert`, is it by design or a mistake? @xlc

runtime/karura/src/lib.rs
```diff
@@ -132,6 +137,7 @@ pub const VERSION: RuntimeVersion = RuntimeVersion {
 	impl_version: 0,
 	apis: RUNTIME_API_VERSIONS,
 	transaction_version: 1,
+	state_version: 1,
```
sander2, line 140: Just a heads-up: basti said on element that this has to be set to 0. I had it set to 1 initially as well, but it caused us to fail to produce blocks on a rococo-local setup

sander2:
Another headsup: I found in kintsugi testing on rococo-local that we needed to canonicalize the chain-local assetids in the `PerSecond` Trader items

zqhxuyuan:
> Another headsup: I found in kintsugi testing on rococo-local that we needed to canonicalize the chain-local assetids in the `PerSecond` Trader items

I'll revisit Trader config, as we're missing transfer native asset test, previous I add some code but delete them as I need to add `integration-tests` a few place. I think `KarPerSecondOfCanonicalLocation` should added to production case, not `integration-tests` case.

xlc (approved):

PR #1769, opened 2022-01-11
make evm always send events

zjb0807:


modules/evm/src/lib.rs
```diff
@@ -571,7 +571,32 @@ pub mod module {
...
+			.map_err(|e| {
+				Pallet::<T>::deposit_event(Event::<T>::ExecutedFailed {
+					from: source,
+					contract: target,
+					exit_reason: ExitReason::Error(ExitError::Other(Into::<&str>::into(e).into())),
+					output: vec![],
+					logs: vec![],
+				});
+				e
+			})?;
```
zjb0807, line 584: This is for deposit event then rollback transaction.

xlc: events are stored in storage so rollback will also remove the events.
so those extrinsic should not fail, they should always have ExtrinsicSuccess event but with another event to indicate the result.

xlc (changes requested):
Should have some tests to verify the events

xlc (approved):

Issue #1767, opened 2022-01-11
Use ExitReason::Revert instead of ExitReason::Error


zjb0807:
EVM does not return the error message when using `ExitReason::Error`
https://github.com/rust-blockchain/evm/blob/78c49debfc65ddf3c0635edd02045a937dcc15d5/runtime/src/eval/system.rs#L432-L448

So we should use `ExitReason::Revert` instead of `ExitReason::Error` like this:
https://github.com/AcalaNetwork/Acala/blob/bb432cb749e9365365c1c4fde96156e499c797e8/modules/evm/src/runner/state.rs#L773

TODO:
`ExitReason::Error(_)` and `ExitError::*.into()`
https://github.com/AcalaNetwork/Acala/blob/bb432cb749e9365365c1c4fde96156e499c797e8/modules/evm/src/runner/state.rs#L532
https://github.com/AcalaNetwork/Acala/blob/bb432cb749e9365365c1c4fde96156e499c797e8/modules/evm/src/runner/state.rs#L551

PR #1772, opened 2022-01-12
Use ExitReason::Revert instead of ExitReason::Error

zjb0807:
Closes: #1767

### DIFF
--- a/modules/evm/src/runner/stack.rs
+++ b/modules/evm/src/runner/stack.rs
@@ -551,7 +551,14 @@ impl<'vicinity, 'config, T: Config> BackendT for SubstrateStackState<'vicinity, 
 	}
 
 	fn code(&self, address: H160) -> Vec<u8> {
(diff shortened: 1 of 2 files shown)


modules/evm/src/runner/stack.rs
```diff
@@ -551,7 +551,14 @@ impl<'vicinity, 'config, T: Config> BackendT for SubstrateStackState<'vicinity,
 	}
 
 	fn code(&self, address: H160) -> Vec<u8> {
-		Pallet::<T>::code_at_address(&address).into_inner()
+		let code = Pallet::<T>::code_at_address(&address).into_inner();
+		log::debug!(
```
xlc, line 555: why log on getter?

zjb0807: If the contract does not exist, evm will return revert without error msg. 
It's hard to debug, so log here.

xlc (approved):

Issue #1734, opened 2021-12-28
Change AllPallets to AllPalletsWithSystem


ferrell-code:
When the next release comes out we should update runtime to https://github.com/paritytech/substrate/pull/10043. This changes the order that hooks are executed to the order the pallets are listed in `construct_runtime` rather than in reverse

ferrell-code:
closed with #1779

PR #1779, opened 2022-01-14
polkadot-release v0.9.15

zqhxuyuan:
- udpate upstream to v0.9.15
- AllPallets change to AllPalletsWithSystem as suggested: https://github.com/AcalaNetwork/Acala/issues/1734
- trivial fix of `inspect` about hex hash

~~PS: try-runtime still failed CI, I think parity wouldn't fix as the v0.9.15 is already released, but we can ignore it.~~


ferrell-code (commented):
I'm not entirely sure why the order of these matter https://github.com/AcalaNetwork/Acala/blob/95c483d706e41e5694093f0ed04401fde60d4532/runtime/acala/src/lib.rs#L1780-#L1784

On statemine/statemint they didn't change the order with the change to `AllPalletsWithSystem` so probably ok

I think they are gonna backport another fix in polkadot so it can pass CI

zqhxuyuan:
I've checked the latest v0.9.15 of polkadot/cumulus runtime, they're using `AllPallets` instead of `AllPalletsWithSystem`. so we're align at them?

ferrell-code:
good catch I was looking at master branch... Its always safer to have someone else go first :shrug:, I defer to bryan on that

zqhxuyuan:
current in v0.9.15, the AllPallets is equal to AllPalletsWithSystem, so it's ok to use later one which is suggested.

https://github.com/paritytech/substrate/blob/polkadot-v0.9.15/frame/support/procedural/src/construct_runtime/mod.rs#L361

xlc (approved):

PR #1782, opened 2022-01-14
stable-asset-it-tests

frankyin-factual:
add stress tests for stable asset


xlc (changes requested):
Should really be some fuzzing test like https://github.com/paritytech/substrate/blob/master/frame/bags-list/fuzzer/src/main.rs

I don't like the idea of random test, because the are flaky and can be slow to run, hard to figure out what's wrong it failed. I don't think you can easily figure out what random value was used for a broken test.

At very least this needs to be its own target, and a separate CI step.

Issue #1773, opened 2022-01-12
support metadata operation on asset-registry


zqhxuyuan:
with existence of `asset-registry` module, we can not only register statemine/statemint inner assets, I think we can also register `Token(TokenSymbol)` asset (may be also including our own native asset). here is quick thought:

1. a trait used to simulate all kinds of asset operation, and `asset-registry` module can implement this trait. we could extend the metadata of `asset-registry` to include `Token/CurrencyId` so that it will store `currency_id` to `location` and vice versa.

```rust
pub trait CurrencyMetadataOps {
        // replace ExistentialDeposits
	fn existential_deposit(currency_id: &CurrencyId) -> Balance;

        // replace Convert<CurrencyId, Option<MultiLocation>>
	fn currency_to_location(currency_id: CurrencyId) -> Option<MultiLocation>;
        // replace Convert<MultiLocation, Option<CurrencyId>>
	fn location_to_currency(location: MultiLocation) -> Option<CurrencyId>;
        // replace Convert<MultiAsset, Option<CurrencyId>>
	fn asset_to_currency(asset: MultiAsset) -> Option<CurrencyId> {
		if let MultiAsset {
			id: Concrete(location), ..
		} = asset
		{
			Self::location_to_currency(location)
		} else {
			None
		}
	}
        // Trader
	fn units_per_second(currency_id: CurrencyId) -> Balance;
}
```

2. refactor CurrencyId related mapping, such as `CurrencyIdConvert`, `Trader`, `ExistentialDeposits`.

IMO this will simplify new foreign asset onboard on acala side. what do you think @xlc @shaunxw @zjb0807

xlc:
This could be helpful to handle breaking changes of `MultiLocation`

PR #1785, opened 2022-01-16
sibling token rework

zqhxuyuan:
relates to: https://github.com/AcalaNetwork/Acala/issues/1773

the `asset-registry` module could used for not only statemine/statemint assets, but also sibling token(i.e BNC, PHA etc). thus for new sibling token onboard to acala, now we can use `register_foreign_asset` dispatch call of `asset-registry` instead of merge PR to acala code base from parachain team.

a unconenient effect is `asset-registry` use `foreign-asset-id` which is number instead of `TokenSymbol` enum, thus in front/dapp we may need `TokenSymbol` mapping to `foreign-asset-id`, make sure the token mapping is correct.

todo list:
- [] Trader

### DIFF
--- a/primitives/src/currency.rs
+++ b/primitives/src/currency.rs
@@ -200,11 +200,11 @@ create_currency_id! {
 		// 149: Reserved for renBTC
 		// 150: Reserved for CASH
 		// 168 - 255: Kusama parachain tokens
-		BNC("Bifrost Native Token", 12) = 168,
-		VSKSM("Bifrost Voucher Slot KSM", 12) = 169,
-		PHA("Phala Native Token", 12) = 170,
-		KINT("Kintsugi Native Token", 12) = 171,
-		KBTC("Kintsugi Wrapped BTC", 8) = 172,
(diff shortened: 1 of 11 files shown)


primitives/src/currency.rs
```diff
@@ -200,11 +200,11 @@ create_currency_id! {
 		// 149: Reserved for renBTC
 		// 150: Reserved for CASH
 		// 168 - 255: Kusama parachain tokens
-		BNC("Bifrost Native Token", 12) = 168,
-		VSKSM("Bifrost Voucher Slot KSM", 12) = 169,
-		PHA("Phala Native Token", 12) = 170,
-		KINT("Kintsugi Native Token", 12) = 171,
-		KBTC("Kintsugi Wrapped BTC", 8) = 172,
+		// BNC("Bifrost Native Token", 12) = 168,
```
xlc, line 203: This will break compatibility for both SDK and existing runtime storages (e.g. BNC swap pair)
It should be possible to do some migration but it isn't that much beneficial to migrate.

PR #1793, opened 2022-01-17
Add Developer Status and Contract Deploy Precompiles

### LINKED ISSUE #1675: EVM precompile for evm.deploy
Also need predeploy contract for it.

We would like to allow contract developers to publish their contracts directly within EVM.

### LINKED ISSUE #1676: EVM precompile for evm.enableContractDevelopment
Need precompile to enable, disable, query status for user about contract development mode

This allow user to manage contract development mode within EVM.

ferrell-code:
closes #1675 
closes #1676


zjb0807 (commented):
LGTM. The predeploy-contracts need to be updated.
https://github.com/AcalaNetwork/predeploy-contracts/blob/master/contracts/state_rent/IStateRent.sol

modules/evm/src/lib.rs
```diff
@@ -1036,14 +1036,8 @@ pub mod module {
 		#[transactional]
 		pub fn deploy(origin: OriginFor<T>, contract: EvmAddress) -> DispatchResultWithPostInfo {
```
xlc, line 1037: also rename this to publish_contract to align with the predeploy contract

ferrell-code: :+1: Note Changing the name will break some tests in bohdi.js and any thing else that uses `tx.evm.deploy()` downstream

modules/evm/src/lib.rs
```diff
@@ -1034,16 +1034,10 @@ pub mod module {
...
 			let who = ensure_signed(origin)?;
-			let address = T::AddressMapping::get_evm_address(&who).ok_or(Error::<T>::AddressNotMapped)?;
-			T::Currency::transfer(
-				&who,
-				&T::TreasuryAccount::get(),
-				T::DeploymentFee::get(),
-				ExistenceRequirement::AllowDeath,
-			)?;
-			Self::mark_deployed(contract, Some(address))?;
+			Self::do_deploy_contract(who, contract)?;
```
xlc, line 1039: should update this as well

ferrell-code: I think I may just do all of them like `deploy_free`, `mark_deployed` ... etc.

modules/evm/src/lib.rs
xlc: yeah that's fine. better to break it before we release it

xlc (approved):

PR #1799, opened 2022-01-18
srtool with features

ermalkaleci:
Update when this gets merged https://github.com/chevdor/srtool-actions/pull/9


xlc:
Can you verify if the resulting wasm is the same as the one from makefile command.

ermalkaleci:
> Can you verify if the resulting wasm is the same as the one from makefile command.

verified

syan095 (approved):

PR #1803, opened 2022-01-19
add module-dex-oracle to provide prices from DEX, and module-prices provide price for CurrencyId::Crowdloan

### LINKED ISSUE #1795: prices provide price for LcDOT(LiquidCroadloan)
There is no reliable market price feed for LcDOT in the initial stage, the price of LcDOT can be calculated by discount based on the market price of DOT, the average staking reward of polkadot, lease block and current polkadot block.

constant AVERAGE_STAKING_REWARD_RATE
constant YEAR_BLOCKS
per_block_staking_reward_rate = (100% + AVERAGE_STAKING_REWARD_RATE)**(1 / YEAR_BLOCKS) - 100%
interval = release_block_number - current_polkadot_block_number
lcdot_price = dot_price /  (100% - per_block_staking_reward_rate)**interval

@xlc  make sure the pr related to this issue have been available on Acala runtime before Acala use LcDOT as collateral

wangjj9219:
- [x] calculate prices for LcDOT/LcKSM
- [x] support cumulative prices for specifc DEX traiding pair

close #1795


modules/dex-oracle/src/lib.rs
```diff
@@ -0,0 +1,286 @@
...
+
+		/// Time provider
+		type Time: Time;
+
+		/// The origin which may manage dex oracle.
+		type UpdateOrigin: EnsureOrigin<Self::Origin>;
+
+		/// The time interval in millisecond for updating the cumulative prices.
+		#[pallet::constant]
+		type IntervalToUpdateCumulativePrice: Get<MomentOf<Self>>;
```
xlc, line 59: I am not sure if we want to use time or just block number. Which one is safer when block stopped for 30 mins?

modules/dex-oracle/src/lib.rs
```diff
@@ -0,0 +1,286 @@
...
+
+	fn get_cumulative_price(trading_pair: &TradingPair) -> Option<(ExchangeRate, ExchangeRate)> {
+		Self::cumulative_prices(trading_pair).map(|(price_0, price_1, _, _)| (price_0, price_1))
+	}
+}
+
+impl<T: Config> Happened<(TradingPair, Balance, Balance)> for Pallet<T> {
+	fn happened(info: &(TradingPair, Balance, Balance)) {
+		let (trading_pair, pool_0, pool_1) = *info;
+		Self::try_update_cumulative(&trading_pair, pool_0, pool_1);
```
xlc, line 232: why is this needed? modify liquidity shouldn't impact price?

wangjj9219: swap will modify the liquidity pool,  dex-module will trigger this hook when liquidity pool changed. Actually, add_liquidity/remove_liquidity doesn't need to trigger this hook because they won't impact the exchange rate

xlc: also why do we need to update the pool on every swap? only the first one on the block is counted anyway, so we can just do it on `on_initialize`

wangjj9219: Actively query the pool and update cumulative in on_initialize, there will always be n * read_and_write(3, 1) and cumulative calculation, and the weight cannot be changed as fee. If handle it at hook, each `swap` will be (path_len - 1) * read_and_write(1, 0) at least, (path_len - 1) * read_and_write(2, 1) at most,  this weight can be charged as fee after re-benchmarking dex. I don't know which is better.

modules/dex-oracle/src/lib.rs
wangjj9219: Maybe uniswap has considered this, TWAP stands for the time-weighted average price, the price is still available for the duration of that unstable block.  I prefer use time as the weight

xlc (approved):

PR #1806, opened 2022-01-24
WIP: Feature/perpetual ksm dot

syan095:
Added a new PRT module.
Refactored Homa-xcm to become a generic relaychain-interface module.
Added a new trait to manage NFT.

### DIFF
--- a/modules/homa-lite/src/lib.rs
+++ b/modules/homa-lite/src/lib.rs
@@ -500,8 +500,8 @@ pub mod module {
 				// If the amount is zero, cancel previous redeem request.
 				if let Some((request_amount, _)) = RedeemRequests::<T>::take(&who) {
 					// Unreserve the liquid fee and remove the redeem request.
-					let unreserved = T::Currency::unreserve(T::LiquidCurrencyId::get(), &who, request_amount);
-					ensure!(unreserved.is_zero(), Error::<T>::InsufficientReservedBalances);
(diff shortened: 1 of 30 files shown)


modules/homa-lite/src/lib.rs
```diff
@@ -127,7 +127,7 @@ pub mod module {
 
 		/// The fixed cost of withdrawing Staking currency via redeem. In Staking currency.
 		#[pallet::constant]
-		type XcmUnbondFee: Get<Balance>;
+		type HomaUnbondFee: Get<Balance>;
```
xlc, line 130: I think we can move those changes into other smaller PR and have it merged, to reduce the future diff size of this

modules/nft/src/lib.rs
```diff
@@ -556,3 +550,25 @@ impl<T: Config> NFT<T::AccountId> for Pallet<T> {
 		Self::do_transfer(from, to, token)
 	}
 }
+
+impl<T: Config> ManageNFT<T::AccountId, CID, Attributes> for Pallet<T> {
+	type ClassId = ClassIdOf<T>;
+	type TokenId = TokenIdOf<T>;
+
+	/// To mint new NFT tokens.
+	fn mint(
```
xlc, line 559: I think we can have `mint` and `mint_batch` to make interface easier to use.
We only have one for extrinsic because there is nothing to return.

PR #1807, opened 2022-01-24
stable-asset

frankyin-factual:
add stable asset to karura


frankyin-factual:
@xlc Not sure what's going on with benchmark: 
```
benchmarking::cdp_engine::tests::benchmark_tests::test_benchmarks
```

xlc (approved):

PR #1823, opened 2022-01-27
Optimize collateral auction

wangjj9219:
if collateral's target is not zero, when auction end, exists following situations:
1. if dex can provider same for target, dex take the auction
2. if the last bid price reach target and higher than dex, the auction dealt by last bidder
3. other, auction aborted.

if collateral's target is zero, when auction end, exists following situations:
1. if dex can provider same or more last bid price, dex take the auction.
2. if exists bid, the auction dealt by last bidder.
3. if no bid, auction aborted.


modules/auction-manager/src/lib.rs
```diff
@@ -584,98 +594,118 @@ impl<T: Config> Pallet<T> {
...
+					refund_collateral,
+				);
+				if let Err(e) = res {
+					log::warn!(
+						target: "auction-manager",
+						"withdraw_collateral: failed to withdraw {:?} {:?} from CDP treasury to {:?}: {:?}. \
+						This is unexpected but should be safe",
+						refund_collateral, collateral_auction.currency_id, collateral_auction.refund_recipient, e
+					);
+					debug_assert!(false);
```
xlc, line 632: this should be a helper method

modules/auction-manager/src/lib.rs
```diff
@@ -584,98 +594,118 @@ impl<T: Config> Pallet<T> {
...
 			Self::deposit_event(Event::CollateralAuctionDealt {
 				auction_id,
 				collateral_type: collateral_auction.currency_id,
 				collateral_amount: collateral_auction.amount,
-				winner: bidder,
+				winner,
 				payment_amount,
 			});
+		} else {
+			// refund stable to the last_bidder, it shouldn't fail and affect the process.
```
xlc, line 686: need some explanation on what happened here

xlc (commented):
not 100% sure if we want to abort the auction or just restart it
maybe we should only abort it if we already have too many ongoing auctions?

shaunxw (commented):
Some nitpicks.

I think we can introduce some restart mechanism if auction aborted, but it could be future improvements.

shaunxw (approved):
LGTM, will need to fix CI errors.

xlc (approved):

shaunxw (approved):

PR #1826, opened 2022-01-29
Fix state-rent precompile

zjb0807:


xlc (dismissed):

modules/evm/src/tests.rs
```diff
@@ -1925,6 +1925,7 @@ fn remove_account_with_provides_should_panic() {
 }
 
 #[test]
+#[should_panic(expected = "removed account while is still linked to contract info")]
```
ermalkaleci, line 1928: This doesn't look correct. Ref count is 1 so calling remove_account will make it zero resulting into account removal

zjb0807: we have `remove_contract` to remove the contract account. `remove_account` is only for removing a normal account. So no matter whether there is a ref or not, it should report an error in the develop mode

modules/evm/src/lib.rs
```diff
@@ -1194,25 +1194,21 @@ impl<T: Config> Pallet<T> {
...
-								account.contract_info = None;
 								*maybe_code_info = None;
 							}
 						}
 					});
+
+					// remove_account can only be called when account is killed. i.e. providers == 0
+					// but contract_info should maintain a provider
+					// so this should never happen
+					debug_assert!(false, "removed account while is still linked to contract info");
```
ferrell-code, line 1205: Maybe we should add a `log::warn!` here too?

ferrell-code (approved):

ermalkaleci (approved):

xlc (approved):

PR #1829, opened 2022-01-30
Fix pallet order

wangjj9219:


xlc (commented):
should have some integration test to verify the fix. but that can happen in later PR

xlc (approved):

modules/homa/src/lib.rs
```diff
@@ -623,6 +623,12 @@ pub mod module {
 
 			Ok(())
 		}
+
+		#[pallet::weight(< T as Config >::WeightInfo::reset_current_era())]
```
xlc, line 627: this should be `on_initialize_with_bump_era`

PR #1844, opened 2022-02-03
Revert reducible balance

zjb0807:
Closes: #1845


ermalkaleci:
@zjb0807 I think `remove_contract` needs to reset nonce https://github.com/AcalaNetwork/Acala/blob/master/modules/evm/src/lib.rs#L1150 so when `remove_account_if_empty` is called it will wipe the account

zjb0807:
> @zjb0807 I think `remove_contract` needs to reset nonce https://github.com/AcalaNetwork/Acala/blob/master/modules/evm/src/lib.rs#L1150 so when `remove_account_if_empty` is called it will wipe the account

https://github.com/AcalaNetwork/Acala/blob/faf97a0647336512dbb8abfb1f0ce47e13ce6f71/modules/evm/src/lib.rs#L1936
It is removed in `DispatchableTask`

xlc (approved):

PR #1848, opened 2022-02-06
Add Foreign State Oracle Module

ferrell-code:
First pass at foreign state oracle


modules/relaychain-oracle/src/lib.rs
```diff
@@ -0,0 +1,136 @@
...
+	}
+
+	#[pallet::call]
+	impl<T: Config> Pallet<T> {
+		#[pallet::weight(0)]
+		pub fn dispatch_task(origin: OriginFor<T>, call_hash: T::Hash, state_hash: T::Hash) -> DispatchResult {
+			ValidateTask::<T>::try_mutate_exists(call_hash, |maybe_verifiable_call| -> DispatchResult {
+				T::OracleOrigin::ensure_origin(origin)?;
+
+				let verifiable_call = maybe_verifiable_call.clone().ok_or(Error::<T>::NoMatchingCall)?;
```
xlc, line 103:
```suggestion
				let verifiable_call = maybe_verifiable_call.take().ok_or(Error::<T>::NoMatchingCall)?;
```

and no need to make it None later

modules/relaychain-oracle/src/lib.rs
```diff
@@ -0,0 +1,136 @@
...
+				ensure!(state_hash == verifiable_call.state_hash, Error::<T>::IncorrectStateHash);
+
+				let result = verifiable_call
+					.dispatchable_call
+					.dispatch(RawOrigin::RelaychainOracle.into());
+				*maybe_verifiable_call = None;
+				match result {
```
xlc, line 110: just have one event with `DispatchResult` argument so we don't need two events for each case.

modules/relaychain-oracle/src/lib.rs
```diff
@@ -0,0 +1,136 @@
...
+			})
+		}
+	}
+}
+
+impl<T: Config> Pallet<T> {}
+
+impl<T: Config> ForeignChainStateQuery<T::VerifiableTask, T::Hash> for Pallet<T> {
+	fn query_task(call: T::VerifiableTask, state_hash: T::Hash) {
+		let call_hash = T::Hashing::hash_of(&call);
```
xlc, line 124: call hash may not be unique, should just use an sequential index as ID

modules/relaychain-oracle/src/lib.rs
```diff
@@ -0,0 +1,136 @@
...
+	#[pallet::error]
+	pub enum Error<T> {
+		IncorrectStateHash,
+		NoMatchingCall,
+	}
+
+	#[pallet::call]
+	impl<T: Config> Pallet<T> {
+		#[pallet::weight(0)]
+		pub fn dispatch_task(origin: OriginFor<T>, call_hash: T::Hash, state_hash: T::Hash) -> DispatchResult {
```
xlc, line 99: not exactly sure what's the purpose of state hash.

it should be some arbitrary data that sent to the call.

it could be `Vec<u8>` and have the task call to decode it, or an enum similar to the IdleSchedulerTask

modules/relaychain-oracle/src/lib.rs
```diff
@@ -0,0 +1,136 @@
...
+			ValidateTask::<T>::try_mutate_exists(call_hash, |maybe_verifiable_call| -> DispatchResult {
+				T::OracleOrigin::ensure_origin(origin)?;
+
+				let verifiable_call = maybe_verifiable_call.clone().ok_or(Error::<T>::NoMatchingCall)?;
+				ensure!(state_hash == verifiable_call.state_hash, Error::<T>::IncorrectStateHash);
+
+				let result = verifiable_call
+					.dispatchable_call
+					.dispatch(RawOrigin::RelaychainOracle.into());
```
xlc, line 108: we need to pass the data into the call somehow.

one way is similar to XCM v3 code to reencode it, another way is to inject the data into the origin

modules/relaychain-oracle/src/lib.rs
ferrell-code: Yeah using a hash would make it so no duplicate calls can be in storage. A counter probably makes sense, but we need to make sure double calling a confirmation will not be problematic

modules/foreign-state-oracle/src/lib.rs
```diff
@@ -0,0 +1,229 @@
...
+mod tests;
+
+pub use module::*;
+
+// Unique Identifier for each query
+pub type QueryIndex = u64;
+
+#[derive(PartialEq, Eq, Clone, RuntimeDebug, Encode, Decode, TypeInfo)]
+pub enum RawOrigin {
+	ForeignStateOracle { data: Vec<u8> },
```
xlc, line 44: I don't think we need to make this an enum if we are not going to have additional variant.
Make it `struct RawOrigin { data: Vec<u8> }` will be enough.

modules/foreign-state-oracle/src/lib.rs
```diff
@@ -0,0 +1,229 @@
...
+	impl<T: Config> Pallet<T> {
+		#[pallet::weight(0)]
+		#[transactional]
+		pub fn dispatch_task(origin: OriginFor<T>, query_index: QueryIndex, data: Vec<u8>) -> DispatchResult {
+			T::OracleOrigin::ensure_origin(origin)?;
+
+			let verifiable_call = ActiveQuery::<T>::take(query_index).ok_or(Error::<T>::NoMatchingCall)?;
+			// check that query has not expired
+			ensure!(
+				verifiable_call.expiry > frame_system::Pallet::<T>::current_block_number(),
```
xlc, line 145: use `BlockNumberProvider`

modules/foreign-state-oracle/src/lib.rs
```diff
@@ -0,0 +1,229 @@
...
+			});
+
+			Ok(())
+		}
+
+		// TODO: Change to use idle scheduler
+		#[pallet::weight(0)]
+		#[transactional]
+		pub fn remove_expired_call(origin: OriginFor<T>, query_index: QueryIndex) -> DispatchResult {
+			ensure_none(origin)?;
```
xlc, line 164: maybe we should make it ensure_signed or none and refund half of the query fee to the caller of this method to give some incentive for people to purge expired call?

needs to implement ValidateUnsigned to actually support unsigned tx

modules/foreign-state-oracle/src/lib.rs
```diff
@@ -0,0 +1,229 @@
...
+		let expiry = frame_system::Pallet::<T>::current_block_number().saturating_add(T::QueryDuration::get());
+		let verifiable_call = VerifiableCall {
+			dispatchable_call,
+			expiry,
+			oracle_reward: T::QueryFee::get(),
+		};
+
+		let index = QueryCounter::<T>::get();
+		// Increment counter
+		QueryCounter::<T>::put(index + 1);
```
xlc, line 205: use checked_add to handle overflow

modules/support/src/lib.rs
```diff
@@ -661,3 +661,11 @@ pub trait HomaSubAccountXcm<AccountId, Balance> {
 	/// The fee of cross-chain transfer is deducted from the recipient.
 	fn get_xcm_transfer_fee() -> Balance;
 }
+
+pub trait ForeignChainStateQuery<AccountId, Call> {
+	/// Call to be validated by foreign state oracle
+	/// params:
+	/// - who: Account that is requesting the query,
+	/// - dispatchable_call: Call to be dispatched on the condition of a foreign chain state
+	fn query_task(who: &AccountId, length_bound: u32, dispatchable_call: Call) -> DispatchResult;
```
xlc, line 670: should also have a way to cancel it to have some of the deposit refunded

modules/foreign-state-oracle/src/lib.rs
ferrell-code: yeah giving half of the fee for oracles to purge storage makes sense.

ferrell-code:
merged into #1865

PR #1856, opened 2022-02-08
Fix transaction payment

zjb0807:
1. Enable integration tests for `mandala` and `acala` runtime
2. Modify `UpdateOrigin`
3. Fix `set_alternative_fee_swap_path`
4. Fix `enable_charge_fee_pool`

### DIFF
--- a/modules/transaction-payment/src/lib.rs
+++ b/modules/transaction-payment/src/lib.rs
@@ -329,7 +329,7 @@ pub mod module {
 		type TreasuryAccount: Get<Self::AccountId>;
 
 		/// The origin which change swap balance threshold or enable charge fee pool.
-		type UpdateOrigin: EnsureOrigin<Self::Origin, Success = Self::AccountId>;
+		type UpdateOrigin: EnsureOrigin<Self::Origin>;
 	}
 
 	#[pallet::extra_constants]
--- a/modules/transaction-payment/src/lib.rs
+++ b/modules/transaction-payment/src/lib.rs
@@ -820,8 +820,7 @@ where
 
 	/// Initiate a charge fee swap pool. Usually used in `on_runtime_upgrade` or manual
 	/// `enable_charge_fee_pool` dispatch call.
-	#[transactional]
-	pub fn initialize_pool(currency_id: CurrencyId, pool_size: Balance, swap_threshold: Balance) -> DispatchResult {
+	fn initialize_pool(currency_id: CurrencyId, pool_size: Balance, swap_threshold: Balance) -> DispatchResult {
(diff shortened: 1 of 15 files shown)


runtime/mandala/src/lib.rs
```diff
@@ -1162,7 +1163,12 @@ impl module_transaction_pause::Config for Runtime {
 
 parameter_types! {
 	// Sort by fee charge order
-	pub DefaultFeeSwapPathList: Vec<Vec<CurrencyId>> = vec![vec![AUSD, ACA], vec![AUSD, LDOT, ACA], vec![AUSD, DOT, ACA], vec![AUSD, RENBTC, ACA]];
+	pub DefaultFeeSwapPathList: Vec<Vec<CurrencyId>> = vec![
+		vec![AUSD, DOT, ACA],
+		vec![DOT, ACA],
+		vec![LDOT, DOT, ACA],
+		vec![RENBTC, AUSD, ACA]
```
zqhxuyuan, line 1170: should this be `[RENBTC, AUSD, DOT, ACA]`?

modules/transaction-payment/src/lib.rs
```diff
@@ -518,6 +520,7 @@ pub mod module {
 
 		/// Enable and initialize charge fee pool.
 		#[pallet::weight(<T as Config>::WeightInfo::enable_charge_fee_pool())]
+		#[transactional]
```
zqhxuyuan, line 523: there're two `#[transactional]` in this dispatch call(inner method also has transactional annotation), is that ok?

xlc: it is ok to have nested transactional but we will still want to avoid too many nesting level if possible

runtime/mandala/src/lib.rs
zjb0807: https://github.com/AcalaNetwork/Acala/blob/c34d941c09c9a3d646f013bdf0d8087dd8932ae4/runtime/mandala/src/lib.rs#L1109-L1115

Here have AUSD/ACA trading pairs

zqhxuyuan: if we already have AUSD/ACA, I think the first element be `[AUSD, ACA]` would better, isn't it? but of course `[AUSD, DOT, ACA]` works too.

zjb0807: Yeah. This is adjusting for the integration test.

https://github.com/AcalaNetwork/Acala/blob/df1daf6e747a157d8022eab358468a348b3240c3/runtime/integration-tests/src/payment.rs#L124-L141

xlc (approved):

PR #1865, opened 2022-02-10
Add new Account Tokenizer and Foreign State Oracle modules

syan095:
Added a new Account Tokenizer account. This module allows the users to convert the control of an proxy account into a NFT and back.

Added a new Foreign State Oracle module - where it interfaces with external oracles to communicate state changes/query current states on the relay chain.

Flow Chart for `request_mint` workflow: https://hackmd.io/8PrOrQfNT2mstTwHNlnUXg?view

Flow Chart for `request_redeem` workflow: https://hackmd.io/EA24goIYTAa472vognuShA?view


modules/account-tokenizer/src/lib.rs
```diff
@@ -0,0 +1,385 @@
...
+	/// The amount of fee paid when burning a Account Token, in additional to the XCM cost.
+	#[pallet::storage]
+	#[pallet::getter(fn burn_fee)]
+	type BurnFee<T: Config> = StorageValue<_, Balance, ValueQuery>;
+
+	/// Deposit locked when requesting to mint. If the mint request is successful, it is returned.
+	/// Otherwise the deposit is confiscated as penalty.
+	#[pallet::storage]
+	#[pallet::getter(fn request_mint_deposit)]
+	type RequestMintDeposit<T: Config> = StorageValue<_, Balance, ValueQuery>;
```
xlc, line 183:
```suggestion
	type MintRequestDeposit<T: Config> = StorageValue<_, Balance, ValueQuery>;
```

This should just be a constant.

modules/account-tokenizer/src/lib.rs
```diff
@@ -0,0 +1,385 @@
...
+	/// Stores mint requests.
+	/// Storage Map: Account to be tokenized  => Requester's Account Id
+	#[pallet::storage]
+	#[pallet::getter(fn mint_requests)]
+	type MintRequests<T: Config> = StorageMap<_, Twox64Concat, T::AccountId, T::AccountId, OptionQuery>;
+
+	/// The amount of fee paid when minting a Account Token
+	#[pallet::storage]
+	#[pallet::getter(fn mint_fee)]
+	type MintFee<T: Config> = StorageValue<_, Balance, ValueQuery>;
```
xlc, line 172: This should be a constant

modules/account-tokenizer/src/lib.rs
```diff
@@ -0,0 +1,385 @@
...
+	#[pallet::pallet]
+	pub struct Pallet<T>(_);
+
+	#[pallet::call]
+	impl<T: Config> Pallet<T> {
+		/// Sets the class ID of the Account Token NFT.
+		/// Only callable by authorized Oracles.
+		#[pallet::weight(0)]
+		pub fn set_nft_id(origin: OriginFor<T>, nft_id: ClassIdOf<T>) -> DispatchResult {
+			T::OracleOrigin::ensure_origin(origin)?;
```
xlc, line 194: No need to have this method

modules/account-tokenizer/src/lib.rs
```diff
@@ -0,0 +1,385 @@
...
+		#[pallet::weight(0)]
+		#[transactional]
+		pub fn request_mint(origin: OriginFor<T>, account: T::AccountId) -> DispatchResult {
+			let who = ensure_signed(origin)?;
+
+			Self::nft_class_id().ok_or(Error::<T>::NFTClassIdNotSet)?;
+
+			// An account can only have a single "requester".
+			ensure!(
+				Self::mint_requests(&account).is_none(),
```
xlc, line 241: We should not check this. Otherwise a false request will block the good one.

modules/account-tokenizer/src/lib.rs
```diff
@@ -0,0 +1,385 @@
...
+	/// Stores the NFT's class ID. Settable by authorized Oracle. Used to mint and burn PRTs.
+	#[pallet::storage]
+	#[pallet::getter(fn nft_class_id)]
+	type NFTClassId<T: Config> = StorageValue<_, ClassIdOf<T>, OptionQuery>;
+
+	/// Stores accounts that are already minted as an NFT.
+	/// Storage Map: Tokenized Account Id  => NFT Token ID
+	#[pallet::storage]
+	#[pallet::getter(fn minted_account)]
+	type MintedAccount<T: Config> = StorageMap<_, Twox64Concat, T::AccountId, T::TokenId, OptionQuery>;
```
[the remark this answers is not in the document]
syan095: This will prevent "double mint". Otherwise a user can apply to mint 2 NFTs of the same account.
This also records the token ID, so we do not need to search for it when the user is burning it.

modules/account-tokenizer/src/lib.rs
```diff
@@ -0,0 +1,385 @@
...
+		}
+
+		/// Request to mint an Account Token. Called after the user of the same Account Id has given
+		/// the proxy control of an account to the parachain account.
+		#[pallet::weight(0)]
+		#[transactional]
+		pub fn request_mint(origin: OriginFor<T>, account: T::AccountId) -> DispatchResult {
+			let who = ensure_signed(origin)?;
+
+			Self::nft_class_id().ok_or(Error::<T>::NFTClassIdNotSet)?;
```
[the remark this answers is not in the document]
syan095: NFT class is now created on runtime upgrade. StorageVersion is used to ensure the class is only created once.

modules/account-tokenizer/src/lib.rs
```diff
@@ -0,0 +1,331 @@
...
+			&requester,
+			&T::TreasuryAccount::get(),
+			T::MintRequestDeposit::get(),
+			BalanceStatus::Free,
+		)?;
+
+		Self::deposit_event(Event::MintRequestRejected { requester: requester });
+		Ok(())
+	}
+}
```
ferrell-code, line 331: I don't think we need to reject mint request as `query_task` takes a deposit
```rust 
T::ForeignStateQuery::query_task(who.clone(), call.using_encoded(|x| x.len()), call)?;
```

modules/account-tokenizer/src/lib.rs
```diff
@@ -0,0 +1,417 @@
...
+			T::Currency::transfer(&who, &T::TreasuryAccount::get(), T::MintFee::get(), KeepAlive)?;
+			T::Currency::reserve_named(&RESERVE_ID, &who, T::MintRequestDeposit::get())?;
+
+			// Submit confiramtion call to be serviced by foreign state oracle
+			let call: <T as Config>::Call = Call::<T>::confirm_mint_request {
+				owner: who.clone(),
+				account: account.clone(),
+			}
+			.into();
+			T::ForeignStateQuery::query_task(who.clone(), call.using_encoded(|x| x.len()), call)?;
```
xlc, line 271:
```suggestion
			T::ForeignStateQuery::query_task(who.clone(), call.encoded_size(), call)?;
```

modules/account-tokenizer/src/lib.rs
```diff
@@ -0,0 +1,417 @@
...
+			} else {
+				Self::accept_mint_request(owner, account)
+			}
+		}
+
+		// Requests to burn an Account Token. Sends XCM message to the relaychain to transfer the
+		// control of the account.
+		// No storage change is done, until confirmed by the Oracle.
+		#[pallet::weight(0)]
+		pub fn request_burn(origin: OriginFor<T>, account: T::AccountId, new_owner: T::AccountId) -> DispatchResult {
```
xlc, line 299:
```suggestion
		pub fn request_redeem(origin: OriginFor<T>, account: T::AccountId, new_owner: T::AccountId) -> DispatchResult {
```

modules/account-tokenizer/src/lib.rs
```diff
@@ -0,0 +1,417 @@
...
+		///
+		/// Only callable by the Oracle.
+		#[pallet::weight(0)]
+		#[transactional]
+		pub fn confirm_burn_account_token(
+			origin: OriginFor<T>,
+			account: T::AccountId,
+			new_owner: T::AccountId,
+		) -> DispatchResult {
+			T::OracleOrigin::ensure_origin(origin)?;
```
xlc, line 349: this could fail if the requester is making duplicated requests, or we should ensure duplicated requests are not possible

modules/foreign-state-oracle/src/lib.rs
```diff
@@ -0,0 +1,297 @@
...
+	// Returns pallet account
+	pub fn account_id() -> T::AccountId {
+		T::PalletId::get().into_account()
+	}
+}
+
+impl<T: Config> ForeignChainStateQuery<T::AccountId, T::VerifiableTask> for Pallet<T> {
+	#[transactional]
+	fn query_task(who: T::AccountId, length_bound: usize, dispatchable_call: T::VerifiableTask) -> DispatchResult {
+		let call_len = dispatchable_call.using_encoded(|x| x.len());
```
xlc, line 238:
```suggestion
		let call_len = dispatchable_call.encoded_size());
```

modules/foreign-state-oracle/src/lib.rs
```diff
@@ -0,0 +1,297 @@
...
+
+		ActiveQuery::<T>::insert(index, verifiable_call);
+		Self::deposit_event(Event::CreateActiveQuery { index, expiry });
+		Ok(())
+	}
+
+	#[transactional]
+	fn cancel_task(who: &T::AccountId, index: QueryIndex) -> DispatchResult {
+		let task = ActiveQuery::<T>::take(index).ok_or(Error::<T>::NoMatchingCall)?;
+		ensure!(who == &task.account_id, Error::<T>::NotQueryAccount);
```
xlc, line 267: this isn't important. we don't need permission check on this level. we will assume the caller will do it

modules/foreign-state-oracle/src/lib.rs
```diff
@@ -0,0 +1,277 @@
...
+pub type QueryIndex = u64;
+
+// Origin with arbitrary bytes included
+#[derive(PartialEq, Eq, Clone, RuntimeDebug, Encode, Decode, TypeInfo)]
+pub struct RawOrigin {
+	data: Vec<u8>,
+}
+
+#[derive(PartialEq, Eq, Clone, Encode, Decode, RuntimeDebug, TypeInfo)]
+pub struct VerifiableCall<Call, BlockNumber, Balance> {
```
xlc, line 53: Not sure if `VerifiableCall` is the right name. Where is the Verifiable coming from? I think it is more of a `RelayRequest` to request to relay some foreign state

modules/foreign-state-oracle/src/lib.rs
```diff
@@ -0,0 +1,277 @@
...
+		#[pallet::constant]
+		type QueryFee: Get<Balance>;
+
+		/// Fee for cancelling query
+		#[pallet::constant]
+		type CancelFee: Get<Balance>;
+
+		/// Timeout for query requests
+		#[pallet::constant]
+		type QueryDuration: Get<Self::BlockNumber>;
```
xlc, line 98: I don't think we want an unified constant. Different pallets may want different timeout depends on the use case.

modules/foreign-state-oracle/src/lib.rs
```diff
@@ -0,0 +1,277 @@
...
+
+		/// Dispatchable task that needs to be verified by oracle for dispatch
+		type VerifiableTask: Parameter
```
xlc, line 101: This is really a dispatchable call. Calling it Task is a bit confusing.

modules/foreign-state-oracle/src/lib.rs
```diff
@@ -0,0 +1,277 @@
...
+		/// Query that has expired is removed from storage, includes index
+		StaleQueryRemoved { query_index: QueryIndex },
+		/// Query is canceled, includes index
+		QueryCanceled { index: QueryIndex },
+	}
+
+	/// Index of Queries, each query gets unique number
+	#[pallet::storage]
+	#[pallet::getter(fn query_index)]
+	pub(super) type QueryCounter<T: Config> = StorageValue<_, QueryIndex, ValueQuery>;
```
xlc, line 136: why getter is index but storage is counter? also better call it `NextQueryId` to be consistent with other pallets

modules/foreign-state-oracle/src/lib.rs
```diff
@@ -0,0 +1,277 @@
...
+
+	/// The tasks to be dispatched with data provideed by foreign state oracle
+	///
+	/// ActiveQuery: map QueryIndex => Option<VerifiableCallOF<T>>
+	#[pallet::storage]
+	#[pallet::getter(fn active_query)]
+	pub(super) type ActiveQuery<T: Config> = StorageMap<_, Identity, QueryIndex, VerifiableCallOf<T>, OptionQuery>;
```
xlc, line 143: Storage map name usually should be plural.
We don't have inactive query so it can be just Queries. But the item is Call instead of Query, the name is not consistent.

modules/foreign-state-oracle/src/lib.rs
```diff
@@ -0,0 +1,277 @@
...
+
+	#[pallet::call]
+	impl<T: Config> Pallet<T> {
+		/// Dispatch task with arbitrary data in origin.
+		///
+		/// - `query_index`: Unique index mapped to a particular query
+		/// - `data`: Aribtrary data to be injected into the call via origin
+		#[pallet::weight(0)]
+		#[transactional]
+		pub fn dispatch_task(origin: OriginFor<T>, query_index: QueryIndex, data: Vec<u8>) -> DispatchResult {
```
xlc, line 167: The name again is a bit confusing. Dispatch a task, for a given query? And we get a query from `ActiveQuery` which is a `VerifiableCall`.

This should be relay_response (answer to a RelayRequest / Query).

modules/foreign-state-oracle/src/lib.rs
```diff
@@ -0,0 +1,277 @@
...
+			Ok(())
+		}
+
+		/// Remove Query that has expired so chain state does not bloat. This rewards the oracle
+		/// with half the query fee
+		///
+		/// - `query_index`: Unique index that is mapped to a particular query
+		#[pallet::weight(0)]
+		#[transactional]
+		pub fn remove_expired_call(origin: OriginFor<T>, query_index: QueryIndex) -> DispatchResult {
```
xlc, line 190:
```suggestion
		pub fn purge_queryl(origin: OriginFor<T>, query_index: QueryIndex) -> DispatchResult {
```

modules/foreign-state-oracle/src/lib.rs
```diff
@@ -0,0 +1,277 @@
...
+
+			let verifiable_call = ActiveQuery::<T>::take(query_index).ok_or(Error::<T>::NoMatchingCall)?;
+			// Make sure query is expired
+			ensure!(
+				verifiable_call.expiry <= T::BlockNumberProvider::current_block_number(),
+				Error::<T>::QueryNotExpired
+			);
+
+			// Gives half of reward for clearing expired query from storage
+			let reward: Balance = verifiable_call.oracle_reward.div(2u32);
```
xlc, line 201: Maybe is better to make the purge fee configurable via constant similar to cancel fee.

modules/foreign-state-oracle/src/lib.rs
```diff
@@ -0,0 +1,277 @@
...
+
+impl<T: Config> Pallet<T> {
+	// Returns pallet account
+	pub fn account_id() -> T::AccountId {
+		T::PalletId::get().into_account()
+	}
+}
+
+impl<T: Config> ForeignChainStateQuery<T::AccountId, T::VerifiableTask> for Pallet<T> {
+	#[transactional]
```
xlc, line 218:
```suggestion
	#[require_transactional]
```

modules/foreign-state-oracle/src/lib.rs
```diff
@@ -0,0 +1,277 @@
...
+	fn query_task(who: &T::AccountId, length_bound: usize, dispatchable_call: T::VerifiableTask) -> DispatchResult {
```
xlc, line 219: why the length bound is coming from caller? the call is also coming from caller so caller can just verify it.
I think it should be a constant in this pallet.
Or we could make it a `DepositPerByte` so the final deposit is `BaseFee + DepositPerByte * call_length` and then we don't need to limit the length

modules/foreign-state-oracle/src/lib.rs
```diff
@@ -0,0 +1,277 @@
...
+		type Event: From<Event<Self>> + IsType<<Self as frame_system::Config>::Event>;
+
+		/// The outer origin type.
+		type Origin: From<RawOrigin>;
+
+		/// Currency for query payments
+		type Currency: NamedReservableCurrency<
+			Self::AccountId,
+			Balance = Balance,
+			ReserveIdentifier = ReserveIdentifier,
```
xlc, line 81: where is the reserve used?

modules/account-tokenizer/src/lib.rs
```diff
@@ -0,0 +1,448 @@
...
+	/// Stores proxy accounts that are already minted as an Account Token NFT.
+	/// Storage Map: Tokenized Account Id  => NFT Token ID
+	#[pallet::storage]
+	#[pallet::getter(fn minted_account)]
+	pub type MintedAccount<T: Config> = StorageMap<_, Twox64Concat, T::AccountId, T::TokenId, OptionQuery>;
+
+	#[pallet::hooks]
+	impl<T: Config> Hooks<BlockNumberFor<T>> for Pallet<T> {
+		/// Create the NFT class once.
+		fn on_runtime_upgrade() -> Weight {
```
xlc, line 188: need to make sure the storage version is initialized correctly for a new chain, and perform the create NFT task in build genesis spec so that those state are valid for any new chain.

modules/account-tokenizer/src/lib.rs
```diff
@@ -0,0 +1,448 @@
...
+			let entropy =
+				(b"modlpy/proxy____", &who, height, ext_index, &PROXYTYPE_ANY, index).using_encoded(blake2_256);
+			let derived_account: T::AccountId = Decode::decode(&mut TrailingZeroInput::new(entropy.as_ref()))
+				.expect("infinite length input; no invalid inputs for type; qed");
+			// ensures signer also spawned anonymous proxy
+			ensure!(&account == &derived_account, Error::<T>::FailedAnonymousCheck);
+
+			// Charge the user fee and lock the deposit.
+			T::Currency::transfer(&who, &T::TreasuryAccount::get(), T::MintFee::get(), KeepAlive)?;
+			T::Currency::reserve_named(&RESERVE_ID, &who, T::MintRequestDeposit::get())?;
```
xlc, line 280: I don't think we need deposit here if `ForeignStateQuery` already takes one

modules/account-tokenizer/src/lib.rs
```diff
@@ -0,0 +1,448 @@
...
+
+			Ok(())
+		}
+	}
+}
+
+impl<T: Config> Pallet<T> {
+	/// Confirms that the Mint request is valid. Mint a NFT that represents an Account Token.
+	/// Only callable by authorized Oracles.
+	#[transactional]
```
xlc, line 406:
```suggestion
	#[require_transactional]
```

modules/account-tokenizer/src/lib.rs
syan095: Added the genesis config and build

modules/account-tokenizer/src/lib.rs
syan095: After discussion, we decided to keep the deposit in this module, due to its unique business logic

modules/account-tokenizer/src/lib.rs
syan095: The NFT token can only be burned once - repeated burn will fail the transaction. I think we are safe

modules/foreign-state-oracle/src/lib.rs
syan095: Renamed to RelayQueryRequest

modules/foreign-state-oracle/src/lib.rs
syan095: Changed to DefaultQueryDuration
When calling query_task, user can pass in an valid duration. If None is passed this default value is used.

modules/foreign-state-oracle/src/lib.rs
syan095: Renamed to DispatchableCall

modules/foreign-state-oracle/src/lib.rs
syan095: Renamed

modules/foreign-state-oracle/src/lib.rs
syan095: Renamed to QueryRequests

modules/foreign-state-oracle/src/lib.rs
syan095: Renamed to respond_query_request

modules/foreign-state-oracle/src/lib.rs
syan095: Renamed to purge_expired_query

modules/foreign-state-oracle/src/lib.rs
syan095: Added type ExpiredCallPurgeReward: Get<Permill>
currently set to 50%

modules/foreign-state-oracle/src/lib.rs
syan095: Added a max call size limit, configurable via runtime.
I think foreign state oracle is mostly called by internal modules, so call size is unclear to the end user. They would have no idea the fee size.

modules/account-tokenizer/src/lib.rs
```diff
@@ -0,0 +1,385 @@
...
+
+			// An account can only have a single "requester".
+			ensure!(
+				Self::mint_requests(&account).is_none(),
+				Error::<T>::AccountAlreadyRequestedMinted
+			);
+
+			// Ensure the account token hasn't already been minted
+			ensure!(
+				Self::minted_account(&account).is_none(),
```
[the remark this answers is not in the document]
ferrell-code: The oracle should only be used for foreign state. Native state should be checked by module

modules/account-tokenizer/src/lib.rs
```diff
@@ -0,0 +1,492 @@
...
+	/// Only callable by authorized Oracles.
+	#[require_transactional]
+	pub fn accept_mint_request(owner: T::AccountId, account: T::AccountId) -> DispatchResult {
+		let nft_class_id = Self::nft_class_id();
+
+		// Ensure the token hasn't already been minted.
+		ensure!(
+			Self::minted_account(&account).is_none(),
+			Error::<T>::AccountTokenAlreadyExists
+		);
```
ferrell-code, line 452: If fails we need to do something with reserved funds

modules/account-tokenizer/Cargo.toml
```diff
@@ -0,0 +1,55 @@
+[package]
+name = "module-account-tokenizer"
+version = "2.2.1"
```
zqhxuyuan, line 3: make it to `2.4.0` so that all modules has same version.

modules/foreign-state-oracle/Cargo.toml
```diff
@@ -0,0 +1,44 @@
+[package]
+name = "module-foreign-state-oracle"
+version = "2.3.0"
```
zqhxuyuan, line 3: make it to `2.4.0` so that all modules has same version.

modules/account-tokenizer/src/lib.rs
```diff
@@ -0,0 +1,498 @@
...
+		///
+		/// Params:
+		/// 	- `account`: Account ID of the Account Token
+		/// 	- `new_owner`: The owner of the proxy account to be transferred to.
+		#[pallet::weight(< T as Config >::WeightInfo::confirm_redeem_account_token())]
+		#[transactional]
+		pub fn confirm_redeem_account_token(
+			origin: OriginFor<T>,
+			account: T::AccountId,
+			new_owner: T::AccountId,
```
zqhxuyuan, line 418: I don't see any use of `new_owner`, do we need here?

syan095: Hi i think it's used to deposit event. I think this can be removed

modules/xcm-interface/src/lib.rs
```diff
@@ -246,4 +248,39 @@ pub mod module {
 			Self::xcm_dest_weight_and_fee(XcmInterfaceOperation::ParachainFee(Box::new(location))).1
 		}
 	}
+
+	impl<T: Config> ProxyXcm<T::AccountId> for Pallet<T> {
+		/// Cross-chain transfer staking currency to sub account on relaychain.
+		fn transfer_proxy(real: T::AccountId, new_owner: T::AccountId) -> DispatchResult {
+			let (xcm_dest_weight, xcm_fee) = Self::xcm_dest_weight_and_fee(XcmInterfaceOperation::ProxyTransferProxy);
```
zqhxuyuan, line 255: do we need some migration on runtime upgrade, or manual issue tx after runtime upgrade?

runtime/acala/src/weights/module_foreign_state_oracle.rs
```diff
@@ -0,0 +1,65 @@
...
+// along with this program. If not, see <https://www.gnu.org/licenses/>.
+
+//! Autogenerated weights for module_foreign_state_oracle
+//!
+//! THIS FILE WAS AUTO-GENERATED USING THE SUBSTRATE BENCHMARK CLI VERSION 4.0.0-dev
+//! DATE: 2022-03-13, STEPS: `50`, REPEAT: 20, LOW RANGE: `[]`, HIGH RANGE: `[]`
+//! EXECUTION: Some(Wasm), WASM-EXECUTION: Compiled, CHAIN: Some("acala-latest"), DB CACHE: 1024
+
+// Executed Command:
+// target/release/acala
```
zqhxuyuan, line 26: need re-benchmark as we're using produciton profile now.

modules/xcm-interface/src/lib.rs
ferrell-code: We could do a manual storage migration, we could also use general council to call `update_xcm_dest_weight_and_fee` when we want feature to be live

zqhxuyuan:
should be `/bench runtime module module-foreign-state-oracle`

modules/account-tokenizer/src/lib.rs
```diff
@@ -0,0 +1,498 @@
...
+pub use module::*;
+pub use weights::WeightInfo;
+
+#[frame_support::pallet]
+pub mod module {
+
+	use super::*;
+
+	#[pallet::config]
+	pub trait Config: frame_system::Config + orml_nft::Config {
```
xlc, line 79:
```suggestion
	pub trait Config: frame_system::Config {
```

Shouldn't need it as we have `NFTInterface`

modules/account-tokenizer/src/lib.rs
```diff
@@ -0,0 +1,498 @@
...
+			)
+				.using_encoded(blake2_256);
+			let derived_account: T::AccountId = Decode::decode(&mut TrailingZeroInput::new(entropy.as_ref()))
+				.expect("infinite length input; no invalid inputs for type; qed");
+			// ensures signer also spawned anonymous proxy
+			ensure!(account == derived_account, Error::<T>::FailedAnonymousProxyCheck);
+
+			// Ensure the token hasn't already been minted.
+			ensure!(
+				Self::minted_account(&account).is_none(),
```
xlc, line 306: should use `contains_key` to avoid the decode

modules/account-tokenizer/src/lib.rs
```diff
@@ -0,0 +1,498 @@
...
+
+		/// Fee for minting an Account Token NFT.
+		#[pallet::constant]
+		type MintFee: Get<Balance>;
+
+		/// The XcmInterface to communicate with the relaychain via XCM.
+		type XcmInterface: ProxyXcm<Self::AccountId>;
+
+		/// Origin used by Oracles. Used to relay information from the Relaychain.
+		type OracleOrigin: EnsureOrigin<Self::Origin, Success = Vec<u8>>;
```
xlc, line 113: this should be part of `ForeignChainStateQuery`

modules/account-tokenizer/src/lib.rs
```diff
@@ -0,0 +1,498 @@
...
+		pub fn confirm_mint_request(
+			origin: OriginFor<T>,
+			owner: T::AccountId,
+			account: T::AccountId,
+		) -> DispatchResult {
+			// Extract confirmation info from Origin.
+			let data = T::OracleOrigin::ensure_origin(origin)?;
+
+			// Checks whether oracle confirms or rejects the mint request
+			if data.get(0).ok_or(Error::<T>::BadOracleData)?.is_one() {
```
xlc, line 346:
```suggestion
			let success: bool = Decode::decode_all(data).map_err(|_| Error::<T>::InvalidQueryResponse);
			if success {
```

xlc: So we don't mix unexpected error and expected failure

modules/account-tokenizer/src/lib.rs
```diff
@@ -0,0 +1,498 @@
...
+		/// 	- `account`: Account ID of the Account Token
+		/// 	- `new_owner`: The owner of the proxy account to be transferred to.
+		#[pallet::weight(< T as Config >::WeightInfo::confirm_redeem_account_token())]
+		#[transactional]
+		pub fn confirm_redeem_account_token(
+			origin: OriginFor<T>,
+			account: T::AccountId,
+			new_owner: T::AccountId,
+		) -> DispatchResult {
+			T::OracleOrigin::ensure_origin(origin)?;
```
xlc, line 420: better to decode the response and make sure it is a `true`

modules/account-tokenizer/src/lib.rs
```diff
@@ -0,0 +1,498 @@
...
+
+			let nft_class_id = Self::nft_class_id();
+
+			// Obtain info about the account token.
+			let token_id = MintedAccount::<T>::take(&account).ok_or(Error::<T>::AccountTokenNotFound)?;
+			T::NFTInterface::owner(&nft_class_id, &token_id).ok_or(Error::<T>::AccountTokenNotFound)?;
+
+			// Find the NFT and burn it
+			T::NFTInterface::burn_from(&nft_class_id, &token_id)?;
```
xlc, line 429: burn NFT will refund the mint deposit to the current owner, which is current pallet id. we should send the feed to treasury

modules/account-tokenizer/src/lib.rs
```diff
@@ -0,0 +1,498 @@
...
+	pub fn account_id() -> T::AccountId {
+		T::PalletId::get().into_account()
+	}
+
+	/// Confirms that the Mint request is valid. Mint a NFT that represents an Account Token.
+	/// Only callable by authorized Oracles.
+	#[require_transactional]
+	pub fn accept_mint_request(owner: T::AccountId, account: T::AccountId) -> DispatchResult {
+		// Ensure the token hasn't already been minted.
+		if Self::minted_account(&account).is_some() {
```
xlc, line 453: use `contains_key` to test existence

modules/account-tokenizer/src/lib.rs
```diff
@@ -0,0 +1,498 @@
...
+			// Mint the Account Token's NFT.
+			let token_id = T::NFTInterface::next_token_id(nft_class_id);
+			T::NFTInterface::mint_into(&nft_class_id, &token_id, &owner)?;
+
+			// Create a record of the Mint and insert it into storage
+			MintedAccount::<T>::insert(account.clone(), token_id);
+
+			// Release the deposit from the owner
+			let remaining = T::Currency::unreserve_named(&RESERVE_ID, &owner, T::MintRequestDeposit::get());
+			ensure!(remaining.is_zero(), Error::<T>::InsufficientReservedBalance);
```
xlc, line 471: this doesn't need to be a hard error. make it a debug_assert

modules/account-tokenizer/src/lib.rs
```diff
@@ -0,0 +1,498 @@
...
+
+			// Pay for minting the token
+			T::NFTInterface::pay_mint_fee(&T::TreasuryAccount::get(), &Self::nft_class_id(), 1u32)?;
+
+			// Mint the Account Token's NFT.
+			let token_id = T::NFTInterface::next_token_id(nft_class_id);
+			T::NFTInterface::mint_into(&nft_class_id, &token_id, &owner)?;
+
+			// Create a record of the Mint and insert it into storage
+			MintedAccount::<T>::insert(account.clone(), token_id);
```
xlc, line 467:
```suggestion
			MintedAccount::<T>::insert(&account, token_id);
```

modules/foreign-state-oracle/src/lib.rs
```diff
@@ -0,0 +1,326 @@
...
+		#[pallet::constant]
+		type QueryFee: Get<Balance>;
+
+		/// Fee for cancelling query
+		#[pallet::constant]
+		type CancelFee: Get<Balance>;
+
+		/// Timeout for query requests
+		#[pallet::constant]
+		type DefaultQueryDuration: Get<Self::BlockNumber>;
```
xlc, line 105: I don't think we should have this. We should allow query that never expires.

runtime/karura/src/lib.rs
```diff
@@ -526,6 +530,23 @@ impl pallet_membership::Config<OperatorMembershipInstanceAcala> for Runtime {
 	type WeightInfo = ();
 }
 
+parameter_types! {
+	pub const ForeignStateMotionDuration: BlockNumber = 7 * DAYS;
+	pub const ForeignStateMaxProposals: u32 = 100;
+	pub const ForeignStateMaxMembers: u32 = 100;
```
xlc, line 536:
```suggestion
	pub const ForeignStateMotionDuration: BlockNumber = 3 * DAYS;
	pub const ForeignStateMaxProposals: u32 = 30;
	pub const ForeignStateMaxMembers: u32 = 30;
```

xlc:
We need a state diagram to ensure we have covered every single cases. I think we have some dead end currently when the query expires.

ferrell-code:
Flow Chart for `request_mint` workflow: https://hackmd.io/8PrOrQfNT2mstTwHNlnUXg?view

Currently funds that are reserved by account tokenizer will be locked indefinitely if query expires

modules/account-tokenizer/src/lib.rs
```diff
@@ -0,0 +1,498 @@
...
+		#[pallet::weight(< T as Config >::WeightInfo::request_mint())]
+		#[transactional]
+		pub fn request_mint(
+			origin: OriginFor<T>,
+			account: T::AccountId,
+			original_owner: T::AccountId,
+			height: T::BlockNumber,
+			ext_index: u32,
+			index: u16,
+		) -> DispatchResult {
```
ferrell-code, line 285: Right now anyone can `request_mint`. There needs to be a solution to not make it a race to who can mint first. My original solution was to only let the account that spawned the anon proxy call `request_mint`. I'm open to other solutions that may be a bit better to allow minting of accounts a user didn't spawn, but need some sort of auth system so only the owner of anon proxy can mint nft

xlc: what should be happening is that bad request should be slashed. there should only be one owner of the NFT and if the owner is making multiple requests, then one of it will success and rest should be slashed.

xlc:
we need to make sure we always have some escaping hutch to ensure nothing can be lost in limbo. i.e. allow governance to restore NFT etc to recover under edge cases (e.g. all oracle died, XCM message failed, network stalled for long time etc)

modules/account-tokenizer/src/lib.rs
ferrell-code: On the relay chain how do we know who owns the Anon Proxy?

ferrell-code:
Gotcha, right now AFAICT I need to add two governance calls `transfer_nft` and `unreserve_currency` to cover the limbo cases.

modules/account-tokenizer/src/lib.rs
xlc: I was about to say use the last `ProxyRemoved` event as owner and only request from this account is valid. But there is no such event yet... Let me make a PR to Substrate.
We definitely need a way to only allow the last owner of the anon proxy to be able to mint.

xlc: https://github.com/paritytech/substrate/pull/11085

ferrell-code: Still isn't ideal because then we need oracle to read historical state

xlc: Right. The other way is the owner don't remove itself from the proxy. So the anon proxy will have two owner, the parachain, and the original owner. And then we can verify that's the case and send XCM to ensure the parachain is the solo owner

modules/account-tokenizer/src/lib.rs
ferrell-code: transfers nft to treasury instead of pallet_id

modules/foreign-state-oracle/src/lib.rs
ferrell-code: None now represents query that never expires

modules/account-tokenizer/src/lib.rs
xlc: We don't need to be generic here. So we can just pul the TokenId/ClassId type from primitives and `Inspect<Self::AccountId, ClassId = ClassId, InstanceId = TokenId>`

Similar to how we handle `Balance` type in other pallets.

modules/account-tokenizer/src/lib.rs
ferrell-code: Done added event for when it is `false`. Should only return false when XCM fails

PR #1877, opened 2022-02-12
tests trader

zqhxuyuan:


xlc (approved):

xlc (changes requested):
Better to ensure this config is only used for integration tests

PR #1887, opened 2022-02-14
check swap path

### LINKED ISSUE #1875: Require input token and output token are different
For DEX, the input token and final output token should be different.

xlc:
closes #1875

### DIFF
--- a/modules/dex/src/lib.rs
+++ b/modules/dex/src/lib.rs
@@ -173,6 +173,8 @@ pub mod module {
 		StillProvisioning,
 		/// The Asset unregistered.
 		AssetUnregistered,
+		/// The trading path is invalid
+		InvalidTradingPath,
 	}
 
 	#[pallet::event]
@@ -1104,11 +1106,9 @@ impl<T: Config> Pallet<T> {
 		path: &[CurrencyId],
 		supply_amount: Balance,
 	) -> sp_std::result::Result<Vec<Balance>, DispatchError> {
+		Self::validate_path(path)?;
+
 		let path_length = path.len();
-		ensure!(
-			path_length >= 2 && path_length <= T::TradingPathLimit::get().saturated_into(),
-			Error::<T>::InvalidTradingPathLength
-		);
 		let mut target_amounts: Vec<Balance> = vec![Zero::zero(); path_length];
 		target_amounts[0] = supply_amount;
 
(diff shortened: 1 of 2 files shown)


modules/dex/src/lib.rs
```diff
@@ -1176,6 +1174,17 @@ impl<T: Config> Pallet<T> {
 		Ok(supply_amounts)
 	}
 
+	fn validate_path(path: &[CurrencyId]) -> DispatchResult {
+		let path_length = path.len();
+		ensure!(
+			path_length >= 2 && path_length <= T::TradingPathLimit::get().saturated_into(),
+			Error::<T>::InvalidTradingPathLength
+		);
+		ensure!(path[0] != path[path_length - 1], Error::<T>::InvalidTradingPath);
```
zjb0807, line 1183:
```suggestion
		ensure!(path.iter().unique().len() == path_length, Error::<T>::InvalidTradingPath);
```

and make it `public` for the `validate_path`

shaunxw: Checking unique is much worse than getting elements & compare in performance.

zjb0807: I think `kUSD, KAR, KAR, KSM` and `KUSD, KSM, KUSD, KAR` should be checked.

shaunxw, line 1183:
```suggestion
		ensure!(path.get(0) != path.get(path_length - 1), Error::<T>::InvalidTradingPath);
```

`vector[n]` could panic, I think we should use `get` instead

shaunxw: It's common to trade across two pools. I don't see the benefits of disabling them.

xlc: we checked length >= 2 so it shouldn't panic, but yeah can still improve it

zjb0807: `KUSD, KSM, KSM, KAR` will generate the tradingpair `KUSD, KSM`, `KSM, KSM`,  `KSM,KAR`. But the `KSM, KSM` is invalid.
`KUSD, KSM, KUSD, KAR` should be `KUSD, KAR`.

zjb0807 (approved):

shaunxw (approved):

PR #1891, opened 2022-02-15
Fix fee

zqhxuyuan:
previous the calcutation of payment fee of exchange rate is `old_rate * (swap_out_native_amount + lefted_amount)/old_pool_size` ≈ `old_rate * swap_amount / pool_size`.  normally if swap_amount is close to pool_size, then the result rate change not too much. buf in reality if `swap_amount` is too big or too small, then the result is either too expensive or too cheap, it will have large different with dex_exchange_rate.
given example of our current running of karura:
image

newly calculation use swap rate(dex exchange rate) as priority instead of swap_amount. if threshold/pool_size=1/2, then 
the result is `(old_rate + swap_rate)/2`. if threshold/pool_size=1/3, then the result is `old_rate/3 + swap_rate*2/3`. 
image

same as any threshold/pool_size value, we call threshold/pool_size=old_threshold_rate, then the result is `old_rate * old_threshold_rate + swap_rate * (1-old_threshold_rate)`. 
image

this PR also add migration which do the disable and enable action for the aleady exist charge fee pool.


runtime/karura/src/lib.rs
```diff
@@ -2169,8 +2169,66 @@ pub type SignedPayload = generic::SignedPayload<Call, SignedExtra>;
...
+	Block,
+	frame_system::ChainContext<Runtime>,
+	Runtime,
+	AllPalletsWithSystem,
+	TransactionPaymentMigration,
+>;
+
+parameter_types! {
+	pub FeePoolSize: Balance = 5 * dollar(KAR);
+	pub SwapBalanceThreshold: Balance = Ratio::saturating_from_rational(35, 100).saturating_mul_int(dollar(KAR));
```
zqhxuyuan, line 2183: do we need to set threshold as 2.5 KAR here this time?

modules/transaction-payment/src/tests.rs
```diff
@@ -1041,24 +1043,41 @@ fn swap_from_pool_and_dex_update_rate() {
...
+		// the swap also has it's own exchange rate by input_amount divide output_amount
+		let swap_exchange_rate = Ratio::saturating_from_rational(supply_in_amount, swap_out_native);
+		assert_eq!(swap_exchange_rate, Ratio::saturating_from_rational(80, 3074));
+
+		let old_threshold_rate = Ratio::saturating_from_rational(swap_balance_threshold, pool_size);
+		let new_threshold_rate = Ratio::one().saturating_sub(old_threshold_rate);
+
+		let new_exchange_rate = old_exchange_rate
+			.saturating_mul(old_threshold_rate)
+			.saturating_add(swap_exchange_rate.saturating_mul(new_threshold_rate));
```
shaunxw, line 1064: Number literals should be used to verify the math in uni tests. Or you're just repeating the implementation and it will always pass.

runtime/karura/src/lib.rs
```diff
@@ -2169,8 +2169,66 @@ pub type SignedPayload = generic::SignedPayload<Call, SignedExtra>;
...
+}
+
+pub struct TransactionPaymentMigration;
+impl frame_support::traits::OnRuntimeUpgrade for TransactionPaymentMigration {
+	fn on_runtime_upgrade() -> frame_support::weights::Weight {
+		let _ = module_transaction_payment::Pallet::<Runtime>::disable_charge_fee_pool(
```
shaunxw, line 2189: The implementation details should be in `transaction-payment` pallet.

modules/transaction-payment/src/tests.rs
shaunxw: Also need to add more `swap_amount` cases to cover different scenarios.

runtime/karura/src/lib.rs
xlc: I think this is fine. It is a one time karura only migration.

runtime/karura/src/lib.rs
xlc: yeah make it 2.5 KAR

modules/transaction-payment/src/tests.rs
zqhxuyuan: I use `calculate_exchange_rate` method in pallet, and use it in testcase, this will be better now.

shaunxw: No, this is still just repeating the impl, not proper unit tests.

USE NUMBER LITERALS.

shaunxw: Let me add more details:

A proper unit tests to verify math should be like

```rust
// impl
fn add_one(num: u32) -> u32 {
  num + 1
}

// tests
#[test]
fn add_one_works() {
  // literal 2 is used
  assert_eq!(add_one(1), 2);
}
```

What you're doing in unit tests currently is similar to:

```rust
// impl
fn add_one(num: u32) -> u32 {
  do_add_one(num)
}

fn do_add_one(num: u32) -> u32 {
  num + 1
}

// tests
#[test]
fn add_one_works() {
  // literal 2 is used
  assert_eq!(add_one(1), do_add_one(1));
}
```

zqhxuyuan: 👍 done NUMBER LITERALS check in the event.

xlc (approved):

shaunxw (approved):

PR #1893, opened 2022-02-16
Storage Migration: HomaXcm to XcmInterface

syan095:
Added migration of storage from the previously named HomaXcm to the newly named XcmInterface, for runtime Acala, Karura and Mandala

### DIFF
--- a/runtime/karura/src/lib.rs
+++ b/runtime/karura/src/lib.rs
@@ -2664,8 +2664,8 @@ mod tests {
 	#[test]
 	fn check_call_size() {
 		assert!(
-			core::mem::size_of::<Call>() <= 280,
-			"size of Call is more than 280 bytes: some calls have too big arguments, use Box to \
+			core::mem::size_of::<Call>() <= 260,
+			"size of Call is more than 260 bytes: some calls have too big arguments, use Box to \
 			reduce the size of Call.
 			If the limit is too strong, maybe consider increasing the limit",
 		);


modules/xcm-interface/src/migrations.rs
```diff
@@ -0,0 +1,46 @@
...
+		let new_prefix = "XcmInterface";
+
+		log::info!(
+			target: "runtime::xcm-interface",
+			"Running migration from HomaXcm to XcmInterface. Old prefix: {:?}, New prefix: {:?}",
+			old_prefix, new_prefix,
+		);
+
+		frame_support::storage::migration::move_pallet(old_prefix.as_bytes(), new_prefix.as_bytes());
+
```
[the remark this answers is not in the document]
syan095: Yes good idea.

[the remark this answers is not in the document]
syan095: Updated i will test it now.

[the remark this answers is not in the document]
syan095: Tested and confirmed the update of storage version works.

ferrell-code (dismissed):
lgtm

runtime/karura/src/lib.rs
```diff
@@ -2613,8 +2628,8 @@ mod tests {
 	#[test]
 	fn check_call_size() {
 		assert!(
-			core::mem::size_of::<Call>() <= 260,
-			"size of Call is more than 260 bytes: some calls have too big arguments, use Box to \
+			core::mem::size_of::<Call>() <= 280,
```
xlc, line 2631: why increase this?

syan095: The new call size is 272, so I increased it to 280.
Feels like this change is from Masters, I will confirm.

syan095: Yes can confirm on Master branch, call size is 272

xlc: but CI passed on master?

syan095: yeah very strange. I'll revert and test using CI. I will also test on my windows machine. Might be build environment issue?

xlc: could be. wasm is 32bits and native is 64bits so size of some data structure will be different

syan095: i think the call size is OK on my windows. If this passes in CI (which master passes) it must be only an mac m1 chip thing. 
I'll keep this reverted anyway, and keep the change local so tests will pass locally.

syan095: Can confirm on my PC's Ubunto, the call size is 256.

xlc (approved):

PR #1897, opened 2022-02-16
add polkadot launch script

zqhxuyuan:


scripts/acala-bifrost.json
```diff
@@ -0,0 +1,106 @@
+{
```
xlc, line 1: move those to scripts/polkadot-launch/

.gitignore
```diff
@@ -8,6 +8,9 @@
 .idea
 *.wasm
 *.state
+# exclude file generate by polkadot-launch
+*.log
+*.json
```
xlc, line 13: I don't think we want to exclude all the json files

xlc (approved):

PR #1898, opened 2022-02-17
update build script

xlc:


zjb0807 (commented):
LGTM.
```
evm-bench
evm-tests
ts-tests
```
This needed to be added in `.dockerignore`

Makefile
```diff
@@ -32,19 +32,19 @@ build-all:
 
 .PHONY: build-release
 build-release:
-	CARGO_PROFILE_RELEASE_LTO=true RUSTFLAGS="-C codegen-units=1" cargo build --locked --features with-all-runtime --release --workspace --exclude runtime-integration-tests --exclude e2e-tests
+	CARGO_PROFILE_RELEASE_LTO=true RUSTFLAGS="-C codegen-units=1" cargo build --locked --features with-all-runtime --release --workspace --exclude runtime-integration-tests --exclude e2e-tests --exclude test-service
```
zqhxuyuan, line 35: test-service is the only child package inside e2e-tests, I guess we could only exclude test-service, but I think its fine to also exlude e2e-tests, as we may be add other e2e test case later on.

xlc: I noticed that test-service was also compiled if we don't exclude it.

zqhxuyuan: yeah, test-service needs excluded.

zjb0807 (approved):

PR #1908, opened 2022-02-19
add DexLiquidity Proxy

### LINKED ISSUE #1902: DexLiquidity proxy
Add new ProxyType::DexLiquidity that's able to add liquidity and withdraw liquidity

ferrell-code:
closes #1902

### DIFF
--- a/runtime/common/src/lib.rs
+++ b/runtime/common/src/lib.rs
@@ -319,9 +319,9 @@ pub enum ProxyType {
 	CancelProxy,
 	Governance,
 	Auction,
-	DexLiquidity,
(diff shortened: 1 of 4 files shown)


runtime/common/src/lib.rs
```diff
@@ -319,6 +319,7 @@ pub enum ProxyType {
 	CancelProxy,
 	Governance,
 	Auction,
+	DexLiquidity,
```
xlc, line 322: this needs to be the last variant to avoid breaking changes on the encoding format

xlc (approved):

PR #1911, opened 2022-02-20
remove old migration

xlc:


runtime/karura/src/lib.rs
```diff
@@ -2188,7 +2188,7 @@ pub type Executive = frame_executive::Executive<
 	frame_system::ChainContext<Runtime>,
 	Runtime,
 	AllPalletsWithSystem,
-	(TransactionPaymentMigration, XcmInterfaceMigrationV1),
+	XcmInterfaceMigrationV1,
```
zqhxuyuan, line 2191: XcmInterfaceMigrationV1 not applied?

xlc: Not yet

xlc: https://github.com/AcalaNetwork/Acala/pull/1893

zqhxuyuan:
LGTM.
For some note, the origin is wrong in TransactionPaymentMigration(should be root), so the migration is taken by manual council dispatch call and validated ok:
https://karura.subsquare.io/council/motion/218
https://karura.subsquare.io/council/motion/219

PR #1925, opened 2022-02-27
fix evm-bench access_list

ermalkaleci:


xlc (approved):
Is this not checked in CI?

ermalkaleci:
> Is this not checked in CI?

Yeah we don't run it. Will include it on benchmark tests

xlc (approved):

PR #1933, opened 2022-03-02
fix clap annotation

zqhxuyuan:


launch/config.yml
```diff
@@ -1,5 +1,5 @@
 relaychain:
-  image: parity/polkadot:v0.9.16
+  image: parity/polkadot:latest
```
xlc, line 2: better to use a specific version to avoid unexpected breaking change

xlc (approved):

PR #1935, opened 2022-03-03
Fix calculation

wangjj9219:
debit_value / exchange_rate = debit_balance,   but debit_balance * exchange_rate may be 1 smaller than debit_value. So need fix calculation to resolve possible repaid debit failure.


xlc (approved):
Some test?

PR #1941, opened 2022-03-04
rearrange precompiles

ermalkaleci:


runtime/common/src/precompile/mod.rs
```diff
@@ -69,11 +69,11 @@ pub const SHA3_256: H160 = H160(hex!("0000000000000000000000000000000000000081")
 pub const SHA3_512: H160 = H160(hex!("0000000000000000000000000000000000000082"));
 
 pub const MULTI_CURRENCY: H160 = H160(hex!("0000000000000000000000000000000000000400"));
-pub const NFT: H160 = H160(hex!("0000000000000000000000000000000000000401"));
-pub const EVM: H160 = H160(hex!("0000000000000000000000000000000000000402"));
-pub const ORACLE: H160 = H160(hex!("0000000000000000000000000000000000000403"));
-pub const SCHEDULER: H160 = H160(hex!("0000000000000000000000000000000000000404"));
-pub const DEX: H160 = H160(hex!("0000000000000000000000000000000000000405"));
+pub const EVM: H160 = H160(hex!("0000000000000000000000000000000000000401"));
```
zjb0807, line 72: It was used here
https://github.com/AcalaNetwork/predeploy-contracts/blob/master/contracts/evm/EVM.sol#L8

ermalkaleci: this is not actually necessary

Issue #1912, opened 2022-02-21
Module Transaction Payment refactor


xlc:
The current custom fee handling requires integrate all the known support pairs to try to find a supported pair to charge tx fee. This is fine if we only have limited supported fee token but will be too slow when number of supported asset grows.

- module-transaction-payment
  - Types
  - Config
    - CustomFeeSurplus: Get\<Percent>
      - 50% for Karura & Acala
    - AlternativeFeeSurplus: Get\<Percent>
      - 25% for Karura & Acala
    - DefaultFeeTokens: Get<Vec\<CurrencyId>>
      - KSM, kUSD, LKSM for Karura, DOT, aUSD, LCDOT, LDOT for Acala
  - Calls
    - fn with_fee_path(origin, path: Vec\<CurrencyId>, call: Call)
      - dispatch call
      - no additional logic within this call
    - fn with_fee_currency(origin, currency: CurrencyId, call: Call)
      - dispatch call
      - no additional logic within this call
  - ChargeTransactionPayment
    - match call
      - with_fee_path:
      - with_fee_currency:
        - use the path / currency tx pool for fee payment
        - CustomFeeSurplus additional tx fee
      - others
        - try native token
          - no additional tx fee
        - try AlternativeFeeSwapPath
          - AlternativeFeeSurplus additional tx fee
        - try DefaultFeeTokens
          - AlternativeFeeSurplus additional tx fee

All tokens that can be converted into native token will be supported as fee token without any additional registration.
Frontend needs to be updated to detect a custom fee token is required and use `with_fee_path` or `with_fee_currency` to wrap calls.
Discounted fee when paying with native token.
Slightly discounted fee when user select a default fee path (that's not native).

zjb0807:
If the user uses `with_fee_currency ` with `ERC20` to pay for the tx fee, how to get the swap path?

xlc:
`with_fee_path` can be used to specify a path directly. `with_fee_currency` requires there to be a tx fee pool first

zjb0807:
> `with_fee_path` can be used to specify a path directly. `with_fee_currency` requires there to be a tx fee pool first

https://github.com/AcalaNetwork/Acala/blob/01c72e8dadcb0af65a55750941b26760ee41ccbc/modules/transaction-payment/src/lib.rs#L616-L624

`enable_charge_fee_pool` needs to provide the path,  the same as the register path.

```
DefaultFeeTokens: Get<Vec<CurrencyId>>
[...]

PR #1943, opened 2022-03-08
Fee discount refactor

zqhxuyuan:
closes: https://github.com/AcalaNetwork/Acala/issues/1912

- [x] direct dex swap:
    - `with_fee_path(Vec<CurrencyId>)`
    - `alternative_fee_swap_path(who) -> Vec<CurrencyId>`
- [x] exchange from tx fee pool(make sure enable_charge_fee_pool first):
    - `with_fee_currency(CurrencyId)`
    - ~~`default fee swap path`~~ -> `DefaultFeeTokens`
- [x] fix weight and benchmark for `with_fee_path/with_fee_currency`
- [x] remove `set_global_fee_swap_path`, add `path` arg to `enable_charge_fee_pool`
- [x] remove `DefaultFeeSwapPathList`, add `DefaultFeeTokens`
- [x] `GlobalFeeSwapPath` stored both `DefaultFeeTokens` related trading path(`DefaultFeeSwapPathList`) and ERC20 path 
    - [x] need migration(karura including ForeignAsset(0) indicate RMRK)
- [x] remove `get_trading_path` and related test
- [x] remove `get_trading_path_by_currency`
- [x] post_dispatch should also consider surplus fund to `OnTransactionPayment::on_unbalanceds()` hook

post_dispatch fee calculation:
image


modules/transaction-payment/src/lib.rs
```diff
@@ -514,6 +532,58 @@ pub mod module {
...
+		#[pallet::weight({
+			let dispatch_info = call.get_dispatch_info();
+			(T::WeightInfo::with_fee_path().saturating_add(dispatch_info.weight), dispatch_info.class,)
+		})]
+		pub fn with_fee_path(
+			origin: OriginFor<T>,
+			fee_swap_path: Vec<CurrencyId>,
+			call: Box<CallOf<T>>,
+		) -> DispatchResult {
+			ensure_signed(origin.clone())?;
```
zqhxuyuan, line 546: should we consider other origin, i.e. root or council account that not issued by user tx?

modules/transaction-payment/src/lib.rs
```diff
@@ -825,31 +821,102 @@ where
...
+			let fee_surplus = T::AlternativeFeeSurplus::get().mul_ceil(fee);
+			let fee_amount = fee_surplus.saturating_add(amount);
+
+			// alter native fee swap path, swap from dex: O(1)
+			if let Some(path) = AlternativeFeeSwapPath::<T>::get(who) {
+				let supply_currency_id = *path.first().expect("should match a non native asset");
+				let supply_balance = T::MultiCurrency::free_balance(supply_currency_id, who);
+				if !supply_balance.is_zero()
+					&& T::DEX::swap_with_specific_path(who, &path, SwapLimit::ExactTarget(supply_balance, fee_amount))
+						.is_ok()
```
zqhxuyuan, line 889: the user setting of `AlternativeFeeSwapPath` maybe invalid, i.e. RMRK->KAR. 
User also may setting `AlternativeFeeSwapPath` that exist in `DefaultFeeTokens` but also invalid when swap, i.e. KUSD->KAR. 
current implements just verify swap operation success or not, if failed, then go to other cae. we may need extra  check on `set_alternative_fee_swap_path` so that user provide trading path is valid, and when swap with this path is valid.

modules/transaction-payment/src/lib.rs
```diff
@@ -813,7 +808,8 @@ where
 		T::WeightToFee::calc(&capped_weight)
 	}
 
-	pub fn ensure_can_charge_fee(who: &T::AccountId, fee: PalletBalanceOf<T>, reason: WithdrawReasons) {
+	/// If native asset is enough, return `None`, else return the fee should be withdrawn.
+	fn native_is_enough(who: &T::AccountId, fee: PalletBalanceOf<T>, reason: WithdrawReasons) -> Option<Balance> {
```
xlc, line 812:
```suggestion
	fn check_native_balance(who: &T::AccountId, fee: PalletBalanceOf<T>, reason: WithdrawReasons) -> Option<Balance> {
```

modules/transaction-payment/src/lib.rs
```diff
@@ -825,31 +821,101 @@ where
 					<T as Config>::Currency::ensure_can_withdraw(who, fee, reason, new_free_balance).is_ok()
 				});
 		if native_is_enough {
-			return;
+			None
+		} else {
+			Some(fee.saturating_add(native_existential_deposit.saturating_sub(total_native)))
```
xlc, line 826:
```suggestion
			Some(fee.saturating_add(native_existential_deposit))
```

Otherwise I think it will have some issue like when user have a lot of locked native token. Then the amount will become zero.

modules/transaction-payment/src/lib.rs
```diff
@@ -825,31 +821,101 @@ where
 					<T as Config>::Currency::ensure_can_withdraw(who, fee, reason, new_free_balance).is_ok()
```
xlc, line 821: why do we need total balance and free balance check and `ensure_can_withdraw`?
I think `ensure_can_withdraw` by itself is enough?

modules/transaction-payment/src/lib.rs
```diff
@@ -825,31 +821,101 @@ where
...
+		who: &T::AccountId,
+		fee: PalletBalanceOf<T>,
+		reason: WithdrawReasons,
+		call: &CallOf<T>,
+	) -> Result<Balance, DispatchError> {
+		let custom_fee_surplus = T::CustomFeeSurplus::get().mul_ceil(fee);
+		let custom_fee_amount = fee.saturating_add(custom_fee_surplus);
+		let fee_surplus: Result<Balance, DispatchError> = match call.is_sub_type() {
+			Some(Call::with_fee_path { fee_swap_path, .. }) => {
+				let supply_currency_id = *fee_swap_path.first().expect("should match a non native asset");
```
xlc, line 844: need to validate `fee_swap_path` here and the expect here is not right

modules/transaction-payment/src/lib.rs
```diff
@@ -825,31 +821,101 @@ where
...
+				T::DEX::swap_with_specific_path(
+					who,
+					fee_swap_path,
+					SwapLimit::ExactTarget(
+						T::MultiCurrency::free_balance(supply_currency_id, who),
```
xlc, line 849:
```suggestion
						Balance::MAX,
```

It will fail anyway if user does not have enough funds

modules/transaction-payment/src/lib.rs
```diff
@@ -825,31 +821,101 @@ where
...
+			Some(_) | None => Self::native_then_alternative_or_default(who, fee, reason),
+		};
+		fee_surplus
+	}
 
-		// make sure add extra gap to keep alive after swap.
-		let amount = fee.saturating_add(native_existential_deposit.saturating_sub(total_native));
-		// native is not enough, try swap native from fee pool to pay fee and gap.
-		Self::swap_native_asset(who, amount);
+	fn ensure_can_charge_fee(who: &T::AccountId, fee: PalletBalanceOf<T>, reason: WithdrawReasons) {
```
xlc, line 863: is this used?

modules/transaction-payment/src/lib.rs
```diff
@@ -825,31 +821,101 @@ where
...
+	/// `AlternativeFeeSwapPath`, this will direct swap from dex. Sometimes, user setting of
+	/// `AlternativeFeeSwapPath` may be wrong or dex is not available, or user do not set any
+	/// `AlternativeFeeSwapPath`, then use the `DefaultFeeTokens` to swap from tx fee pool.
+	fn native_then_alternative_or_default(
+		who: &T::AccountId,
+		fee: PalletBalanceOf<T>,
+		reason: WithdrawReasons,
+	) -> Result<Balance, DispatchError> {
+		// native asset is not enough
+		if let Some(amount) = Self::native_is_enough(who, fee, reason) {
```
xlc, line 878:
```suggestion
		if let Some(amount) = Self::native_is_enough(who, fee, reason) {
				// native asset is not enough
```

modules/transaction-payment/src/lib.rs
```diff
@@ -825,31 +821,101 @@ where
...
+			let fee_surplus = T::AlternativeFeeSurplus::get().mul_ceil(fee);
+			let fee_amount = fee_surplus.saturating_add(amount);
+
+			// alter native fee swap path, swap from dex: O(1)
+			if let Some(path) = AlternativeFeeSwapPath::<T>::get(who) {
+				let supply_currency_id = *path.first().expect("should match a non native asset");
```
xlc, line 884: this shouldn't fail but still better to avoid expect here and just return a DispatchError

modules/transaction-payment/src/lib.rs
```diff
@@ -825,31 +821,101 @@ where
...
+				let supply_balance = T::MultiCurrency::free_balance(supply_currency_id, who);
```
xlc, line 885: I don't think we need to read free balance. just make the supply amount MAX

modules/transaction-payment/src/lib.rs
xlc: yeah we should validate the path is valid in `set_alternative_fee_swap_path` but a valid path may become invalid if for example a pool is removed. although that's not something we need to worry about here. people can always set the path to something else, just that in that case they have to have some native token to ensure they have fee to call the set fee swap path call

modules/transaction-payment/src/lib.rs
```diff
@@ -825,31 +821,101 @@ where
...
+			// default fee tokens, swap from tx fee pool: O(1)
+			for supply_currency_id in T::DefaultFeeTokens::get() {
+				if !T::MultiCurrency::free_balance(supply_currency_id, who).is_zero()
+					&& Self::swap_from_pool_or_dex(who, fee_amount, supply_currency_id).is_ok()
+				{
+					return Ok(fee_surplus);
+				}
+			}
+
+			// migration of `GlobalFeeSwapPath`. after Dapp using `with_fee_path`, we can delete this.
```
xlc, line 903: write an issue for this

modules/transaction-payment/src/lib.rs
```diff
@@ -514,6 +532,58 @@ pub mod module {
...
+			let dispatch_info = call.get_dispatch_info();
+			(T::WeightInfo::with_fee_path().saturating_add(dispatch_info.weight), dispatch_info.class,)
+		})]
+		pub fn with_fee_path(
+			origin: OriginFor<T>,
+			fee_swap_path: Vec<CurrencyId>,
+			call: Box<CallOf<T>>,
+		) -> DispatchResult {
+			ensure_signed(origin.clone())?;
+			ensure!(
```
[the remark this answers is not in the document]
zqhxuyuan: move to `ensure_can_charge_fee_with_call`, but this leave unused parameter begin with `_`.

modules/transaction-payment/src/lib.rs
zqhxuyuan: `reserve_fee` and `charge_fee` in `TransactionPayment` trait used it without `call` parameter.
- `reserve_fee` is used by evm schedule precompile.
- `charge_fee` is used by renvm-bridge(as it's not used, mabye need cleanup too)

modules/transaction-payment/src/lib.rs
zqhxuyuan: https://github.com/AcalaNetwork/Acala/issues/1952

modules/transaction-payment/src/lib.rs
zqhxuyuan: delete `ensure_can_charge_fee` invoke in `reserve_fee`, does this effect evm schedule precompile? @zjb0807

modules/transaction-payment/src/lib.rs
zqhxuyuan: `ensure_can_withdraw` refer to substrate frame `Balances`: https://github.com/paritytech/substrate/blob/f5f286db0da99761792b69bf4a997535dcc8f260/frame/balances/src/lib.rs#L1473-L1484

I think the simpliest way is just compare `fee + ed` with `free`. if fee+ed>free, native is not enough, result=`fee+ed-free`.  The reason we need `-free` is for newly user, first tx, native is not enough, result=`fee+ed-0`, after withdraw fee, free=`ed`.  second tx, native is still not enough, result=`fee+ed-free`=fee.

modules/transaction-payment/src/lib.rs
zqhxuyuan: current result is fee+(ed-total), even though total(reserve balance) is too large, ed-total=0 thanks to saturating_sub, so the result amount would't be zero instead of fee.  if change to (fee+ed)-total then it's possible getting 0 if total have large reserve balance.

modules/transaction-payment/src/lib.rs
```diff
@@ -813,43 +758,98 @@ where
...
+				if let Some(supply_currency_id) = path.first() {
+					if Self::swap_from_pool_or_dex(who, fee_amount, *supply_currency_id).is_ok() {
+						return Ok(fee_surplus);
+					}
 				}
 			}
 		}
+
+		// native asset is enough
+		Ok(0)
```
ferrell-code, line 852: I think this should be in an else block to cover the edge case where native is not enough and none of the other options work. Right now it would return Ok(()) but I think it should return an error in that scenario

modules/transaction-payment/src/lib.rs
```diff
@@ -1173,7 +1163,7 @@ impl<T: Config + Send + Sync> sp_std::fmt::Debug for ChargeTransactionPayment<T>
 
 impl<T: Config + Send + Sync> ChargeTransactionPayment<T>
 where
-	<T as frame_system::Config>::Call: Dispatchable<Info = DispatchInfo, PostInfo = PostDispatchInfo>,
+	CallOf<T>: Dispatchable<Info = DispatchInfo, PostInfo = PostDispatchInfo>,
 	PalletBalanceOf<T>: Send + Sync + FixedPointOperand,
```
ferrell-code, line 1167: still more ;)

modules/transaction-payment/src/lib.rs
```diff
@@ -1294,17 +1292,18 @@ where
 impl<T: Config + Send + Sync> SignedExtension for ChargeTransactionPayment<T>
 where
 	PalletBalanceOf<T>: Send + Sync + From<u64> + FixedPointOperand,
-	<T as frame_system::Config>::Call: Dispatchable<Info = DispatchInfo, PostInfo = PostDispatchInfo>,
+	CallOf<T>: Dispatchable<Info = DispatchInfo, PostInfo = PostDispatchInfo>,
```
ferrell-code, line 1295: and here too

modules/transaction-payment/src/lib.rs
zqhxuyuan: in the case of native is not enough, even return Ok(0), when withdraw_asset, it'll throw error. for fail-fast, return Error now.

ferrell-code (dismissed):

runtime/acala/src/lib.rs
```diff
@@ -1675,8 +1675,38 @@ pub type SignedPayload = generic::SignedPayload<Call, SignedExtra>;
...
+	pub SwapBalanceThreshold: Balance = Ratio::saturating_from_rational(25, 10).saturating_mul_int(dollar(ACA));
+}
+
+impl OnRuntimeUpgrade for TransactionPaymentMigration {
+	fn on_runtime_upgrade() -> frame_support::weights::Weight {
+		let poo_size = FeePoolSize::get();
+		let threshold = SwapBalanceThreshold::get();
+		let tokens = vec![
+			(AUSD, vec![AUSD, ACA]),
+			(LDOT, vec![LDOT, AUSD, ACA]),
```
xlc, line 1700:
```suggestion
			(LCDOT, vec![LCDOT, AUSD, ACA]),
```

modules/transaction-payment/src/lib.rs
```diff
@@ -1359,6 +1358,10 @@ where
 				refund = refund_fee.saturating_add(refund_tip);
 				actual_tip = tip.saturating_sub(refund_tip);
 			}
+			// fee surplus not included in refund
+			if let Some(surplus) = surplus {
+				refund = refund.saturating_sub(surplus);
```
xlc, line 1363: we still want to do a partial refund or surplus similar how we refund tip

modules/transaction-payment/src/lib.rs
```diff
@@ -1202,11 +1200,12 @@ where
...
-		Pallet::<T>::ensure_can_charge_fee(who, fee, reason);
+		let fee_surplus =
+			Pallet::<T>::ensure_can_charge_fee_with_call(who, fee, call).map_err(|_| InvalidTransaction::Payment)?;
 
-		// withdraw native currency as fee
-		match <T as Config>::Currency::withdraw(who, fee, reason, ExistenceRequirement::KeepAlive) {
-			Ok(imbalance) => Ok((fee, Some(imbalance))),
+		// withdraw native currency as fee, also consider surplus when swap from dex or pool.
+		match <T as Config>::Currency::withdraw(who, fee + fee_surplus, reason, ExistenceRequirement::KeepAlive) {
+			Ok(imbalance) => Ok((fee + fee_surplus, Some(imbalance), Some(fee_surplus))),
```
xlc, line 1208: return be returning `None` when `fee_surplus` is zero?

modules/transaction-payment/src/lib.rs
zqhxuyuan: actual_surplus=actual_fee * percent, refund_surplus=surplus-actual_surplus, should also passing `percent` as `Pre`.

xlc (approved):

modules/transaction-payment/src/lib.rs
zqhxuyuan: no need to return `None` as it's only used in post_dispatch.

xlc (approved):

PR #1946, opened 2022-03-09
precompile gas calculation

ermalkaleci:


xlc (approved):

runtime/common/src/precompile/mock.rs
```diff
@@ -572,13 +574,72 @@ frame_support::construct_runtime!(
...
+				balance: 1000 * crate::dollar(ACA),
+				storage: BTreeMap::new(),
+				code: vec![],
+				enable_contract_development: true,
+			};
+			accounts.insert(dev_acc, account);
+		}
+
+		accounts
+	}
```
zjb0807, line 641: why not use `use acala_service::chain_spec::mandala::evm_genesis;` here?

ermalkaleci: @zjb0807 dependency loop

ermalkaleci: > Maybe can move `evm_genesis` into `acala_primitives`

agree

ermalkaleci: runtime-common actually

zjb0807 (approved):

PR #1958, opened 2022-03-14
Abstract bonding logic

xlc:
abstract bonding logic so it can be reused by other pallets later

### DIFF
--- a/modules/nominees-election/src/lib.rs
+++ b/modules/nominees-election/src/lib.rs
@@ -158,14 +158,14 @@ pub mod module {
 		pub fn bond(origin: OriginFor<T>, #[pallet::compact] amount: Balance) -> DispatchResult {
 			let who = ensure_signed(origin)?;
 
-			let mut ledger = Self::ledger(&who);
+			let ledger = Self::ledger(&who);
 			let free_balance = T::Currency::free_balance(&who);
 			if let Some(extra) = free_balance.checked_sub(ledger.total()) {
 				let extra = extra.min(amount);
 
 				let old_active = ledger.active();
 
-				ledger.bond(extra).map_err(Into::<Error<T, I>>::into)?;
+				let ledger = ledger.bond(extra).map_err(Into::<Error<T, I>>::into)?;
 
 				let old_nominations = Self::nominations(&who);
 
@@ -180,13 +180,13 @@ pub mod module {
 		pub fn unbond(origin: OriginFor<T>, #[pallet::compact] amount: Balance) -> DispatchResult {
 			let who = ensure_signed(origin)?;
 
-			let mut ledger = Self::ledger(&who);
+			let ledger = Self::ledger(&who);
 
 			let old_active = ledger.active();
 
 			// Note: in case there is no current era it is fine to bond one era more.
 			let era = Self::current_era() + T::BondingDuration::get();
-			let amount = ledger.unbond(amount, era).map_err(Into::<Error<T, I>>::into)?;
+			let (ledger, amount) = ledger.unbond(amount, era).map_err(Into::<Error<T, I>>::into)?;
 
 			if !amount.is_zero() {
 				let old_nominations = Self::nominations(&who);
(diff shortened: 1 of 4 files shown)


primitives/src/bonding.rs
```diff
@@ -0,0 +1,429 @@
...
+	MinBondThreshold: Get<Balance>,
+{
+	pub fn new() -> Self {
+		Self {
+			unlocking: Default::default(),
+			total: Default::default(),
+			active: Default::default(),
+			_phantom: Default::default(),
+		}
+	}
```
shaunxw, line 79: I think `new` impl can be moved into `default`, and we always use `default`.

modules/nominees-election/src/lib.rs
```diff
@@ -232,20 +158,18 @@ pub mod module {
...
-				ledger.active += extra;
-				ensure!(
-					ledger.active >= T::MinBondThreshold::get(),
-					Error::<T, I>::BelowMinBondThreshold
-				);
-				ledger.total += extra;
+
+				let old_active = ledger.active();
+
+				let ledger = ledger.bond(extra).map_err(Into::<Error<T, I>>::into)?;
```
shaunxw, line 168: IMO we should avoid variable shadowing which is bug prone. `new_ledge` might be better.

modules/nominees-election/src/lib.rs
```diff
@@ -256,30 +180,21 @@ pub mod module {
...
 
-			let mut ledger = Self::ledger(&who);
+			let ledger = Self::ledger(&who);
 
-			let amount = amount.min(ledger.active);
+			let old_active = ledger.active();
+
+			// Note: in case there is no current era it is fine to bond one era more.
+			let era = Self::current_era() + T::BondingDuration::get();
+			let (ledger, amount) = ledger.unbond(amount, era).map_err(Into::<Error<T, I>>::into)?;
```
shaunxw, line 189: Same here.

modules/nominees-election/src/lib.rs
```diff
@@ -288,35 +203,35 @@ pub mod module {
...
 			let ledger = Self::ledger(&who);
-			ensure!(!ledger.unlocking.is_empty(), Error::<T, I>::NoUnlockChunk);
-			let old_active = ledger.active;
-			let old_ledger_unlocking = ledger.unlocking.len();
+
+			let old_active = ledger.active();
+			let old_ledger_unlocking = ledger.unlocking_len();
 			let old_nominations = Self::nominations(&who);
-			let ledger = ledger.rebond(amount);
+			let (ledger, amount) = ledger.rebond(amount).map_err(Into::<Error<T, I>>::into)?;
```
shaunxw, line 210: and here

modules/nominees-election/src/lib.rs
```diff
@@ -288,35 +203,35 @@ pub mod module {
...
 		#[pallet::weight(T::WeightInfo::withdraw_unbonded(T::MaxUnlockingChunks::get()))]
 		#[transactional]
 		pub fn withdraw_unbonded(origin: OriginFor<T>) -> DispatchResultWithPostInfo {
 			let who = ensure_signed(origin)?;
-			let mut ledger = Self::ledger(&who);
-			let old_ledger_unlocking = ledger.unlocking.len();
-			ledger.consolidate_unlocked(Self::current_era());
+			let ledger = Self::ledger(&who);
+			let old_ledger_unlocking = ledger.unlocking_len();
+			let ledger = ledger.consolidate_unlocked(Self::current_era());
```
shaunxw, line 225: and here

shaunxw (commented):
LGTM, some nitpicks

modules/nominees-election/src/lib.rs
xlc: It is actually better to do shadowing than `ledger1`, `ledger2`, `ledger3` which you will eventually lost track.
It basically implements mutation without mutation.

PR #1978, opened 2022-03-18
improve dex-support-tai-pool-asset

zjb0807:
1. Update comment
https://github.com/AcalaNetwork/Acala/blob/2e454f3891e486f55456368f89d96b2511da1df0/primitives/src/evm.rs#L132-L133
2. Update the check in the dex module
https://github.com/AcalaNetwork/Acala/blob/2e454f3891e486f55456368f89d96b2511da1df0/modules/dex/src/lib.rs#L520-L528
3. StableAssetPoolId is `u32`, so needs 4 bytes.
https://github.com/AcalaNetwork/Acala/blob/2e454f3891e486f55456368f89d96b2511da1df0/primitives/src/currency.rs#L336-L338

### DIFF
--- a/primitives/src/currency.rs
+++ b/primitives/src/currency.rs
@@ -225,9 +225,9 @@ pub type Lease = BlockNumber;
 pub enum DexShare {
 	Token(TokenSymbol),
 	Erc20(EvmAddress),
-	StableAssetPoolToken(StableAssetPoolId),
 	LiquidCrowdloan(Lease),
 	ForeignAsset(ForeignAssetId),
+	StableAssetPoolToken(StableAssetPoolId),
 }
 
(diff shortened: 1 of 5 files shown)


primitives/src/currency.rs
```diff
@@ -225,9 +225,9 @@ pub type Lease = BlockNumber;
 pub enum DexShare {
 	Token(TokenSymbol),
 	Erc20(EvmAddress),
+	StableAssetPoolToken(StableAssetPoolId),
```
xlc, line 228: wouldn't this cause issue with incentives pallet? I think we have some storages storing `CurrencyId::DexShare`

xlc (approved):

PR #1995, opened 2022-03-25
connect earning and incentives

xlc:
Will only enable it for Mandala for now.
Also updated the `OnUpdateLoan` to use add/remove share instead of set share. Need to verify if this is ok.

TODOs:

- [x] tests


wangjj9219 (dismissed):

ferrell-code (approved):
lgtm

shaunxw (approved):

PR #2000, opened 2022-03-29
Update construct_runtime's syntax

### LINKED ISSUE #1652: Switch to latest construct_runtime! syntax
https://github.com/paritytech/polkadot/pull/4389

syan095:
Changed construct_runtime's syntax to the new version's for Runtimes
Increased the recursion limit to 512.

Closes #1652


runtime/mandala/src/lib.rs
```diff
@@ -1849,122 +1849,122 @@ construct_runtime! {
...
-		AcalaOracle: orml_oracle::<Instance1>::{Pallet, Storage, Call, Event<T>} = 80,
-		OperatorMembershipAcala: pallet_membership::<Instance5>::{Pallet, Call, Storage, Event<T>, Config<T>} = 82,
+		AcalaOracle: orml_oracle::<Instance1> = 80,
+		OperatorMembershipAcala: pallet_membership::<Instance5> = 82,
 
 		// ORML Core
-		Auction: orml_auction::{Pallet, Storage, Call, Event<T>} = 100,
-		Rewards: orml_rewards::{Pallet, Storage, Call} = 101,
-		OrmlNFT: orml_nft::{Pallet, Storage, Config<T>} = 102,
+		Auction: orml_auction  = 100,
```
xlc, line 1900: looks like we have some duplicated spaces here

xlc (commented):
can you verify if the metadata is still the same

runtime/mandala/src/lib.rs
syan095: Checked and fixed spacing errors.

runtime/acala/src/lib.rs
```diff
@@ -1538,103 +1538,103 @@ construct_runtime!(
...
+		Preimage: pallet_preimage = 8,
 
 		// Tokens & Related
-		Balances: pallet_balances::{Pallet, Call, Storage, Config<T>, Event<T>} = 10,
-		Tokens: orml_tokens::{Pallet, Storage, Event<T>, Config<T>} = 11,
-		Currencies: module_currencies::{Pallet, Call, Event<T>} = 12,
-		Vesting: orml_vesting::{Pallet, Storage, Call, Event<T>, Config<T>} = 13,
-		TransactionPayment: module_transaction_payment::{Pallet, Call, Storage, Event<T>} = 14,
+		Balances: pallet_balances = 10,
+		Tokens: orml_tokens = 11,
```
xlc, line 1553: need to exclude the `Call`

syan095: Compared metadata of mandala, karura and acala to ensure no extra Calls are added.

xlc:
should use exclude_parts

https://github.com/paritytech/substrate/blob/master/frame/support/procedural/src/lib.rs#L322

syan095:
Screen Shot 2022-03-30 at 22 04 34

syan095:
Compared the metadata of mandala, karura and acala.
only difference is XTokens, which had Storage decleared in runtime, but no storage defined in module. The new syntax will remove the empty storage metadata. (see screenshot attached above)

shaunxw (commented):
Why the recursion limit is increased to 512? The macro syntax change shouldn't have any performance impact.

syan095:
> Why the recursion limit is increased to 512? The macro syntax change shouldn't have any performance impact.

When I build locally the recursion increase to around 380-ish. I thought we usually keep the max recursion a multiple of 2 so I changed it to 512.

xlc:
Let's merge after we cut new release for Karura & Acala.

xlc (approved):

PR #2007, opened 2022-03-30
update precompile gas_limit

ermalkaleci:
precompiles consumes 90% of gas_limit in case of revert


runtime/common/src/precompile/mod.rs
```diff
@@ -76,6 +76,10 @@ pub const ORACLE: H160 = H160(hex!("0000000000000000000000000000000000000403"));
 pub const SCHEDULER: H160 = H160(hex!("0000000000000000000000000000000000000404"));
 pub const DEX: H160 = H160(hex!("0000000000000000000000000000000000000405"));
 
+pub fn target_gas_limit(target_gas: Option<u64>) -> Option<u64> {
+	target_gas.map(|x| ((x as u128) * 9 / 10) as u64)
```
zjb0807, line 80:
```suggestion
	target_gas.map(|x| x. saturating_div(10u64).saturating_mul(9) // 90%
```

zjb0807 (approved):
LGTM

PR #2011, opened 2022-03-31
remove static bound on Pallet

zqhxuyuan:
- nomineses-election


zqhxuyuan:
this module don't used anymore.

PR #2013, opened 2022-04-01
Fix srtool

zjb0807:


xlc (approved):

xlc:
why our CI didn't catch this? it should be using the same version of Rust as srtool?

zjb0807:
We are using 1.58 to support rust 2021. 
The `srtool` is not upgraded so the polkadot is not supporting rust 2021
```
rustc 1.58.0-nightly (072799443 2021-11-06)
```

We should push the srtool upgrade

PR #2030, opened 2022-04-08
Implement fungible traits for Module-currencies

syan095:
Implemented relevent traits from fungibles for module-currencies


syan095:
Continuation of the work from 
https://github.com/AcalaNetwork/Acala/pull/1363

modules/currencies/src/lib.rs
```diff
@@ -637,12 +657,332 @@ impl<T: Config> MultiReservableCurrency<T::AccountId> for Pallet<T> {
...
+		amount: Self::Balance,
+		best_effort: bool,
+	) -> Result<Self::Balance, DispatchError> {
+		match asset_id {
+			CurrencyId::Erc20(_) => {
+				if amount.is_zero() {
+					return Ok(amount);
+				}
+				ensure!(
+					amount <= <Self as MultiReservableCurrency<_>>::reserved_balance(asset_id, who) || best_effort,
```
shaunxw, line 926:
```suggestion
					best_effort || amount <= <Self as MultiReservableCurrency<_>>::reserved_balance(asset_id, who),
```

do `best_effort`check first, to reduce one database read if it's true

modules/currencies/src/lib.rs
```diff
@@ -637,12 +657,332 @@ impl<T: Config> MultiReservableCurrency<T::AccountId> for Pallet<T> {
...
+		on_hold: bool,
+	) -> Result<Self::Balance, DispatchError> {
+		match asset_id {
+			CurrencyId::Erc20(_) => {
+				if amount.is_zero() {
+					return Ok(amount);
+				}
+
+				let reserved_balance = <Self as fungibles::InspectHold<_>>::balance_on_hold(asset_id, source);
+				ensure!(amount <= reserved_balance || best_effort, Error::<T>::BalanceTooLow);
```
shaunxw, line 954:
```suggestion
				ensure!(best_effort || amount <= <Self as fungibles::InspectHold<_>>::balance_on_hold(asset_id, source), Error::<T>::BalanceTooLow);
```

Same here.

modules/currencies/src/lib.rs
```diff
@@ -637,12 +657,332 @@ impl<T: Config> MultiReservableCurrency<T::AccountId> for Pallet<T> {
...
+				{
+					return DepositConsequence::Overflow;
+				}
+
+				if <Self as fungibles::Inspect<_>>::balance(asset_id, who)
+					.checked_add(&amount)
+					.is_none()
+				{
+					return DepositConsequence::Overflow;
+				}
```
shaunxw, line 718: If adding to total issuance can't overflow, there is no chance this can.

modules/currencies/src/lib.rs
```diff
@@ -637,12 +657,332 @@ impl<T: Config> MultiReservableCurrency<T::AccountId> for Pallet<T> {
...
+
+				if <Self as fungibles::Inspect<_>>::balance(asset_id, who)
+					.checked_sub(&amount)
+					.is_none()
+				{
+					return WithdrawConsequence::NoFunds;
+				}
+
+				WithdrawConsequence::Success
+			}
```
shaunxw, line 755: Looks duplicated with `ensure_can_withdraw` to me.

modules/currencies/src/lib.rs
```diff
@@ -637,12 +657,332 @@ impl<T: Config> MultiReservableCurrency<T::AccountId> for Pallet<T> {
...
+					InvokeContext {
+						contract,
+						sender,
+						origin: origin_address,
+					},
+					address,
+					amount,
+				)?;
+				Ok(amount)
+			}
```
shaunxw, line 838: duplicated with `MultiCurrency` transfer impl https://github.com/AcalaNetwork/Acala/blob/80ba6122a8adb3ecff33154e0b3415ad55162db4/modules/currencies/src/lib.rs#L339-L353

syan095: Looks like there's quite a few duplicates, I'm working through to clean it up

modules/currencies/src/lib.rs
```diff
@@ -637,12 +657,274 @@ impl<T: Config> MultiReservableCurrency<T::AccountId> for Pallet<T> {
...
+		amount: Self::Balance,
+		keep_alive: bool,
+	) -> Result<Self::Balance, DispatchError> {
+		if amount.is_zero() || source == dest {
+			return Ok(amount);
+		}
+
+		let transferred_amount = match asset_id {
+			CurrencyId::Erc20(_) => {
+				<Self as MultiCurrency<_>>::transfer(asset_id, source, dest, amount)?;
```
shaunxw, line 782: This will end up with depositing `Transferred` event twice.

shaunxw: I think the `transfer` impl can be delegated to `<Self as MultiCurrency<_>>::transfer`. They are the same except return types.

syan095: Yeah i had a look - it's the keepalive parameter that's making it difficult. I wanted to keep this so the keepalive can be passed into the trait bounded implementation.

syan095: I'll move the deposit event to the inside of the Erc20 match branch so we don't double deposit as you said

shaunxw: Then you will need to deposit events in the last match branch, `orml-tokens`'s `MultiCurrency` impl has no events deposit.

syan095: Addressed here.
Also tidied up events in moduke-currencies and unit tests
https://github.com/open-web3-stack/open-runtime-module-library/pull/729

wangjj9219 (changes requested):
No need to add new events now, because in the future we plan to remove all events in module-currencies, like orml-currencies. However, before the exchange switches to orml-tokens as the deposit basis, as long as the Transferred event is still triggered correctly in the module-currencies.

syan095:
> No need to add new events now, because in the future we plan to remove all events in module-currencies, like orml-currencies. However, before the exchange switches to orml-tokens as the deposit basis, as long as the Transferred event is still triggered correctly in the module-currencies.

Then should i remove all events except for transfers? (only transfers in evm accounts will emit module-currencies's event

wangjj9219:
> > No need to add new events now, because in the future we plan to remove all events in module-currencies, like orml-currencies. However, before the exchange switches to orml-tokens as the deposit basis, as long as the Transferred event is still triggered correctly in the module-currencies.
> 
> Then should i remove all events except for transfers? (only transfers in evm accounts will emit module-currencies's event

I didn't notice that the new events are only triggered for Erc20, so keep it

wangjj9219 (approved):

modules/currencies/src/lib.rs
```diff
@@ -797,17 +797,37 @@ impl<T: Config> fungibles::Transfer<T::AccountId> for Pallet<T> {
 			return Ok(amount);
 		}
 
-		let transferred_amount = match asset_id {
+		Ok(match asset_id {
```
xlc, line 800: make it 

```
let amount = match ...
Ok(amount)
```

just a bit easier to read

syan095:
Added 

> we still want to emit Currencies.Transferred event for all transfers to avoid breaking changes

Ensured that both fn transfer, in Multicurrency and in fungibles::Transfer will emit events in all cases (but no duplicate events)

modules/currencies/src/lib.rs
```diff
@@ -369,46 +388,42 @@ impl<T: Config> MultiCurrency<T::AccountId> for Pallet<T> {
...
 			id if id == T::GetNativeCurrencyId::get() => T::NativeCurrency::can_slash(who, amount),
 			_ => T::MultiCurrency::can_slash(currency_id, who, amount),
 		}
 	}
 
 	fn slash(currency_id: Self::CurrencyId, who: &T::AccountId, amount: Self::Balance) -> Self::Balance {
 		match currency_id {
-			CurrencyId::Erc20(_) => Default::default(),
+			CurrencyId::Erc20(_) => {
+				Self::deposit_event(Event::Slashed {
```
xlc, line 420: this should be `Erc20InvalidOperation`?

modules/currencies/src/lib.rs
```diff
@@ -482,7 +497,14 @@ impl<T: Config> MultiReservableCurrency<T::AccountId> for Pallet<T> {
 
 	fn slash_reserved(currency_id: Self::CurrencyId, who: &T::AccountId, value: Self::Balance) -> Self::Balance {
 		match currency_id {
-			CurrencyId::Erc20(_) => value,
+			CurrencyId::Erc20(_) => {
+				Self::deposit_event(Event::ReserveSlashed {
```
xlc, line 501: this is no slash happening, so shouldn't emit event

xlc (changes requested):
I don't think we should be emitting additional events for ERC20 operations. They are not useful at all.

modules/currencies/src/lib.rs
syan095: The function signature require return of "Balance" - so error cannot be returned

modules/currencies/src/lib.rs
```diff
@@ -231,11 +207,6 @@ pub mod module {
 				}
 				if free_balance < Self::minimum_balance(currency_id) {
 					T::OnDust::on_dust(&account, currency_id, free_balance);
-					Self::deposit_event(Event::DustSwept {
```
xlc, line 234: keep this one

xlc (commented):
@wangjj9219 please confirm the removed events are not currently used by our services

shaunxw (dismissed):

Issue #1846, opened 2022-02-04
Keep evm nonce alive


zjb0807:
https://github.com/AcalaNetwork/Acala/blob/faf97a0647336512dbb8abfb1f0ce47e13ce6f71/modules/evm/src/lib.rs#L1791-L1798

Need more reason not to reset evm nonce.

zjb0807:
#1844

zjb0807:
https://github.com/AcalaNetwork/Acala/blob/e9a0f70efd341911bc6ccedd84411c1f9ab1d890/modules/evm/src/runner/state.rs#L706-L712

The nonce of the caller will affect the contract address. After reset, there will be conflicts in the deployment contract.

zjb0807:
https://acala-testnet.subscan.io/extrinsic/786715-2

In this case, the `claim_account` causes the default account to be killed, the nonce to be reset, and the deployment of the new contract fails.

Issue #1668, opened 2021-12-03
Dusting Accounts with mapped Evm Address 


ferrell-code:
https://github.com/AcalaNetwork/Acala/blob/b2054d585665478422e9d14c0c8620dc832f708d/modules/evm-accounts/src/lib.rs#L310

Currently the weights are 2 writes underweight for when an account has a claimed an evm address and gets reaped... we could change the logic or update the benchmarks

Lawrence-Acala:
Hi @ferrell-code is it still required? I'm trying to pick up what is these aged issues, can you please elaborate a bit?

ferrell-code:
Yeah there are benchmarks where the worst case scenario is an account is killed. But those benchmarks don't account for the account being killed having an evm mapping that will be removed too. This means that a few of our benchmarks are slightly undercharging the worst case scenario... It is a minor issue but still valid

xlc:
This won't be a blocking issue for the EVM deployment.

Issue #1613, opened 2021-11-16
module-evm-accounts should offer a way for evm account to dispatch call with default substrate account


xlc:
After account claiming, the default account of an evm address will become inaccessible which could be problematic.
Should have a way to dispatch call with the default substrate account so user can recover funds if needed

PR #2032, opened 2022-04-11
Keep evm nonce alive

zjb0807:
Closes: #1846 
Closes: #1668
Closes: #1613

https://github.com/AcalaNetwork/Acala/blob/05ae43770cf8cc747a2aad1c2891046789da51c3/modules/evm-accounts/src/lib.rs#L312-L322

In `get_or_create_evm_address`, it will generate the map between the account and the default eth address. So it can't claim_account for another eth address.
If there is a bug in the code and the map isn't inserted, we can kill the account and then `claim_default_account` to recover the map.

### DIFF
--- a/modules/evm-accounts/src/lib.rs
+++ b/modules/evm-accounts/src/lib.rs
@@ -337,11 +337,7 @@ where
 pub struct CallKillAccount<T>(PhantomData<T>);
 impl<T: Config> OnKilledAccount<T::AccountId> for CallKillAccount<T> {
 	fn on_killed_account(who: &T::AccountId) {
-		// remove the reserve mapping that could be created by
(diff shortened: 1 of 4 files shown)


ermalkaleci (approved):

ferrell-code (approved):

shaunxw (dismissed):

shaunxw (commented):
IMO `remove_account` should be made not public, to encourage other modules using `remove_account_if_empty` for safety. (not to misuse) https://github.com/AcalaNetwork/Acala/blob/59b5ba2fba50fee150e0eff39448a6398f4f619c/modules/evm/src/lib.rs#L1248

Issue #2039, opened 2022-04-13
use ChargeTransactionPayment to charge the storage fee in evm


zjb0807:
In the current implementation, the storage fee requires native tokens.
If erc20 is used to pay the tx fee, the transaction payment will only be exchanged for the gas fee, which will cause the storage fee to fail.

PR #2045, opened 2022-04-19
charge storage fee to support non native asset

zqhxuyuan:
closes: https://github.com/AcalaNetwork/Acala/issues/2039

change evm reserve_storage from `T::Currency::reserve_named()` to `T::ChargeTransactionPayment::reserve_fee()`, and we can add our own fee strategy in transaction payment module's `reserve_fee()`. That is, `reserve_fee` now can support none native token.

https://github.com/AcalaNetwork/Acala/blob/7c16ceca094f47d8dea49c000f1c0f169c20b641/modules/evm/src/lib.rs#L1609

we do not need to change [`unreserve_storage`](https://github.com/AcalaNetwork/Acala/blob/7c16ceca094f47d8dea49c000f1c0f169c20b641/modules/evm/src/lib.rs#L1630) and [`charge_storage`](https://github.com/AcalaNetwork/Acala/blob/7c16ceca094f47d8dea49c000f1c0f169c20b641/modules/evm/src/lib.rs#L1656-L1668) in evm, because `reserve_storage` happend before those two operation. That is, `unreserve_storage` and `charge_storage` always have native token.

---

If we have this feature, I think ~~bodhi.js~~ Metamask needs support this too, currently if an eth account not bind to substrate account, it have enough none-native token, but not enough native token, the transfer can't proceed
image


xlc:
It is ok if Metamask can't handle it as that's basically the same behaviour of Ethereum. This however will be useful for polkadot.js extension user or dApp using EIP712 signature format.

modules/transaction-payment/src/lib.rs
```diff
@@ -1391,21 +1406,32 @@ impl<T: Config + Send + Sync> TransactionPayment<T::AccountId, PalletBalanceOf<T
...
 {
-	fn reserve_fee(who: &T::AccountId, weight: Weight) -> Result<PalletBalanceOf<T>, DispatchError> {
-		let fee = Pallet::<T>::weight_to_fee(weight);
-		<T as Config>::Currency::reserve_named(&RESERVE_ID, who, fee)?;
+	fn reserve_fee(
+		who: &T::AccountId,
+		fee: PalletBalanceOf<T>,
+		named: Option<ReserveIdentifier>,
+	) -> Result<PalletBalanceOf<T>, DispatchError> {
+		let fee = Pallet::<T>::calculate_final_fee(fee);
```
xlc, line 1414: if the method is passing fee directly, then we shouldn't modify it with multiplier otherwise it will be confusing
also the storage deposit is not subject to multiplier, only the weight fee is

zqhxuyuan: add a `fee_multiplier` flag as parameter, for storage fee, passing it as false, while other case may set it to true.

modules/transaction-payment/src/lib.rs
```diff
@@ -1391,21 +1400,32 @@ impl<T: Config + Send + Sync> TransactionPayment<T::AccountId, PalletBalanceOf<T
...
+		fee: PalletBalanceOf<T>,
+		named: Option<ReserveIdentifier>,
+	) -> PalletBalanceOf<T> {
+		let fee = Pallet::<T>::calculate_final_fee(fee, None);
+		<T as Config>::Currency::unreserve_named(&named.unwrap_or(RESERVE_ID), who, fee)
 	}
 
 	fn unreserve_and_charge_fee(
 		who: &T::AccountId,
 		weight: Weight,
```
xlc, line 1425: this should be passing Balance as well? we will want to apply multiplier here

modules/transaction-payment/src/lib.rs
```diff
@@ -1425,6 +1445,7 @@ where
 		payed: NegativeImbalanceOf<T>,
 	) -> Result<(), TransactionValidityError> {
 		let refund = Pallet::<T>::weight_to_fee(refund_weight);
+		let refund = Pallet::<T>::calculate_final_fee(refund, None);
```
xlc, line 1448: this is noop when multiplier is None. I think we want to pass refund amount instead of weight

modules/transaction-payment/src/lib.rs
zqhxuyuan: add a `Multiplier` parameter to `calculate_final_fee` of Pallet, invoker should invoke this if they want to apply multiplier. Notice, if None is provide, then it will use `next_fee_multiplier` from onchain storage.

modules/transaction-payment/src/lib.rs
zqhxuyuan: still use weight, as `unreserve_and_charge_fee` is only used by module evm scheduled call. if change to Balance, then module evm either depend on payment module, or add an etra `weight_to_fee` method to trait which seems no benefit.

modules/transaction-payment/src/lib.rs
zqhxuyuan: remove it now, leave to invoker to decide whether invoke `apply_multiplier_to_fee`/`calculate_final_fee` or not.

scripts/polkadot-launch/acala-launch.json
```diff
@@ -68,4 +68,4 @@
   ],
   "types": {},
   "finalization": false
-}
+}
```
xlc, line 71: unintended change

modules/transaction-payment/src/lib.rs
```diff
@@ -751,13 +751,21 @@ where
...
 		// cap the weight to the maximum defined in runtime, otherwise it will be the
 		// `Bounded` maximum of its data type, which is not desired.
 		let capped_weight = weight.min(T::BlockWeights::get().max_block);
 		T::WeightToFee::calc(&capped_weight)
 	}
 
+	/// Apply multiplier to fee, return the final fee. If multiplier is `None`, use
+	/// `next_fee_multiplier`.
+	#[allow(dead_code)]
+	fn apply_multiplier_to_fee(fee: PalletBalanceOf<T>, multiplier: Option<Multiplier>) -> PalletBalanceOf<T> {
```
xlc, line 764: this and weight_to_fee should be exposed via a trait otherwise other pallets won't be able to use it

xlc (dismissed):

modules/evm/src/bench/mock.rs
```diff
@@ -188,7 +187,7 @@ impl Config for Runtime {
 	type PrecompilesValue = ();
 	type ChainId = ConstU64<1>;
 	type GasToWeight = GasToWeight;
-	type ChargeTransactionPayment = DefaultTransactionPayment;
+	type ChargeTransactionPayment = DefaultTransactionPayment<Balances>;
```
ermalkaleci, line 190: We don't need to implement this. It's not used

modules/evm/src/lib.rs
```diff
@@ -1606,7 +1606,8 @@ impl<T: Config> Pallet<T> {
 			caller, user, limit, amount
 		);
 
-		T::Currency::reserve_named(&RESERVE_ID_STORAGE_DEPOSIT, &user, amount)
+		T::ChargeTransactionPayment::reserve_fee(&user, amount, Some(RESERVE_ID_STORAGE_DEPOSIT))?;
```
ermalkaleci, line 1609: don't we need to do the same for unreserve?

zqhxuyuan: As I comment in the beginning:
> we do not need to change [unreserve_storage](https://github.com/AcalaNetwork/Acala/blob/7c16ceca094f47d8dea49c000f1c0f169c20b641/modules/evm/src/lib.rs#L1630) and charge_storage in evm, because reserve_storage happend before those two operation. That is, unreserve_storage and [charge_storage](https://github.com/AcalaNetwork/Acala/blob/7c16ceca094f47d8dea49c000f1c0f169c20b641/modules/evm/src/lib.rs#L1656-L1668) always have native token.

In other words, If user don't have enough native token, and `T::ChargeTransactionPayment::reserve_fee` success, it makes user got enough reserved native token. then `unreserve` based on native token will always success. of course, the amount of `unreserve` should not large then `reserve`, but that's invoker's duty.

modules/evm/src/bench/mock.rs
zqhxuyuan: any reason why? maybe move evm bench to runtime/mandala/benchmarking will make things(I mean use relay impl which depend on transaction payment module) a lot easier?

ermalkaleci: we don't have to maintain mirror this implementation

modules/evm/src/lib.rs
ermalkaleci: it's going to be the same but for consistency `reserve_fee` and `un reserve_fee`

modules/evm/src/bench/mock.rs
zqhxuyuan: Add module_dex, and use it's real implementation.
do we need to re-benchmark?

xlc (approved):

xlc (approved):

PR #2050, opened 2022-04-21
Update Honzon and Cdp Treasury Benchmarks

ferrell-code:
Since `AlternativeSwapPathJointList` is a parameter set in runtime and not user controlled data we really can just update benchmarks to ensure each path works. 

Downside is every time a new path is added to `AlternateSwapPathJoinList` we will need to update these benchmarks accordingly.

Relates to #2044, this solution seems simpler, but the other one would not need to be rebenchmarked every time we add a path to `AlternateSwapPathJoinList`


ferrell-code:
stale

PR #2051, opened 2022-04-21
fix rpc gas limit

ermalkaleci:
* GasLimit and StorageLimit based on block limits
* Disable RPC estimate mode for more accurate gas usage calculation
* Add special methods for RPC calls
* RPC methods to query `evm_blockLimits`


modules/evm/rpc/src/lib.rs
```diff
@@ -128,9 +128,9 @@ fn to_u128(val: NumberOrHex) -> std::result::Result<u128, ()> {
 }
 
 // 20M. TODO: use value from runtime
```
xlc, line 130:
```suggestion
// 33M. TODO: use value from runtime
```

ermalkaleci: @xlc still working. I'm thinking about moving the limit to runtime

modules/evm/rpc/src/lib.rs
```diff
@@ -241,14 +257,15 @@ where
 			.map_err(|err| internal_err(format!("execution fatal: {:?}", err)))?;
 
 		// Determine the highest possible gas limits
-		let max_gas_limit = MAX_GAS_LIMIT;
-		let mut highest = U256::from(request.gas_limit.unwrap_or(max_gas_limit));
+		let max_gas_limit = request.max_gas_limit;
+		let max_storage_limit = request.max_storage_limit;
+		let mut highest = std::cmp::max(request.gas_limit, max_gas_limit);
```
zjb0807, line 262: This will cause each transaction to be estimated from the maximum gas_limit, increasing the number of executions

modules/evm/rpc/src/lib.rs
```diff
@@ -387,22 +404,21 @@ where
 		// rpc_binary_search_estimate block
 		{
 			// Define the lower bound of the binary search
-			const MIN_GAS_PER_TX: U256 = U256([21_000, 0, 0, 0]);
-			let mut lowest = MIN_GAS_PER_TX;
+			let mut lowest = BASE_CALL_GAS;
```
zjb0807, line 407: image
The minimum value should not be changed. This affects the lower limit of gas_limit

modules/evm/rpc/src/lib.rs
```diff
@@ -185,7 +201,7 @@ where
 						gas_limit,
 						storage_limit,
 						access_list,
-						true,
+						false,
```
zjb0807, line 204: make sure the tests work.
https://github.com/AcalaNetwork/bodhi.js/tree/master/examples/waffle

modules/evm/rpc/src/lib.rs
ermalkaleci: @zjb0807 this is supposed to be `cmp::min`, fixing it

modules/evm/rpc/src/lib.rs
ermalkaleci: @zjb0807 we can leave that 21k, no problem

modules/evm/rpc/src/lib.rs
```diff
@@ -241,14 +257,15 @@ where
 			.map_err(|err| internal_err(format!("execution fatal: {:?}", err)))?;
 
 		// Determine the highest possible gas limits
-		let max_gas_limit = MAX_GAS_LIMIT;
-		let mut highest = U256::from(request.gas_limit.unwrap_or(max_gas_limit));
+		let max_gas_limit = request.max_gas_limit;
+		let max_storage_limit = request.max_storage_limit;
+		let mut highest = std::cmp::min(request.gas_limit, max_gas_limit);
```
zjb0807, line 262: `request.max_gas_limit` and `request.max_storage_limit` fields can be removed.

runtime/common/src/lib.rs
```diff
@@ -395,6 +395,23 @@ where
...
+		let normal_weight = weights.get(DispatchClass::Normal);
+		WeightToGas::convert(normal_weight.max_extrinsic.unwrap_or(weights.max_block))
+	}
+
+	pub fn max_storage_limit() -> u32 {
+		let length = T::BlockLength::get();
+		*length.max.get(DispatchClass::Normal)
+	}
+}
+
```
zjb0807, line 414: Can you add integration test to get the `max_gas_limit` and `max_storage_limit` value?

modules/evm/rpc/src/lib.rs
ermalkaleci: @zjb0807 why? I'm using them. I'm also thinking about exposing them to RPC. It may be useful

ermalkaleci: You're right I don't need `max_storage_limit` but I still need `max_gas_limit`, it's been used below

modules/evm/rpc/src/lib.rs
ermalkaleci: @zjb0807 extended Runner with special methods for RPC. StorageRent will be disabled with `rpc_mode` set true

xlc (commented):
The max gas limit for eth_call is usually like 10x of the block gas limit. It is higher to allow people to make some calls that could be resource consuming but not intended to be called within transaction

modules/evm/src/runner/stack.rs
```diff
@@ -346,6 +351,77 @@ impl<T: Config> RunnerT<T> for Runner<T> {
...
+		value: BalanceOf<T>,
+		gas_limit: u64,
+		storage_limit: u32,
+		access_list: Vec<(H160, Vec<H256>)>,
+		config: &evm::Config,
+	) -> Result<CallInfo, DispatchError> {
+		// if the contract not published, the caller must be developer or contract or maintainer.
+		// if the contract not exists, let evm try to execute it and handle the error.
+		ensure!(
+			Pallet::<T>::can_call_contract(&target, &source),
```
xlc, line 371: we shouldn't need this check

runtime/acala/src/lib.rs
```diff
@@ -1820,27 +1828,33 @@ impl_runtime_apis! {
 			gas_limit: u64,
 			storage_limit: u32,
 			access_list: Option<Vec<AccessListItem>>,
-			estimate: bool,
+			rpc_mode: bool,
```
xlc, line 1831: is this backward compatible?

modules/evm/rpc/runtime_api/src/lib.rs
```diff
@@ -49,7 +51,7 @@ sp_api::decl_runtime_apis! {
 			gas_limit: u64,
 			storage_limit: u32,
 			access_list: Option<Vec<AccessListItem>>,
-			estimate: bool,
+			rpc_mode: bool,
```
xlc, line 54: the name is a bit confusing. what exactly does rpc_mode do? also it is false from estimate resources rpc so it is more confusing

ermalkaleci: @xlc this needs to be true only when call coming from RPC `evm_call` so it will skip storage rent. For estimate it's ok to charge storage because execution of tx you're estimating will be the same when you submit it.

ermalkaleci: @xlc I don't have a better name. Maybe `rpc_call` is better?

modules/evm/src/runner/stack.rs
ermalkaleci: @xlc agree

runtime/acala/src/lib.rs
ermalkaleci: what do you mean?

modules/evm/rpc/runtime_api/src/lib.rs
xlc: `view_mode`?

runtime/acala/src/lib.rs
xlc: this is runtime API, so two thing will happen:
client upgraded before runtime upgrade, so new client + old runtime
runtime upgraded before client upgrade, so old client + new runtime

in this case the runtime API remains the same so at lease it won't be crashing, but does the behaviour still remains reasonable when runtime and client are on different version?

modules/evm/rpc/runtime_api/src/lib.rs
ermalkaleci: @xlc it's not just view. Maybe `rpc_call` is better. Indicating that the call is coming from rpc

runtime/acala/src/lib.rs
ermalkaleci: @xlc this will require new client + new runtime

modules/evm/rpc/runtime_api/src/lib.rs
xlc: eth_call is view mode.
my point is, why disable storage charge when the call is coming from rpc? that's the part isn't explained just by reading the code.

ermalkaleci: @xlc what about `ignore_storage: bool`?

runtime/acala/src/lib.rs
xlc: we are not going to be able to upgrade both runtime & client at same time, so we need to have some versioning

but in this particular case, will it be any incompatibility with old client + new runtime? the function signature is same and the semantic is still somewhat compatible?

xlc (approved):

zjb0807 (approved):

Issue #1999, opened 2022-03-29
Ability to transfer debit between loans of different collateral type


xlc:
Add new method to Honzon pallet to allow people transfer debit between loans.

PR #2053, opened 2022-04-25
Ability to transfer debits between loans

ferrell-code:
closes: #1999


modules/honzon/src/lib.rs
```diff
@@ -307,6 +313,33 @@ pub mod module {
...
+		pub fn transfer_debit(
+			origin: OriginFor<T>,
+			from_currency: CurrencyId,
+			to_currency: CurrencyId,
+			debit_transfer: Balance,
+		) -> DispatchResult {
+			let who = ensure_signed(origin)?;
+			let debit_amount: Amount = debit_transfer.try_into().map_err(|_| ArithmeticError::Overflow)?;
+			let negative_debit = debit_amount.checked_neg().ok_or(ArithmeticError::Overflow)?;
+			<cdp_engine::Pallet<T>>::adjust_position(&who, from_currency, Zero::zero(), negative_debit)?;
```
xlc, line 333: I think this requires user have enough aUSD on hand to close the debit?
We really want is user can still do this without having any free aUSD. Should have a test for this.

xlc (dismissed):

xlc (dismissed):

xlc (approved):

PR #2066, opened 2022-05-02
Added minimal collateral amount for cdp-engine module

### LINKED ISSUE #1956: Minimal Collateral amount
Add a MinimalCollateralAmount conifg
When debit value is zero, we ensure the collateral amount is gte MinimalCollateralAmount

Make it to be 100x of the ED in runtime config.

syan095:
Added new Config for cdp-engine : minimum collateral amount
Added new unit tests
Configured the collateral amount to be 100x ED in runtimes

Closes #1956

### DIFF
--- a/modules/cdp-engine/src/lib.rs
+++ b/modules/cdp-engine/src/lib.rs
@@ -35,7 +35,7 @@ use frame_system::{
 	pallet_prelude::*,
 };
 use loans::Position;
-use orml_traits::{Change, MultiCurrency};
+use orml_traits::{Change, GetByKey, MultiCurrency};
 use orml_utilities::OffchainErr;
 use primitives::{Amount, Balance, CurrencyId};
 use rand_chacha::{
@@ -149,6 +149,9 @@ pub mod module {
 		#[pallet::constant]
 		type MinimumDebitValue: Get<Balance>;
 
+		/// Gets the minimum collateral value for the given currency.
+		type MinimumCollateralValue: GetByKey<CurrencyId, Balance>;
+
 		/// Stablecoin currency id
 		#[pallet::constant]
 		type GetStableCurrencyId: Get<CurrencyId>;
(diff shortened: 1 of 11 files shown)


modules/cdp-engine/src/lib.rs
```diff
@@ -1316,6 +1321,12 @@ impl<T: Config> RiskManager<T::AccountId, CurrencyId, Balance, Balance> for Pall
 				debit_value >= T::MinimumDebitValue::get(),
 				Error::<T>::RemainDebitValueTooSmall,
 			);
+
+			// check the minimum_collateral_value
+			ensure!(
```
xlc, line 1326: check check should only happen when debit is zero
otherwise we enforce minimal debit which will indirectly enforce minimal collateral

syan095: Moved the collateral check to the case where debit is 0.
updated unit tests accordingly

runtime/acala/src/lib.rs
```diff
@@ -1018,6 +1018,12 @@ where
 	type Extrinsic = UncheckedExtrinsic;
 }
 
+parameter_type_with_key! {
+	pub MinimumCollateralAmount: |currency_id: CurrencyId| -> Balance {
+		ExistentialDeposits::get(currency_id).saturating_mul(100u128)
```
xlc, line 1023: it should be something similar to 

https://github.com/AcalaNetwork/Acala/blob/c4f40d1bfba1405c775ba87f57dd17d309290403/runtime/karura/src/lib.rs#L1917-L1921

because the ExistentialDeposits for KAR / ACA is configured to be infinite

xlc (changes requested):

runtime/acala/src/lib.rs
xlc: I think we want an integration test to confirm

syan095: Fixed and added an integration test

xlc (approved):

PR #2071, opened 2022-05-03
Remove honzon global paramter

### LINKED ISSUE #2006: Remove honzon global parameters
They are never used and we are unlikely to enable it.

syan095:
Also updated how interest rates are queried in cdp-engine module

Closes #2006


syan095:
Currently get_interest_rate_per_sec is only called by `accumulate_interest`, which is infallible.
@xlc can you confirm that we do not need to gate keep people from using collaterals without interest rate set (in adjust_position)?

modules/cdp-engine/src/lib.rs
```diff
@@ -579,8 +554,10 @@ impl<T: Config> Pallet<T> {
 			let interval_secs = now_secs.saturating_sub(last_accumulation_secs);
 
 			for currency_id in T::CollateralCurrencyIds::get() {
-				let rate_to_accumulate =
-					Self::compound_interest_rate(Self::get_interest_rate_per_sec(currency_id), interval_secs);
+				let rate_to_accumulate = Self::compound_interest_rate(
+					Self::get_interest_rate_per_sec(currency_id).unwrap_or_default(),
```
xlc, line 558: should just skip it if no interest rate is set

modules/cdp-engine/src/lib.rs
```diff
@@ -772,11 +749,10 @@ impl<T: Config> Pallet<T> {
 		Self::collateral_params(currency_id).required_collateral_ratio
 	}
 
-	pub fn get_interest_rate_per_sec(currency_id: CurrencyId) -> Rate {
+	pub fn get_interest_rate_per_sec(currency_id: CurrencyId) -> Result<Rate, DispatchError> {
 		Self::collateral_params(currency_id)
 			.interest_rate_per_sec
-			.unwrap_or_default()
-			.saturating_add(Self::global_interest_rate_per_sec())
+			.ok_or_else(|| Error::<T>::CollateralInterestRateNotSet.into())
```
xlc, line 755: make it `InvalidCollateralType`

xlc (dismissed):

modules/cdp-engine/src/lib.rs
```diff
@@ -777,11 +751,10 @@ impl<T: Config> Pallet<T> {
 		Self::collateral_params(currency_id).required_collateral_ratio
 	}
 
-	pub fn get_interest_rate_per_sec(currency_id: CurrencyId) -> Rate {
+	pub fn get_interest_rate_per_sec(currency_id: CurrencyId) -> Result<Rate, DispatchError> {
 		Self::collateral_params(currency_id)
 			.interest_rate_per_sec
-			.unwrap_or_default()
-			.saturating_add(Self::global_interest_rate_per_sec())
+			.ok_or_else(|| Error::<T>::InvalidCollateralType.into())
```
ferrell-code, line 757:
```suggestion
			.ok_or(Error::<T>::InvalidCollateralType.into())
```

modules/cdp-engine/src/tests.rs
```diff
@@ -29,6 +29,7 @@ use sp_io::offchain;
 use sp_runtime::{
 	offchain::{DbExternalities, StorageKind},
 	traits::BadOrigin,
+	ModuleError,
```
ferrell-code, line 32: remove, not used

ferrell-code (commented):
Some nits, but lgtm

modules/cdp-engine/src/lib.rs
syan095: i think clippy check fails with ok_or

ferrell-code: hmm that's weird, but doesn't matter it is the same thing

ferrell-code (approved):

PR #2073, opened 2022-05-03
currencies transfer erc20 will charge gas

### LINKED ISSUE #2072: improve currencies.transfer weight
Adjust the weight to consider the potential gas usage from erc20 transfer and refund the unused weight if the transfer is not erc20

ermalkaleci:
Closes #2072


xlc:
Can you check what's the tx fee difference for non erc20 vs erc20? Just sanity check it is still reasonable.

xlc (approved):

PR #2074, opened 2022-05-03
2.6.0

xlc:


runtime/karura/src/weights/module_dex.rs
```diff
@@ -46,156 +47,163 @@ use sp_std::marker::PhantomData;
 /// Weight functions for module_dex.
 pub struct WeightInfo<T>(PhantomData<T>);
 impl<T: frame_system::Config> module_dex::WeightInfo for WeightInfo<T> {
+	// Storage: unknown [0x3a7472616e73616374696f6e5f6c6576656c3a] (r:1 w:1)
```
xlc, line 50: need to investigate what is this unknown storage

zqhxuyuan: I've notice the result, seems substrate repo also have unknow storage. https://github.com/paritytech/substrate/blob/1a32a31ee8289b603c665766cd0dea0f6c96ea41/frame/proxy/src/weights.rs#L141

zqhxuyuan: this's because prefix not found: https://github.com/paritytech/substrate/blob/1a32a31ee8289b603c665766cd0dea0f6c96ea41/utils/frame/benchmarking-cli/src/pallet/writer.rs#L405-L407

PR #2083, opened 2022-05-05
Add Liquid Staking Precompiles

ferrell-code:
Todo:

- [x] create predeploy contracts
- [ ] fix homa benchmarks (maybe better to do in follow up pr) issue: #2084


runtime/common/src/precompile/input.rs
```diff
@@ -186,6 +187,21 @@ where
...
+			Ok(true)
+		} else if param.is_zero() {
+			Ok(false)
+		} else {
+			Err(PrecompileFailure::Revert {
+				exit_status: ExitRevert::Reverted,
+				output: "failed to decode bool".into(),
+				cost: self.target_gas.unwrap_or_default(),
+			})
+		}
```
zjb0807, line 203: I think `param.try_into().map_err(|_| PrecompileFailure::Revert {` is enough.

ferrell-code: I agree but unfortunately TryFrom<bool> is not implemented for unsigned integers. I also feel like it should be simpler but I don't know of a better way than if statement

runtime/common/src/precompile/homa.rs
```diff
@@ -0,0 +1,443 @@
...
+
+			// mint(address,uint256) -> 0x40c10f19
+			// who
+			// amount
+			let input = hex! {"
+                40c10f19
+                000000000000000000000000 1000000000000000000000000000000000000001
+                00000000000000000000000000000000 0000000000000000000000003b9aca00
+            "};
+
```
zjb0807, line 280: format

runtime/common/src/precompile/homa.rs
```diff
@@ -0,0 +1,443 @@
...
+			// who
+			// amount
+			// fast_match
+			let input = hex! {
+				"c941744a
+                000000000000000000000000 1000000000000000000000000000000000000001
+                00000000000000000000000000000000 000000000000000000000000000aca00
+                00000000000000000000000000000000 00000000000000000000000000000000
+                "
+			};
```
zjb0807, line 324: format

runtime/common/src/precompile/homa.rs
```diff
@@ -0,0 +1,442 @@
...
+					logs: Default::default(),
+				})
+			}
+			Action::GetExchangeRate => {
+				let rate =
+					<module_homa::Pallet<Runtime> as HomaManager<Runtime::AccountId, Balance>>::get_exchange_rate();
+				Ok(PrecompileOutput {
+					exit_status: ExitSucceed::Returned,
+					cost: gas_cost,
+					output: Output::default().encode_u128(rate.into_inner()),
```
ferrell-code, line 148: Not sure about what fixed point arithmetic libraries are best in solidity, but I figure that this format will be usable enough

runtime/common/src/precompile/homa.rs
ferrell-code: ?

runtime/common/src/precompile/homa.rs
zjb0807: I think it's ok. The same as here.
https://github.com/AcalaNetwork/Acala/blob/45c41201fc1dc7702d5cb27eb214f0c2996c4324/runtime/common/src/precompile/oracle.rs#L102

zjb0807 (approved):
LGTM

runtime/common/src/precompile/input.rs
```diff
@@ -186,6 +187,21 @@ where
 
 		Ok(bytes.to_vec())
 	}
+
+	fn bool_at(&self, index: usize) -> Result<bool, Self::Error> {
+		let param = self.u64_at(index)?;
```
ermalkaleci, line 192: `self.u256_at(index)
			.map(|x| !x.is_zero())
			.map_err(|_| PrecompileFailure::Revert {
				exit_status: ExitRevert::Reverted,
				output: "failed to decode bool".into(),
				cost: self.target_gas.unwrap_or_default(),
			})`

runtime/common/src/precompile/homa.rs
```diff
@@ -0,0 +1,442 @@
...
+				let cost = InputPricer::<Runtime>::read_accounts(1);
+				let weight = <Runtime as module_homa::Config>::WeightInfo::request_redeem();
+
+				cost.saturating_add(WeightToGas::convert(weight))
+			}
+			Action::GetExchangeRate => {
+				// Homa::TotalVoidLiquid (r: 1)
+				// Homa::ToBondPool (r: 1)
+				// Tokens::TotalIssuance(r: 1)
+				// Homa::StakingLedgers(r: ?)
```
ermalkaleci, line 215: is this verified to be 1?

ferrell-code: Yeah that's the homa benchmarking issue #2084. 

It depends on storage for number of reads (three reads on karura right now), needs to be fixed as currently `mint` and `request_redeem` extrinsic are wrong in the same manner

runtime/common/src/precompile/input.rs
ferrell-code: Yeah I thought about that too, but then a random integer value will be true rather than erroring out. 

But maybe this behavior is fine too, I don't have a strong opinion

ermalkaleci: the call will come from predeployed contract and will be encoded so it will either be one or zero

zjb0807 (approved):

xlc (approved):

zjb0807 (approved):

Issue #2081, opened 2022-05-04
Auto publish contracts that deployed from published contracts


xlc:
So that contract maintainer don't need to manually publish contracts for all the new contracts deployed by contract factory

PR #2099, opened 2022-05-10
Auto publish contracts

zjb0807:
Closes: #2081 

And fixed the maintainer of the contract creates a new contract.


modules/evm/src/runner/state.rs
```diff
@@ -147,6 +147,8 @@ pub struct StackSubstateMetadata<'config> {
 	caller: Option<H160>,
 	// save the contract to charge storage
 	target: Option<H160>,
+	// save the call contract address, publish status will sync from it.
+	call_address: Option<H160>,
```
ermalkaleci, line 151: this will make more sense to be `code_address`

ermalkaleci: or `origin_code_address` might be better

modules/evm/src/tests.rs
```diff
@@ -2080,3 +2080,168 @@ fn remove_account_works() {
...
+			})
+		);
+		assert_eq!(
+			EVM::accounts(H160::from_str("0x769a55efaf4dbdd6f44efce668455522b61abb82").unwrap())
+				.unwrap()
+				.contract_info,
+			Some(ContractInfo {
+				code_hash: H256::from_str("0x21da9211a5ef5e10cf7efc58b70c09e2a586844a4798b5e506927cc05808ca4c")
+					.unwrap(),
+				maintainer: H160::from_str("0x39b26a36a8a175ce7d498b5ef187d1ab2f381bbd").unwrap(),
```
ermalkaleci, line 2242: shouldn't this be `alice()`?

zjb0807: @xlc Do we want to inherit maintainer from origin_code_address?

xlc: I think this is fine. The contract factor should be coded in such way that can manage child contract. e.g. if the deployer want to have the child contract published with unpublished factory, they should coded the factory to publish the child contract after deploy.

modules/evm/src/lib.rs
```diff
@@ -1296,13 +1293,8 @@ impl<T: Config> Pallet<T> {
...
 		// if source is account, the maintainer of the new contract is source.
-		// if source is contract, the maintainer of the new contract is the maintainer of the contract.
-		let maintainer = Self::accounts(source).map_or(source, |account_info| {
-			account_info
-				.contract_info
-				.map_or(source, |contract_info| contract_info.maintainer)
-		});
-
+		// if source is contract, the maintainer of the new contract is the contract.
+		let maintainer = source;
```
zjb0807, line 1297: There was something wrong with the original logic. 
If the contract creates the contract, its maintainer should be the contract

xlc (approved):

PR #2102, opened 2022-05-10
Add a default swap with best price implementation

zqhxuyuan:
- [x] add a `swap_with_best_price` interface to `DexManager`. real implementation in module dex return `Error::<T>::CannotSwap`
- [x] add `CannotSwap` error type in module dex, invoke of `swap_with_best_price` now use module dex's CannotSwap. ~~The error in trait retuning `DispatchError::CannotLookup` needs to change to module specific error type, such as `CannotSwap` in cdp-engine or cdp-treasury.~~
- [x] remove `AlternativeSwapPathJointList` in `cdp_treasury` and `cdp_engine` to `dex` as it's only used by `swap_with_best_price`.

### DIFF
--- a/modules/cdp-engine/src/lib.rs
+++ b/modules/cdp-engine/src/lib.rs
@@ -129,10 +129,6 @@ pub mod module {
 		/// always do this.
 		type UpdateOrigin: EnsureOrigin<Self::Origin>;
 
-		/// The list of valid collateral currency types
-		#[pallet::constant]
-		type CollateralCurrencyIds: Get<Vec<CurrencyId>>;
-
 		/// The default liquidation ratio for all collateral types of CDP
 		#[pallet::constant]
 		type DefaultLiquidationRatio: Get<Ratio>;
@@ -214,7 +210,8 @@ pub mod module {
 		InvalidCollateralType,
 		/// Remain debit value in CDP below the dust amount
 		RemainDebitValueTooSmall,
-		/// Remain collateral value in CDP below the dust amount
+		/// Remain collateral value in CDP below the dust amount.
+		/// Withdraw all collateral or leave more than the minimum.
 		CollateralAmountBelowMinimum,
 		/// Feed price is invalid
 		InvalidFeedPrice,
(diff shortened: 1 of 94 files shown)


modules/cdp-engine/src/lib.rs
```diff
@@ -816,6 +816,37 @@ impl<T: Config> Pallet<T> {
 		Ok(())
 	}
 
+	/// If reverse is false, exchange stable coin with given token.
+	/// If reverse is true, exchange given token with stable coin.
```
shaunxw, line 820:
```suggestion
	/// If reverse is false, swap stable coin to given `token`.
	/// If reverse is true, swap given `token` to stable coin.
```

modules/cdp-engine/src/lib.rs
```diff
@@ -1103,7 +1086,7 @@ impl<T: Config> Pallet<T> {
 		// refund remain collateral to CDP owner
 		let refund_collateral_amount = collateral
 			.checked_sub(actual_supply_collateral)
-			.expect("swap succecced means collateral >= actual_supply_collateral; qed");
+			.expect("swap succeed means collateral >= actual_supply_collateral; qed");
```
shaunxw, line 1089:
```suggestion
			.expect("swap success means collateral >= actual_supply_collateral; qed");
```

modules/cdp-engine/src/lib.rs
```diff
@@ -1224,7 +1207,7 @@ impl<T: Config> Pallet<T> {
 			) {
 			let refund_collateral_amount = amount
 				.checked_sub(actual_supply_collateral)
-				.expect("swap succecced means collateral >= actual_supply_collateral; qed");
+				.expect("swap succeed means collateral >= actual_supply_collateral; qed");
```
shaunxw, line 1210:
```suggestion
				.expect("swap success means collateral >= actual_supply_collateral; qed");
```

shaunxw (commented):
Default error should be mapped to pallet specific error on calling `T::DEX::swap_supply_with_best_price_swap_path`.

modules/cdp-engine/src/tests.rs
```diff
@@ -416,7 +416,7 @@ fn expand_position_collateral_work() {
 
 		assert_noop!(
 			CDPEngineModule::expand_position_collateral(&ALICE, DOT, 0, 1),
-			Error::<Runtime>::CannotSwap
+			DispatchError::CannotLookup
```
shaunxw, line 419: This breaking change doesn't make sense to me. The reason of swap failure could be different, instead of just lookup errors.

shaunxw (changes requested):

modules/cdp-engine/src/tests.rs
zqhxuyuan: Add another Error type `CannotLookup` in dex module, now when invoke `swap_supply_with_best_price_swap_path`/`swap_with_best_price`, the error is CannotLookup defined in dex module.

wangjj9219 (approved):

shaunxw (approved):

PR #2131, opened 2022-05-18
fix stable asset token

xlc:


runtime/acala/src/lib.rs
```diff
@@ -1560,7 +1560,7 @@ type RebaseTokens = orml_tokens::Combiner<
 	AccountId,
 	IsLiquidToken,
 	orml_tokens::Mapper<AccountId, Tokens, ConvertBalanceHoma, Balance, GetLiquidCurrencyId>,
```
zjb0807, line 1562:
```suggestion
	orml_tokens::Mapper<AccountId, Currencies, ConvertBalanceHoma, Balance, GetLiquidCurrencyId>,
```

It's `Currencies` on mandala runtime

PR #2138, opened 2022-05-19
Removed the integration-tests feature from karura-runtime

### LINKED ISSUE #2009: remove `integration_tests_config` file and `integration-tests` feature flag
after https://github.com/AcalaNetwork/Acala/blob/master/runtime/karura/src/integration_tests_config.rs#L43
we can register local asset in integration-tests by hand, instead use another different config of production runtime, and remove some code now.

syan095:
Closes #2009


runtime/karura/src/xcm_config.rs
```diff
@@ -161,6 +157,14 @@ parameter_types! {
 		// BNC:KSM = 80:1
 		ksm_per_second() * 80
 	);
+	pub BncPerSecondOfCanonicalLocation: (AssetId, u128) = (
+		MultiLocation::new(
+			0,
+			X1(GeneralKey(parachains::bifrost::BNC_KEY.to_vec())),
```
zqhxuyuan, line 163: BNC shouldn't be our side token, for the testcase working, use register native asset.

zqhxuyuan: try register foreign asset, as foreign asset has Trader while native asset do not exist Trader.

primitives/src/currency.rs
```diff
@@ -430,3 +430,9 @@ pub struct AssetMetadata<Balance> {
 	pub decimals: u8,
 	pub minimal_balance: Balance,
 }
+
+#[derive(Clone, Eq, PartialEq, RuntimeDebug, Encode, Decode, TypeInfo)]
+pub struct ForeignAssetExtraMetadata {
+	pub maybe_currency_id: Option<CurrencyId>,
+	pub maybe_location: Option<xcm::latest::MultiLocation>,
```
zqhxuyuan, line 437: rename to `override_currency_id` and `override_location`. IIUC, `maybe_currency_id` will override ForeignAsset(id), and `maybe_location` will override registered location using in the `AssetIdMapping` trait.

syan095: Yep renamed

runtime/integration-tests/src/relaychain/kusama_cross_chain_transfer.rs
```diff
@@ -97,13 +97,31 @@ fn transfer_sibling_chain_asset() {
 	fn karura_reserve_account() -> AccountId {
 		polkadot_parachain::primitives::Sibling::from(2000).into_account()
 	}
+	fn sibling_reserve_account() -> AccountId {
+		polkadot_parachain::primitives::Sibling::from(2002).into_account()
```
shaunxw, line 101: define a `const` to avoid repeating magic number `2002`

runtime/integration-tests/src/relaychain/kusama_cross_chain_transfer.rs
```diff
@@ -364,75 +409,75 @@ fn test_asset_registry_module() {
...
+	}
```
shaunxw, line 414: duplicated definitions with in `transfer_sibling_chain_asset`

runtime/integration-tests/src/relaychain/kusama_cross_chain_transfer.rs
```diff
@@ -115,7 +136,7 @@ fn transfer_sibling_chain_asset() {
 				MultiLocation::new(
 					1,
 					X2(
-						Parachain(2001),
+						Parachain(SIBLING_ID),
```
zqhxuyuan, line 139: This is NonReserve case, Karura(2000)-[Bifrost2002 token]->Sibling(2001), can we add another testcase: Karura-[Bifrost token]->Bifrost?

syan095: That would be the same integration test as before - the problem is that Native BNC token is not setup correctly in Karura runtime(it is defined as external token with a different parachain ID, not a native currency). If we want the setup for the test, we will have to add BNC to 
https://github.com/AcalaNetwork/Acala/blob/950a179bcb53e5a84e5f960fdbed34128e3d23d7/runtime/karura/src/xcm_config.rs#L453
which would not fit the Karura runtime.

see how Bifrost setup BNC and how Karura setup KAR
https://github.com/bifrost-finance/bifrost/blob/f862ae756eccb444d5ed0aa8617803c2da94d2e5/runtime/bifrost-kusama/src/xcm_config.rs#L207

Perhaps we can test transfering KARs from Karura to sibling chain? that will fit the current setup

Alternatively I will have to actually bring in Bifrost's Runtime to our integration test, but i think that would be a bit of overkill

syan095: With the current setup, when Karura sends BNC over to 2001, it's gonna arrive as (parent 0, X1([0, 1)], which we can register as a foreign asset. The only way to change that is to setup the currency correctly (like how bifrost do it here:
https://github.com/bifrost-finance/bifrost/blob/f862ae756eccb444d5ed0aa8617803c2da94d2e5/runtime/bifrost-kusama/src/xcm_config.rs#L207
)

I've added the test, but i dont think there's much benefit we get from it

zqhxuyuan (approved):

PR #2152, opened 2022-05-23
Return default value instead of error

zjb0807:
The contract query interface, if the data does not exist, return the default value

### DIFF
--- a/runtime/common/src/precompile/dex.rs
+++ b/runtime/common/src/precompile/dex.rs
@@ -116,23 +116,19 @@ where
 					currency_id_a, currency_id_b
 				);
 
-				let value = <module_dex::Pallet<Runtime> as DEXManager<Runtime::AccountId, Balance, CurrencyId>>::get_liquidity_token_address(currency_id_a, currency_id_b)
-					.ok_or_else(||
-								PrecompileFailure::Revert {
-									exit_status: ExitRevert::Reverted,
-									output: "Dex get_liquidity_token_address failed".into(),
-									cost: target_gas_limit(target_gas).unwrap_or_default(),
-								})?;
+				// If it does not exist, return address(0x0). Keep the behavior the same as mapping[key]
+				let address = <module_dex::Pallet<Runtime> as DEXManager<Runtime::AccountId, Balance, CurrencyId>>::get_liquidity_token_address(currency_id_a, currency_id_b)
+					.unwrap_or_default();
 
 				Ok(PrecompileOutput {
 					exit_status: ExitSucceed::Returned,
 					cost: gas_cost,
-					output: Output::default().encode_address(&value),
+					output: Output::default().encode_address(&address),
 					logs: Default::default(),
 				})
 			}
 			Action::GetSwapTargetAmount => {
-				// solidity abi enocde array will add an offset at input[1]
+				// solidity abi encode array will add an offset at input[1]
 				let supply_amount = input.balance_at(2)?;
 				let path_len = input.u32_at(3)?;
 				let mut path = vec![];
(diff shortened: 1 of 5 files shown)


ermalkaleci (approved):

ermalkaleci:
@zjb0807 we need to recheck all precompiles. I think all reads need to return value instead of revert

zjb0807:
> @zjb0807 we need to recheck all precompiles. I think all reads need to return value instead of revert

This decoding fails, I think it's an input problem and should return an error.
https://github.com/AcalaNetwork/Acala/blob/da6da90a738f77675829c14e994f52e0c1ea8ccc/runtime/common/src/precompile/multicurrency.rs#L149

runtime/common/src/precompile/multicurrency.rs
```diff
@@ -98,11 +98,8 @@ where
 
 		match action {
 			Action::QueryName => {
-				let name = Runtime::Erc20InfoMapping::name(currency_id).ok_or_else(|| PrecompileFailure::Revert {
-					exit_status: ExitRevert::Reverted,
-					output: "Get name failed".into(),
-					cost: target_gas_limit(target_gas).unwrap_or_default(),
-				})?;
+				// If get name failed, return ''
+				let name = Runtime::Erc20InfoMapping::name(currency_id).unwrap_or_default();
```
xlc, line 102: I think those one can revert. Similar to querying an invalid ERC20

xlc (commented):
we should check the behaviour of similar contracts on Eth and mimic the behaviour

runtime/common/src/precompile/multicurrency.rs
zjb0807: Return `0x` is the default behavior

zjb0807:
```
{
    "jsonrpc": "2.0",
    "method": "eth_call",
    "params": [
        {
            "to": "0xdac17f958d2ee523a2206206994597c13d831ec7",
            "data": "0x06fdde03"
        },
        "latest"
    ],
    "id": 1
}
```

on ethereum, this call return the name of USDT(ERC20).
```
{
    "jsonrpc": "2.0",
    "id": 1,
    "result": "0x0000000000000000000000000000000000000000000000000000000000000020000000000000000000000000000000000000000000000000000000000000000a5465746865722055534400000000000000000000000000000000000000000000"
}
```

zjb0807: change the `to` address to others non-erc20 address, it'll return `0x`

zjb0807: https://etherscan.io/tx/0x78810f6185dec0029243d49808e657b66b034d68a493b1edd659158a19479136

FUNCTIONHASHES : `"06fdde03": "name()",`
```
{
    "jsonrpc": "2.0",
    "method": "eth_call",
    "params": [
        {
            "to": "0x881d40237659c251811cec9c364ef91dc08d300c",
            "data": "0x06fdde03"
        },
        "latest"
    ],
    "id": 1
}
```
return:
```
{
    "jsonrpc": "2.0",
    "id": 1,
    "error": {
        "code": -32000,
        "message": "execution reverted"
    }
}
```

zjb0807: If the contract does not exist, return 0x, if it is a non-erc20 contract, return revert. I'll revert it.

zjb0807:
> @zjb0807 I think we can update return data to have status as well i.e. (0, false) balance zero and call failed. I think this is important because those calls aren't trying to change anything but just checking so it won't make sense to revert entire transaction if a check isn't successful

https://github.com/certusone/wormhole/blob/dev.v2/ethereum/contracts/bridge/Bridge.sol#L264-L265

No matter what data we return(`(0, false)` or `0`), the caller needs to check the data, so I think it's fine.

ermalkaleci:
> > @zjb0807 I think we can update return data to have status as well i.e. (0, false) balance zero and call failed. I think this is important because those calls aren't trying to change anything but just checking so it won't make sense to revert entire transaction if a check isn't successful
> 
> https://github.com/certusone/wormhole/blob/dev.v2/ethereum/contracts/bridge/Bridge.sol#L264-L265
> 
> No matter what data we return(`(0, false)` or `0`), the caller needs to check the data, so I think it's fine.

I'm not talking about name or symbol specifically. I am talking about read methods in general

runtime/common/src/precompile/dex.rs
```diff
@@ -145,24 +141,20 @@ where
 					path, supply_amount
 				);
 
-				let value = <module_dex::Pallet<Runtime> as DEXManager<Runtime::AccountId, Balance, CurrencyId>>::get_swap_amount(&path, SwapLimit::ExactSupply(supply_amount, Balance::MIN))
+				// If get_swap_amount fail, return 0.
```
xlc, line 144: need to make sure we have some docs on predeploy to require caller to handle the case when the return is zero

xlc (approved):

Issue #2139, opened 2022-05-19
Network fee distribution and Incentive Pools


zqhxuyuan:
Detail in: https://hackmd.io/yjZYbK2oRpaLsFSQlP6ACg

Current state Vs Expected state:

<img width="1144" alt="image" src="https://user-images.githubusercontent.com/1088525/169225286-5e492dcf-5096-4709-be0f-1ef9367ee7e2.png">

PR #2153, opened 2022-05-23
fee distributionn framework

zqhxuyuan:
relates to: https://github.com/AcalaNetwork/Acala/issues/2139


xlc:
swap fee is not yet implemented so leave it for now
xcm execution fee are charged by relaychain so we don't actually have those fees and we can also ignore it

zqhxuyuan:
> swap fee is not yet implemented so leave it for now xcm execution fee are charged by relaychain so we don't actually have those fees and we can also ignore it

xcm fee here means `buy_weight` revenue, and it go to our treasury account: https://github.com/AcalaNetwork/Acala/blob/7727abeac1cb999ae07495a7836a695a5b1f8583/runtime/karura/src/xcm_config.rs#L100-L116

PR #2164, opened 2022-05-26
Liquidation improvement

shaunxw:
TODO:
- [x] benchmarks
- [x] unit tests
- [x] integration tests - Roy Yang


modules/cdp-engine/src/lib.rs
```diff
@@ -1216,59 +1272,212 @@ impl<T: Config> Pallet<T> {
...
 			)?;
 		}
 
 		Ok(())
 	}
-	pub fn get_collateral_currency_ids() -> Vec<CurrencyId> {
-		CollateralParams::<T>::iter_keys().collect()
+}
+
+pub struct LiquidateViaContracts<T>(PhantomData<T>);
```
ermalkaleci, line 1342: @shaunxw I think this needs to have a gasLimit or blockWeight ratio and we can iterate until OOG

shaunxw: Currently for each contract call there is a gas limit (and storage limit) set, like for `liquidate` https://github.com/AcalaNetwork/Acala/blob/ce08df858d3bdb08e7c6b3c5f201ab3bc5dc77a1/modules/evm-bridge/src/lib.rs#L282-L289

Additionally there is a max limit for the number of contracts. I think these should do the work?

ermalkaleci: yeah if we manage the limit then yes, but I still think we need to make this smarter. how do we know the gas limit is enough? do we need to retry with more gas? I think we can do some improvements

wangjj9219 (commented):
Look good! 
I want to confirm, is `LiquidateViaContracts` actually valid only for ERC20 type collateral? Will it work for native token collateral?

modules/cdp-engine/src/lib.rs
shaunxw: We don't need to retry. Smart contracts need to be implemented to meet gas limit requirements.

shaunxw:
> I want to confirm, is LiquidateViaContracts actually valid only for ERC20 type collateral? Will it work for native token collateral?

Technically `LiquidateViaContracts` only works with ERC20 tokens liquidation. Native tokens will be supported as pre-deployed ERC20 token contracts, via multi-currency precompile.

wangjj9219:
better add the `CDPEnginePalletId ` to the `fn get_all_module_accounts()` if native tokens are supported by `LiquidateViaContracts `

modules/evm-bridge/src/lib.rs
```diff
@@ -250,6 +257,77 @@ impl<T: Config> EVMBridgeTrait<AccountIdOf<T>, BalanceOf<T>> for EVMBridge<T> {
...
+
+	fn on_collateral_transfer(context: InvokeContext, collateral: EvmAddress, amount: Balance) {
+		// liquidation contract method hash
+		let mut input = Into::<u32>::into(Action::OnCollateralTransfer).to_be_bytes().to_vec();
+		// append collateral ERC20 address
+		input.extend_from_slice(H256::from(collateral).as_bytes());
+		// append collateral amount
+		input.extend_from_slice(H256::from_uint(&U256::from(amount)).as_bytes());
+
+		let _ = T::EVM::execute(
```
syan095, line 302: is it OK to ignore the success status here?

shaunxw: `on_collateral_transfer` is for notifying contracts, there isn't much could be done for runtime if something is wrong, other than probably logging the result

syan095: So basically we assume if something goes wrong in that function it's the contract's fault and we are not responsible for its errors.

ermalkaleci:
need to update `liquidate_unsafe_cdp` returned weight

modules/cdp-engine/src/lib.rs
```diff
@@ -1215,59 +1269,207 @@ impl<T: Config> Pallet<T> {
...
+		let collateral = currency_id
+			.erc20_address()
+			.ok_or(Error::<T>::CollateralContractNotFound)?;
+		let repay_dest = Pallet::<T>::evm_address();
+		let repay_dest_account_id = Pallet::<T>::account_id();
+
+		let stable_coin = T::GetStableCurrencyId::get();
+
+		let contracts_by_priority = {
+			let now: u32 = frame_system::Pallet::<T>::current_block_number()
```
ermalkaleci, line 1373:
```suggestion
			let now: usize = frame_system::Pallet::<T>::current_block_number()
```

modules/cdp-engine/src/lib.rs
```diff
@@ -1215,59 +1269,207 @@ impl<T: Config> Pallet<T> {
...
+			right
+		};
+
+		// try liquidation on each contract
+		for contract in contracts_by_priority.into_iter() {
+			let repay_dest_balance = CurrencyOf::<T>::free_balance(stable_coin, &repay_dest_account_id);
+			if T::LiquidationEvmBridge::liquidate(
+				InvokeContext {
+					contract,
+					sender: Pallet::<T>::evm_address(),
```
ermalkaleci, line 1390:
```suggestion
					sender: repay_dest,
```

modules/cdp-engine/src/lib.rs
```diff
@@ -1215,59 +1269,207 @@ impl<T: Config> Pallet<T> {
...
+						log::error!(
+							target: "cdp-engine",
+							"LiquidateViaContracts: transfer collateral to contract failed. \
+							Collateral: {:?}, amount: {:?} contract: {:?}, error: {:?}. \
+							This is unexpected, need extra action.",
+							currency_id, collateral_supply, contract, e,
+						);
+					}
+					// notify liquidation success
+					T::LiquidationEvmBridge::on_collateral_transfer(
```
ermalkaleci, line 1419: shouldn't we do this if above operation succeed?

modules/cdp-engine/src/lib.rs
```diff
@@ -1215,59 +1269,207 @@ impl<T: Config> Pallet<T> {
...
+						},
+						collateral,
+						target_stable_amount,
+					);
+					// refund rest collateral to CDP owner
+					let refund_collateral_amount = amount
+						.checked_sub(collateral_supply)
+						.expect("Ensured collateral supply <= amount; qed");
+					if !refund_collateral_amount.is_zero() {
+						<T as Config>::CDPTreasury::withdraw_collateral(who, currency_id, refund_collateral_amount)?;
```
ermalkaleci, line 1433: I think if any of `method()?;` will fail it will return `Err` and not continue with other contracts

modules/cdp-engine/src/lib.rs
```diff
@@ -1215,59 +1269,207 @@ impl<T: Config> Pallet<T> {
...
+					.expect("the oracle price should be available because liquidation are triggered by it.")
+					.saturating_mul_int(target_stable_amount),
+			);
+		let collateral_supply = amount.min(max_supply_limit);
+
+		let collateral = currency_id
+			.erc20_address()
+			.ok_or(Error::<T>::CollateralContractNotFound)?;
+		let repay_dest = Pallet::<T>::evm_address();
+		let repay_dest_account_id = Pallet::<T>::account_id();
```
syan095, line 1368: Here we are using CdpEngine's pallet account as the repayment account for contract liquidation. Is there a specific reason for this instead of using the Treasury's account? 
Since the Treasury has the collateral and the debt, would it be easier if we use Treasury's account here instead?

shaunxw: I don't see how it gets "easier" using treasury account. Mixing up all balances in one treasury account will make it difficult to tell what goes wrong if anything unexpected happened.

modules/cdp-engine/src/lib.rs
shaunxw: updated

modules/cdp-engine/src/lib.rs
```diff
@@ -186,6 +190,21 @@ pub mod module {
 		/// Swap
 		type Swap: Swap<Self::AccountId, Balance, CurrencyId>;
 
+		/// When settle collateral with smart contracts, he acceptable max slippage for the price
```
xlc, line 193:
```suggestion
		/// When settle collateral with smart contracts, the acceptable max slippage for the price
```

modules/cdp-engine/src/lib.rs
```diff
@@ -48,16 +48,19 @@ use sp_runtime::{
...
-	DispatchError, DispatchResult, FixedPointNumber, RuntimeDebug,
+	ArithmeticError, DispatchError, DispatchResult, FixedPointNumber, RuntimeDebug,
 };
-use sp_std::prelude::*;
+use sp_std::{marker::PhantomData, prelude::*};
 use support::{
-	CDPTreasury, CDPTreasuryExtended, DEXManager, EmergencyShutdown, ExchangeRate, Price, PriceProvider, Rate, Ratio,
-	RiskManager, Swap, SwapLimit,
+	AddressMapping, CDPTreasury, CDPTreasuryExtended, DEXManager, EmergencyShutdown, ExchangeRate, InvokeContext,
+	LiquidateCollateral, LiquidationEvmBridge as LiquidationEvmBridgeT, Price, PriceProvider, Rate, Ratio, RiskManager,
```
xlc, line 62: I don't think we have to rename `LiquidationEvmBridge` to `LiquidationEvmBridgeT`?

modules/cdp-engine/src/lib.rs
```diff
@@ -1215,59 +1269,218 @@ impl<T: Config> Pallet<T> {
...
+							Collateral: {:?}, amount: {:?} contract: {:?}, error: {:?}. \
+							This is unexpected, need extra action.",
+							currency_id, collateral_supply, contract, e,
+						);
+					} else {
+						// notify liquidation success
+						T::LiquidationEvmBridge::on_collateral_transfer(
+							InvokeContext {
+								contract,
+								sender: Pallet::<T>::evm_address(),
```
xlc, line 1422:
```suggestion
								sender: repay_dest,
```

modules/cdp-engine/src/lib.rs
```diff
@@ -490,6 +524,26 @@ pub mod module {
 			CollateralParams::<T>::insert(currency_id, collateral_params);
 			Ok(())
 		}
+
+		#[pallet::weight(<T as Config>::WeightInfo::register_liquidation_contract())]
+		#[transactional]
+		pub fn register_liquidation_contract(origin: OriginFor<T>, address: EvmAddress) -> DispatchResult {
+			T::UpdateOrigin::ensure_origin(origin)?;
```
xlc, line 531: we want a different origin to register this. the current UpdateOrigin is financial council but register liquidation contracts should be GeneralCouncil

xlc (dismissed):

xlc (approved):

Issue #2167, opened 2022-05-27
Support ERC-20 xcm transfer


zqhxuyuan:
following https://github.com/AcalaNetwork/Acala/pull/2165. Current Erc20 token withdraw operation not support in module_currencies, that'll cause xcm transfer failed in our side.
https://github.com/AcalaNetwork/Acala/blob/0ae648f4bacfaec7905beed522f1f7b875b55f47/modules/currencies/src/lib.rs#L375

zqhxuyuan:
asset-registry also need support Erc20, or else there'll be none matched Trader: https://github.com/AcalaNetwork/Acala/blob/ded6de57234c4367401dbd758db609254c2e00e0/modules/asset-registry/src/lib.rs#L598

zqhxuyuan:
when Sibling transfer (1, 2000, Erc20) to Karura, xcm execution error in Karura side:

```
2022-06-02 20:44:42.055 TRACE tokio-runtime-worker xcm::execute_xcm_in_credit: [Parachain] 
result: Err(ExecutorError { index: 0, xcm_error: FailedToTransactAsset("RealOriginNotFound"), weight: 600000000 })    
```

PR #2168, opened 2022-05-27
erc20 xtoken transfer

zqhxuyuan:
orml companion: https://github.com/open-web3-stack/open-runtime-module-library/pull/752
closes: https://github.com/AcalaNetwork/Acala/issues/2167

- [x] support ERC20 withdraw and deposit, as evm not support it, we're using `Erc20HoldingAccount` to transfer.
- [x] rename `FixedRateOfForeignAsset` to `FixedRateOfAssetRegistry`, and support both `ForeignAsset` and `Erc20` 
    - introduce `BuyWeightRate` to calculate rate by location.
- [x] testcase for Erc20 xcm transfer
    - Karura send USDC to Sibling with location: (1, 2000, GeneralKey(Ecr20(evm_address))). Sibling will take it as ForeignAsset.
    - Sibling send back USDC to Karura with location (1, 2000, GeneralKey(Ecr20(evm_address))), then in Karura, it'll be (0, GeneralKey(Ecr20(evm_address))), and it'll be consider as Erc20 token
    - Sibling send USDC to Karura with location: (1, 2001, GeneralKey(Ecr20(evm_address))). Karura will take it as ForeignAsset.


xlc (dismissed):

runtime/acala/src/lib.rs
```diff
@@ -839,13 +839,15 @@ parameter_types! {
 	pub const GetStableCurrencyId: CurrencyId = AUSD;
 	pub const GetLiquidCurrencyId: CurrencyId = LDOT;
 	pub const GetStakingCurrencyId: CurrencyId = DOT;
+	pub AccountForTransfer: AccountId = AccountId::new([0u8; 32]);
```
zjb0807, line 842: We should avoid using the 0 account id, it is not secure.
We can use `NetworkContractSource`

modules/currencies/src/lib.rs
```diff
@@ -90,6 +90,9 @@ pub mod module {
 		#[pallet::constant]
 		type GetNativeCurrencyId: Get<CurrencyId>;
 
+		/// Used as temporary account for ERC20 token `withdraw` and `deposit`.
+		type AccountForTransfer: Get<Self::AccountId>;
```
xlc, line 94: I don't like this name. Maybe Erc20HoldingAccount?

runtime/acala/src/lib.rs
zqhxuyuan: `NetworkContractSource` in the runtime is `H160::from_low_u64_be(0)`, it'll be error when trying send to it. can we use `H160::from_low_u64_be(1)`?

runtime/karura/src/lib.rs
```diff
@@ -854,13 +854,15 @@ parameter_types! {
 	pub const GetStableCurrencyId: CurrencyId = KUSD;
 	pub const GetLiquidCurrencyId: CurrencyId = LKSM;
 	pub const GetStakingCurrencyId: CurrencyId = KSM;
+	pub Erc20HoldingAccount: H160 = H160(hex_literal::hex!("000000000000000000ff00000000000000000000"));
```
shaunxw, line 857: This address looks pretty arbitrary to me. Can we use an account derived from currencies module pallet id?

xlc: we want to use a system account, that's the one starts `000000000000000000`

I think we want to have this defined around here https://github.com/AcalaNetwork/Acala/blob/ded6de57234c4367401dbd758db609254c2e00e0/primitives/src/evm.rs#L129-L153

xlc (dismissed):

shaunxw (dismissed):

zjb0807 (approved):

PR #2169, opened 2022-05-28
Convert rebase token amount when calculate swap amount for StableAsset

wangjj9219:


modules/aggregated-dex/src/lib.rs
```diff
@@ -169,6 +192,110 @@ pub mod module {
...
+		})
+	}
+
+	fn taiga_get_swap_input_amount(
+		pool_id: StableAssetPoolId,
+		input_asset_index: PoolTokenIndex,
+		output_asset_index: PoolTokenIndex,
+		output_amount: Balance,
+	) -> Option<(Balance, Balance)> {
+		let pool_info = T::StableAsset::pool(pool_id)?;
```
zqhxuyuan, line 221: add `.ok_or(Error::<T>::InvalidPoolId)` here.

modules/aggregated-dex/src/lib.rs
```diff
@@ -169,6 +192,110 @@ pub mod module {
...
+		let input_currency_id = pool_info.assets.get(input_asset_index as usize)?;
+		let output_currency_id = pool_info.assets.get(output_asset_index as usize)?;
```
zqhxuyuan, line 223: add `.ok_or(Error::<T>::InvalidTokenIndex)` here would be better too.

modules/aggregated-dex/src/lib.rs
zqhxuyuan: Ah, the return value is Option, not Result. ignore me.

modules/aggregated-dex/src/mock.rs
```diff
@@ -122,340 +121,83 @@ impl module_dex::Config for Runtime {
...
-			Some(SwapResult {
-				dx: supply_amount,
-				dy: dy_bal,
-				..Default::default()
-			})
-		})
+pub struct IsLiquidToken;
+impl Contains<CurrencyId> for IsLiquidToken {
+	fn contains(currency_id: &CurrencyId) -> bool {
+		matches!(currency_id, CurrencyId::Token(TokenSymbol::LDOT))
```
zqhxuyuan, line 157: This can simplify: 
```rust
match_types! {
	pub type IsLiquidToken: impl Contains<CurrencyId> = {
		CurrencyId::Token(TokenSymbol::LDOT)
	};
}
```

xlc (commented):
Another way to handle the rebase token is to create another implementation of StableAsset that wraps the original one and handle the rebase logic there.

wangjj9219:
> Another way to handle the rebase token is to create another implementation of StableAsset that wraps the original one and handle the rebase logic there.

You are right, it's better to add a wrapper to hide the processing logic, and it will be easier to use for other pallet modules. `StableAsset` has lots of function definitions, I will do this in the next pr.

xlc (approved):

zqhxuyuan (approved):

PR #2171, opened 2022-05-30
no need to convert used_gas into U256

ermalkaleci:


ermalkaleci:
this will require versioning

PR #2174, opened 2022-05-31
extra precompile filter

ermalkaleci:


runtime/common/src/precompile/mod.rs
```diff
@@ -223,6 +223,16 @@ where
 		if !self.is_precompile(address) {
 			return None;
 		}
+
+		// Filter known precompile addresses except Ethereum officials
+		if address > BLAKE2F && context.address != address {
```
xlc, line 228: better to add a new const ETH_PRECOMPILE_END

xlc (commented):
should have some tests to verify

xlc (approved):

zjb0807 (approved):

PR #2191, opened 2022-06-10
add Homa to ProxyType

wangjj9219:


xlc (commented):
how about `fast_match_redeems` and `claim_redemption`?

wangjj9219:
> how about `fast_match_redeems` and `claim_redemption`?

`fast_match_redeems` and `claim_redemption` can be called by other accounts

xlc (approved):

PR #2192, opened 2022-06-10
Add Incentives Precompile

ferrell-code:


zjb0807 (approved):
LGTM. Update the predeploy-contracts.

runtime/common/src/precompile/incentives.rs
```diff
@@ -0,0 +1,690 @@
...
+					Runtime::AccountId,
+					Balance,
+					CurrencyId,
+					PoolId,
+				>>::get_pending_rewards(pool_id, who, reward_currency_ids);
+
+				Ok(PrecompileOutput {
+					exit_status: ExitSucceed::Returned,
+					cost: gas_cost,
+					output: Output::default().encode_u128_array(value),
```
ermalkaleci, line 240: value is not a fixed array but `encode_u128_array` will encode a fixed array without any size or offset which will be impossible to decode to the correct length. So `encode_u128_array` need to be updated to handle dynamic arrays. Same story for `encode_address_array` which is used by stableAsset precompile.

ermalkaleci: I've opened a PR #2196

zjb0807 (approved):

PR #2199, opened 2022-06-15
Add bifrost

zjb0807:


xlc (commented):
can you make a new config `karura-bifrost.yml` so we have one just karura, one for karura + bifrost

xlc (approved):

PR #2201, opened 2022-06-15
if transfer reward failed, put back the rewards into pool

wangjj9219:


modules/incentives/src/lib.rs
```diff
@@ -317,10 +317,25 @@ pub mod module {
...
-				// affect the process, ignore the result to continue. if it fails, just the user will not
-				// be rewarded, there will not increase user balance.
-				T::Currency::transfer(currency_id, &Self::account_id(), &who, actual_amount)?;
+				let res = T::Currency::transfer(currency_id, &Self::account_id(), &who, actual_amount);
+				if let Err(e) = res {
+					log::warn!(
+						target: "incentives",
+						"transfer: failed to transfer {:?} {:?} from {:?} to {:?}: {:?}. \
+						This is unexpected but should be safe",
+						actual_amount, currency_id, Self::account_id(), who.clone(), e
```
xlc, line 326:
```suggestion
						actual_amount, currency_id, Self::account_id(), &who, e
```

xlc: this is not really unexpected that now we know what could trigger this, so better to add a comment to explain

xlc (approved):

PR #2211, opened 2022-06-16
Mandala to compile runtime-benchmarks

ferrell-code:


Makefile
```diff
@@ -85,6 +85,12 @@ check-all: check-runtimes check-benchmarks check-integration-tests
 .PHONY: check-runtimes
 check-runtimes:
 	SKIP_WASM_BUILD= cargo check --features with-all-runtime --tests --all
+	SKIP_WASM_BUILD= cargo check -p mandala-runtime --features "runtime-benchmarks try-runtime with-ethereum-compatibility full-metadata-docs on-chain-release-build" --tests
+	SKIP_WASM_BUILD= cargo check -p mandala-runtime --features disable-runtime-api
+	SKIP_WASM_BUILD= cargo check -p karura-runtime --features "runtime-benchmarks try-runtime full-metadata-docs on-chain-release-build" --tests
```
xlc, line 89: I don't think we need `full-metadata-docs`

ferrell-code: Yeah I just made it exhaustive, I'm not sure we really need `on-chain-release-build` either. 

Also I'm not sure we even need to support `disable-runtime-api` in our runtimes either, seems more like a cumulus testing thing, but I guess its best to mirror polkadot in case it comes in handy

xlc: `on-chain-release-build` removes log messages and it is used for prod wasm build so we want to ensure it works
`disable-runtime-api` I think the main purpose is so that it can be used by e2e tests along with other parachain runtime without having conflicts

xlc (approved):

PR #2216, opened 2022-06-18
stableAsset precompile fixes

ermalkaleci:


runtime/common/src/precompile/stable_asset.rs
```diff
@@ -394,6 +421,18 @@ mod tests {
 			"};
 			assert_eq!(resp.exit_status, ExitSucceed::Returned);
 			assert_eq!(resp.output, expected_output.to_vec());
+
+			// empty output if pool doesn't exists
+
+			// getStableAssetPoolTokens(uint32) -> 0x7172c6aa
```
ferrell-code, line 427: comment has different function selector

runtime/common/src/precompile/stable_asset.rs
```diff
@@ -86,115 +86,140 @@ where
...
+							.map(|x| InputPricer::<Runtime>::read_currency(*x))
+							.sum::<u64>(),
+					);
+					// make sure there's enough gas
+					if let Some(gas_limit) = target_gas {
+						if gas_limit < gas_cost {
+							return Err(PrecompileFailure::Error {
+								exit_status: ExitError::OutOfGas,
+							});
 						}
```
ferrell-code, line 105: Could move this logic to `Pricer`, I think overlay would make it take the same amount of time. This way is fine too

ferrell-code (approved):

runtime/common/src/precompile/stable_asset.rs
ferrell-code: Yeah make sense, would iterate over twice

xlc (approved):
make sure we have docs on the contract side indicates those behaviour

zjb0807 (approved):
LGTM. 
I think `output: Default::default()` is better. Can you update others in #2152?

ermalkaleci:
> LGTM.
> I think `output: Default::default()` is better. Can you update others in #2152?

I will have a look. The problem is we can't update interfaces for without some kind of versioning

ermalkaleci:
on the interface I will have `function getStableAssetPoolTotalSupply(uint32 poolId) external view returns (bool, uint256)` where bool indicates if pool exists

xlc:
stable asset contract is not yet deployed to Karura / Acala so it is not too late to change it

Issue #2137, opened 2022-05-19
Offchain worker to automatically rebalance swap pools


xlc:
New offchain worker that try to identify unbalanced swap pools and making swaps to balance them. The profit goes to treasury & collator.

Basically we want to identify swap paths A -> B -> C -> A that resulting more A in the end.

New unsigned tx to execute the swap. Requires transaction source to be `InBlock` or `Local`.

PR #2219, opened 2022-06-20
Dex offchain triangle swap

zqhxuyuan:
closes: https://github.com/AcalaNetwork/Acala/issues/2137

### DIFF
--- a/modules/dex/src/lib.rs
+++ b/modules/dex/src/lib.rs
@@ -35,6 +35,7 @@
 
 use codec::MaxEncodedLen;
 use frame_support::{log, pallet_prelude::*, transactional, PalletId};
+use frame_system::offchain::{SendTransactionTypes, SubmitTransaction};
 use frame_system::pallet_prelude::*;
 use orml_traits::{Happened, MultiCurrency, MultiCurrencyExtended};
 use primitives::{Balance, CurrencyId, TradingPair};
@@ -52,6 +53,7 @@ mod tests;
 pub mod weights;
 
 pub use module::*;
+use orml_utilities::OffchainErr;
 pub use weights::WeightInfo;
 
 /// Parameters of TradingPair in Provisioning status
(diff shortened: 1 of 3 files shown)


modules/dex/src/lib.rs
```diff
@@ -843,6 +921,57 @@ pub mod module {
 
 			Ok(())
 		}
+
+		#[pallet::weight(1000)]
+		#[transactional]
+		pub fn triangle_swap(
```
xlc, line 927: call it `RebalanceSwap` instead and rename the other triangle to rebalance

modules/dex/src/lib.rs
```diff
@@ -851,6 +980,198 @@ impl<T: Config> Pallet<T> {
...
+			TradingPairStatuses::<T>::iter()
+				.filter(|(_, status)| status.enabled())
+				.for_each(|(pair, _)| {
+					trading_pair_values_map
+						.entry(pair.first())
+						.or_insert_with(Vec::<CurrencyId>::new)
+						.push(pair.second());
+				});
+			for (currency_id, trading_tokens) in trading_pair_values_map {
+				let _ = TradingPairNodes::<T>::try_mutate(currency_id, |maybe_trading_tokens| -> DispatchResult {
```
xlc, line 1026: I don't think you can mutate storages in offchain worker? at least the storages won't be saved. you need to use offchain worker db for that

xlc (commented):
I am thinking if we want to implement this in aggregated-dex so it can also rebalance pools with stable asset pools.

modules/dex/src/lib.rs
zqhxuyuan: `TradingPairNodes` type is CurrencyId->[CurrencyId], using offchain db I think it's too complicate. Current I'm moving this storage mutate part to `on_idle`.

zqhxuyuan: remove `TradingPairNodes`, and calculation without on chain storage.

Issue #2190, opened 2022-06-09
Transaction payment suport Aggregated DEX


zqhxuyuan:
Current, transaction payment is depend on DEX module, we should support aggregated dex in some cases. i.e.
- transfer back ERC20 to our chain, if recipient has not enough native token to pay for storage fee, user may want to use ERC20 to pay storage fee. but ERC20 only exist in aggregrated dex instead of our DEX.
- etc..

zqhxuyuan:
duplicates: https://github.com/AcalaNetwork/Acala/issues/1980

Issue #1980, opened 2022-03-21
Transaction payment support dex aggregator interface


xlc:
Update transaction payment pallet to use the dex aggregator interface from #1979.

This will allow tx fee pool to be created with stable assets pool.

Depends on #1979

PR #2237, opened 2022-06-30
Payment support aggregated dex

zqhxuyuan:
TransactionPayment:
- [x] use `Swap` instead of `DEX`, Don't need store swap path list anymore, `Swap::swap()` only need supply_currency_id, and target_currency_id is always NativeToken. so `enable_charge_fee_pool` don't need `fee_swap_path` parameter anymore.
- [x] add `with_fee_aggregated_path`
- [x] remove `GlobalFeeSwapPath` related operation, but keep it for storage migration. after migration, we can delete it.
- [x] `with_fee_currency` support two case: token not in charge fee pool use direct swap, else use charge fee pool mechanism.

AggregatedDex:
- [x] add `swap_by_path` to `Swap` trait, it's used by `with_fee_path` or user has `AlternativeFeeSwapPath`. *it's only work in Dex mode.*
- [x] add `swap_by_aggregated_path` to `Swap` trait, *it's only work in aggregated mode.*

|swap type|swap_by_path(swap_path: &[CurrencyId])|swap_by_aggregated_path(swap_path: &[AggregatedSwapPath])|
|----------|-------------------------------------------|-----------------------------------------------------------------|
|DexSwap|DEX::swap_with_specific_path()|Error|
|TaigaSwap|Error|Error|
|EitherDexOrTaigaSwap|DexSwap::swap_by_path()|Error|
|**AggregatedSwap**|swap_by_aggregated_path(AggregatedSwapPath::Dex(swap_path))|Pallet::do_aggregated_swap()

Misc:
- [x] add KUSD/AUSD for `AlternativeSwapPathJointList` to `module_aggregated_dex` module config.
- [ ] ~~evm storage charge fee support erc20 if it can be swap to native token.~~
    - we need a way to manual passing erc20 token to `reserve_storage`, but due to we can't distinguish an erc20 contract or a normal contract now, we may need another way to support this feature.
- [x] runtime upgrade
    - [x] disable and enable fee pool for existing token
    - [ ] use `enable_charge_fee_pool()` to support USDT/USDC swap with KAR in runtime upgrade? this need aggregated-dex setting too, do we need governance vote, or just enable it here? i.e. 
        - Taiga(USDT, AUSD), Dex(AUSD, KAR) to support USDT as fee token
        - Taiga(USDC, AUSD), Dex(AUSD, KAR) to support USDC as fee token

closes: https://github.com/AcalaNetwork/Acala/issues/2190
closes: https://github.com/AcalaNetwork/Acala/issues/1980


xlc:
the aggregated-dex pallet already handles the swap path so we shouldn't need to configured it again in the payment pallet

zqhxuyuan:
Yeah, I'm tring remove swap path related storage/operation, as `Swap::swap` interface only need `supply_currency_id`(the target_currency_id is always native token). payment module now only need tokens instead of swap path.

modules/transaction-payment/src/lib.rs
```diff
@@ -807,18 +805,15 @@ where
 				let fee = Self::check_native_is_not_enough(who, fee, reason).map_or_else(|| fee, |amount| amount);
 				let custom_fee_surplus = T::CustomFeeSurplus::get().mul_ceil(fee);
 				let custom_fee_amount = fee.saturating_add(custom_fee_surplus);
-				T::DEX::swap_with_specific_path(
+				T::Swap::swap(
```
xlc, line 808: we still want people to specify path here, so for example it could have better slippage compare to registered path

runtime/acala/src/lib.rs
```diff
@@ -1755,24 +1755,53 @@ pub type Executive = frame_executive::Executive<
...
-					(1, Parachain(parachains::statemint::ID)).into(),
-				)),
-				Some(4_000_000_000),
-				Some(50_000_000),
-			)],
-		);
+		let poo_size = 5 * dollar(ACA);
+		let threshold = Ratio::saturating_from_rational(1, 2).saturating_mul_int(dollar(ACA));
+		for token in FeeTokens::get() {
+			let _ = module_transaction_payment::Pallet::<Runtime>::disable_pool(token);
```
xlc, line 1770: why do we need to reinitialize the pool?

modules/transaction-payment/src/lib.rs
zqhxuyuan: If so, either introduce DEX back or IMO we should add a method which support swap path to `Swap` interface.

runtime/acala/src/lib.rs
zqhxuyuan: `initialize_pool` method now don't store `GlobalFeeSwapPath` anymore, reinitialize pool here is prepare for migration. after migration we can remove `GlobalFeeSwapPath`.

modules/transaction-payment/src/lib.rs
xlc: yeah let's go with second option

modules/support/src/dex.rs
```diff
@@ -85,6 +85,12 @@ pub trait Swap<AccountId, Balance, CurrencyId> {
 		target_currency_id: CurrencyId,
 		limit: SwapLimit<Balance>,
 	) -> Result<(Balance, Balance), DispatchError>;
+
+	fn swap_by_path(
+		who: &AccountId,
+		swap_path: &[CurrencyId],
```
xlc, line 91: should we make it `&[SwapPath]` instead?

xlc (commented):
I think we also want a `with_fee_aggregated_path` that takes `Vec<SwapPath>` so people can specify path involved with stable swap.

The current `with_fee_path` can convert the `Vec<CurrencyId>` to `Vec<SwapPath>` and call `with_fee_aggregated_path`

modules/aggregated-dex/src/tests.rs
```diff
@@ -348,6 +375,20 @@ fn taiga_swap_swap_work() {
 			),
 			Error::<Runtime>::CannotSwap
 		);
+
+		assert_eq!(
+			TaigaSwap::<Runtime>::swap_by_path(&ALICE, &vec![DOT, LDOT], SwapLimit::ExactTarget(1_000_000_000u128, 0))
+				.is_ok(),
+			false
+		);
```
shaunxw, line 383: why not using `assert_noop`?

runtime/acala/src/lib.rs
```diff
@@ -1755,24 +1755,53 @@ pub type Executive = frame_executive::Executive<
...
-			Origin::root(),
-			vec![(
-				module_xcm_interface::XcmInterfaceOperation::ParachainFee(Box::new(
-					(1, Parachain(parachains::statemint::ID)).into(),
-				)),
-				Some(4_000_000_000),
-				Some(50_000_000),
-			)],
-		);
+		let poo_size = 5 * dollar(ACA);
```
ferrell-code, line 1767:
```suggestion
		let pool_size = 5 * dollar(ACA);
```

lol

ferrell-code (commented):
Changes look good to me

Transaction Payment benchmarks will need to be updated for worst case as well as others, perhaps easier for us to do in follow up PR as linked in https://github.com/AcalaNetwork/Acala/pull/2271

Overall with the DEX aggregator there is a certain point at which we can only check so many paths before it becomes too computationally heavy. 

Its nice we can default common swap paths on chain, but also allow for off-chain calculations of best path so its possible to scale to a large number of tokens supported. We probably need to eventually build some acala.js library for getting best swap path to be able to scale to a large number of tokens

xlc (approved):

PR #2241, opened 2022-06-30
Support transfer back statemint assets

zqhxuyuan:
~~when statemint assets is ready to transfer asset to acala, then we need to~~ enable this feature to transfer back asset back to statemint.

- [x] acala runtime upgrade initialize XcmInterface setting for ParachainFee.
- [x] change karura runtime ParachainFee fee value by calculate `WeightToFee` instead of `IdentityFee`


StrawberryFlavor:
nice, verified

xlc (approved):

PR #2246, opened 2022-07-05
Add set_bridged_stable_coin_address

zjb0807:
The PR will disable honzon-bridge. After call `set_bridged_stable_coin_address` to set the address, the `honzon-bridge` will be enable.


xlc (approved):

xlc:
This breaks transaction_version. We should always put new extrinsic in the end.

PR #2250, opened 2022-07-07
Honzon Liquidation penalty goes to OnFeeDeposit

syan095:
* Base branch is liquidation_improvement(https://github.com/AcalaNetwork/Acala/pull/2164) merged into fee_distribution.
* Honzon 's liquidation penalty from all sources (liquidate via DEX, LP withdrawal, Contract and Auction) goes to OnFeeDeposit.
* Liquidation's Base and Penalty amount is passed throughout the liquidation process.
* Added tests accordingly.
* After liquidate via contract, fund is transferred to the treasury to repay the debt

Should wait until liquidation_improvement is merged into master, and master is merged into fee_distribution


modules/auction-manager/src/lib.rs
```diff
@@ -82,42 +86,49 @@ pub struct CollateralAuctionItem<AccountId, BlockNumber> {
...
-	/// Target sales amount of this auction
+	/// Base sales amount of this auction
 	/// if zero, collateral auction will never be reverse stage,
 	/// otherwise, target amount is the actual payment amount of active
 	/// bidder
 	#[codec(compact)]
-	target: Balance,
+	base: Balance,
+	/// Penalty amount as part of the auction.
+	#[codec(compact)]
```
xlc, line 96: We shouldn't really mark those compact if they are not part of extrinsic parameters. Given we need a migration, now is good time to remove all the compact here.

modules/auction-manager/src/lib.rs
```diff
@@ -603,26 +616,36 @@ impl<T: Config> Pallet<T> {
...
 				collateral_auction.amount.saturating_sub(actual_supply_amount),
 			);
+
+			// Since liquidation is done through DEX, return the bid.
 			Self::try_refund_bid(&collateral_auction, last_bid);
 
-			// Note: for StableAsset, the swap of cdp treasury is always on `ExactSupply`
-			// regardless of this swap_limit params. There will be excess stablecoins that
-			// need to be returned to the refund_recipient from cdp treasury account.
+			Self::deposit_event(Event::DEXTakeCollateralAuction {
```
xlc, line 640: why move this?

modules/support/src/honzon.rs
```diff
@@ -62,6 +62,32 @@ impl<AccountId, CurrencyId, Balance: Default, DebitBalance> RiskManager<AccountI
...
+pub trait OnLiquidationSuccess<AccountId> {
+	fn on_liquidate_success(
+		who: &AccountId,
+		currency_id: CurrencyId,
+		collateral_amount: Balance,
+		actual_collateral_consumed: Balance,
+		stable_base_amount: Balance,
+		stable_penalty_amount: Balance,
+		actual_stable_amount: Balance,
+	) -> DispatchResult;
```
xlc, line 74: I don't think this needs to return a Result.

PR #2264, opened 2022-07-14
stable-asset-precompile

ukby1234:
stable-asset precompile update

### DIFF
--- a/modules/evm/src/runner/stack.rs
+++ b/modules/evm/src/runner/stack.rs
@@ -678,11 +678,7 @@ impl<'vicinity, 'config, T: Config> BackendT for SubstrateStackState<'vicinity, 
 	}
 
 	fn original_storage(&self, address: H160, index: H256) -> Option<H256> {
-		if let Some(value) = self.substate.known_original_storage(address, index) {
-			Some(value)
-		} else {
-			Some(self.storage(address, index))
-		}
+		self.substate.known_original_storage(address, index)
 	}
 
 	fn block_base_fee_per_gas(&self) -> sp_core::U256 {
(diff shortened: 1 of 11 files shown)


modules/evm/src/runner/stack.rs
```diff
@@ -678,11 +678,7 @@ impl<'vicinity, 'config, T: Config> BackendT for SubstrateStackState<'vicinity,
 	}
 
 	fn original_storage(&self, address: H160, index: H256) -> Option<H256> {
-		if let Some(value) = self.substate.known_original_storage(address, index) {
-			Some(value)
-		} else {
-			Some(self.storage(address, index))
-		}
+		self.substate.known_original_storage(address, index)
 	}
```
ferrell-code, line 682: Any particular reason for this? Would change the behavior, although I do find it a bit wierd that this function returns an `Option<H256>` when it has no branch that could return `None`

frankyin-factual: That’s generated by cargo clippy, so this might be equivalent. If not, I am happy to revert those unrelated changes.

ferrell-code: Yeah the change seems reasonable, but will change the behavior to return `None` rather then `Some(H256::default())`. Maybe jingbo will have insight into where this function is used because I can't quite figure that out

xlc: This looks like a false positive from cargo clippy. The new code is clearly not same as the old one. Please revert.

zjb0807: What's your rust version? I use `nightly-2022-07-14` but can't reproduce the Clippy warning.

ukby1234: I was on `rustc 1.62.0-nightly (8f36334ca 2022-04-06)`. Anyway, gonna revert those clippy changes.

zjb0807 (commented):
LGTM. Please add unit tests

ukby1234:
> LGTM. Please add unit tests

tests added

ferrell-code:
LGTM, I'll let bryan/Jingbo take a final look

xlc (approved):

PR #2273, opened 2022-07-21
Add TAI & TAP for xcm

zjb0807:


xlc (dismissed):

shaunxw (dismissed):

zqhxuyuan:
We need add trader config item too, or do we want it be registered as ForeignAsset?

xlc:
Yeah let's add to trader config for now. Later once those tokens are listed on swap then we can enable tx fee pool.

zjb0807:
`aUSD:DOT = 40:1`
`kUSD:KSM = 400:1`
should we adjust the `Trader` fee?

zqhxuyuan:
> `aUSD:DOT = 40:1` `kUSD:KSM = 400:1` should we adjust the `Trader` fee?

DOT/KSM/aUSD is prior processed by `BuyWeightRateOfTransactionFeePool`, the config here is used for testcase mostly.

zqhxuyuan (approved):

runtime/acala/src/xcm_config.rs
```diff
@@ -135,6 +135,15 @@ parameter_types! {
 		).into(),
 		aca_per_second()
 	);
+	pub TapPerSecond: (AssetId, u128) = (
+		MultiLocation::new(
+			0,
+			X1(GeneralKey(TAP.encode())),
+		).into(),
+		// TODO: No price yet, assumed set at 4340
```
xlc, line 143: add an issue to update this later

xlc (approved):

PR #2297, opened 2022-07-28
Update karura & acala tokens

zjb0807:
https://github.com/AcalaNetwork/predeploy-contracts/pull/121


xlc (approved):
Should we move this to a new file? Also maybe convert it into a js script that load registered assets from onchain data to generate the file?

PR #2303, opened 2022-08-01
Refactor Benchmarks and update calls that use `T::Swap` (Honzon and Cdp Treasury)

ferrell-code:


runtime/mandala/src/benchmarking/prices.rs
```diff
@@ -16,28 +16,26 @@
...
-const STAKING: CurrencyId = GetStableCurrencyId::get();
-
 runtime_benchmarks! {
 	{ Runtime, module_prices }
 
 	lock_price {
 		// feed price
-		feed_price(vec![(STAKING, dollar(STAKING).into())])?;
-	}: _(RawOrigin::Root, STAKING)
+		feed_price(vec![(STABLECOIN, dollar(STABLECOIN).into())])?;
```
xlc, line 31: why change to Stablecoins? we have some special handle logic for Stablecoins in prices module so better to still use staking here

runtime/mandala/src/benchmarking/utils.rs
```diff
@@ -98,6 +107,148 @@ pub fn dollar(currency_id: CurrencyId) -> Balance {
...
+	)?;
+	Ok(())
+}
+
+pub fn register_stable_asset() -> DispatchResult {
+	let asset_metadata = AssetMetadata {
+		name: b"Token Name".to_vec(),
+		symbol: b"TN".to_vec(),
+		decimals: 12,
+		minimal_balance: 1,
```
xlc, line 149:
```suggestion
		minimal_balance: 1_000_000_000,
```
so it is bit more realistic

runtime/mandala/src/benchmarking/prices.rs
ferrell-code: I actually kept it the same look at line 26

```rust

const STAKING: CurrencyId = GetStableCurrencyId::get();
```

I can change it to staking cause yeah using stablecoin seems a bit nonsensical

xlc: oh ok that must be a mistake 😓

xlc (approved):

PR #2310, opened 2022-08-03
Honzon Precompile to output all collateral parameters

### LINKED ISSUE #2298: Make Precompile Benchmarks Overlay Aware
Motivation: For honzon precompile I thought it was best to have individual getters for `RiskManagmentParams` fields. seen here: https://github.com/AcalaNetwork/Acala/blob/8e92786dca81cf5d6c2c6a8cabeed923471cdd5c/modules/honzon/src/lib.rs#L430

We probably want to add getters for the rest of the fields to Honzon precompile https://github.com/AcalaNetwork/Acala/blob/8e92786dca81cf5d6c2c6a8cabeed923471cdd5c/modules/cdp-engine/src/lib.rs#L83

If a smart contract calls into 2 of the getters then they will be charged 2 reads rather than 1. Is there a way to check storage overlay so we don't double count reads. Do we care?

ferrell-code:
closes #2298

Note this will be breaking change to honzon predeploy

### DIFF
--- a/modules/honzon/Cargo.toml
+++ b/modules/honzon/Cargo.toml
@@ -9,6 +9,7 @@ serde = { version = "1.0.136", optional = true }
 codec = { package = "parity-scale-codec", version = "3.0.0", default-features = false }
 scale-info = { version = "2.1", default-features = false, features = ["derive"] }
 sp-runtime = { git = "https://github.com/paritytech/substrate", branch = "polkadot-v0.9.26", default-features = false }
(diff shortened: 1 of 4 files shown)


xlc (approved):

ermalkaleci:
@ferrell-code we can't remove existing method, we need to add it as new method

ferrell-code:
@ermalkaleci its not deployed anywhere except mandala, so I think removing it is fine

Issue #2254, opened 2022-07-08
honzon-distribution pallet


xlc:
- Types
  - enum DistributionDestination
    - StableAsset { poolId: StableAssetPoolId }
  - struct DistributionParams
    - cap: Balance
    - maxStep: Balance
    - targetMin: Ratio
    - targetMax: Ratio
- Config
  - type UpdateOrigin: EnsureOrigin
  - type AdjustPeriod: Get\<(BlockNumber, BlockNumber)>
- Storages
  - Params: map DistrubutionDestination => DistributionParams
  - Amount: map DistrubutionDestination => Balance
- Calls
  - fn update_params(origin, params: Option\<DistributionParams>)
    - ensure UpdateOrigin
  - fn force_adjust(origin, dest: DistributionDestination)
    - ensure UpdateOrigin
    - apply distrubution adjustment
- Pallet
  - fn on_initialize
    - if now % AdjustPeriod.0 == AdjustPeriod.1
      - for (dest, params) in Params
        - Check dest value
        - if lower than min target
          - Mint & add liqudiity
        - if higher than max target
          - Withdraw liquidity and burn

Issue #368, opened 2020-08-11
Auction Manager avoid special case handling for is_zero


xlc:
There are many `if value.is_zero()` check in auction manager. The logic can be abstracted into struct method to make sure we don't forget check is_zero somewhere.

https://github.com/AcalaNetwork/Acala/pull/356/files#diff-92e2cb9814da46b1efd7bcd8b3868948R68

PR #2312, opened 2022-08-05
honzon distribution

zqhxuyuan:
closes: https://github.com/AcalaNetwork/Acala/issues/2254


modules/honzon-distribution/src/lib.rs
```diff
@@ -0,0 +1,329 @@
...
+		},
+	}
+
+	#[pallet::pallet]
+	pub struct Pallet<T>(_);
+
+	#[pallet::hooks]
+	impl<T: Config> Hooks<T::BlockNumber> for Pallet<T> {
+		fn on_initialize(now: T::BlockNumber) -> Weight {
+			if now % T::AdjustPeriod::get() == Zero::zero() {
```
xlc, line 145: I want to make it `now % period === offset` so that we can purposely ensure the `on_initialize` hooks for different pallets won't be all scheduled on a same block

modules/honzon-distribution/src/lib.rs
```diff
@@ -0,0 +1,329 @@
...
+			if let Some(max_step) = max_step {
+				params.max_step = max_step;
+			}
+			if let Some(target_min) = target_min {
+				params.target_min = target_min;
+			}
+			if let Some(target_max) = target_max {
+				params.target_max = target_max;
+			}
+			DistributionDestinationParams::<T>::insert(destination.clone(), params.clone());
```
xlc, line 186:
```suggestion
			DistributionDestinationParams::<T>::insert(&destination, &params);
```

modules/honzon-distribution/src/lib.rs
```diff
@@ -0,0 +1,329 @@
...
+
+			Self::deposit_event(Event::<T>::UpdateDistributionParams {
+				destination: destination.clone(),
```
xlc, line 189:
```suggestion
				destination,
```

modules/honzon-distribution/src/lib.rs
```diff
@@ -0,0 +1,329 @@
...
+
+			Self::do_adjust_to_destination(destination)?;
+			Ok(())
+		}
+	}
+}
+
+impl<T: Config> Pallet<T> {
+	pub fn do_adjust_to_destination(destination: DistributionDestination<T::AccountId>) -> DispatchResult {
+		let params = DistributionDestinationParams::<T>::get(destination.clone())
```
xlc, line 212:
```suggestion
		let params = DistributionDestinationParams::<T>::get(&destination)
```

modules/honzon-distribution/src/lib.rs
```diff
@@ -0,0 +1,329 @@
...
+		destination: DistributionDestination<T::AccountId>,
+		stable_asset: DistributionToStableAsset<T::AccountId>,
+		params: DistributionParams,
+	) -> Result<Amount, DispatchError> {
+		let pool_id = stable_asset.pool_id;
+		let pool_info = T::StableAsset::pool(pool_id).ok_or(Error::<T>::InvalidDestination)?;
+		let account_id = stable_asset.account_id;
+
+		let total_supply = pool_info.total_supply;
+		let ausd_supply = pool_info.balances[stable_asset.stable_token_index as usize];
```
xlc, line 257: use `.get` to avoid panic on out of bound access

modules/honzon-distribution/src/lib.rs
```diff
@@ -0,0 +1,329 @@
...
+		let remain_rate = one.saturating_sub(target_rate);
+		let remain_reci = one.div(remain_rate);
+		if current_rate < params.target_min {
+			let numerator = target_rate.saturating_mul_int(total_supply).saturating_sub(ausd_supply);
+			let mint_amount = remain_reci.saturating_mul_int(numerator).min(params.max_step);
+
+			// should failed if after this mint, the balance of distributed exceed the capacity.
+			if let Some(exist_minted) = DistributedBalance::<T>::get(&destination) {
+				let newly_amount = mint_amount.saturating_add(exist_minted);
+				ensure!(newly_amount < params.capacity, Error::<T>::ExceedCapacity);
```
xlc, line 275: no need to fail, just cap to the reaming amount

modules/honzon-distribution/src/lib.rs
```diff
@@ -0,0 +1,329 @@
...
+				amount: burn_amount,
+			});
+			return Ok(burn_amount);
+		}
+
+		Ok(0 as Amount)
+	}
+}
+
+pub struct RebasedStableAssetErrorConvertor<T>(PhantomData<T>);
```
xlc, line 321: this is only used for test? should move to mock

modules/honzon-distribution/src/mock.rs
```diff
@@ -0,0 +1,275 @@
...
+	fn convert_balance(balance: Balance, _asset_id: CurrencyId) -> Balance {
+		balance
+	}
+
+	fn convert_balance_back(balance: Balance, _asset_id: CurrencyId) -> Balance {
+		balance
+	}
+}
+
+pub type StableAssetWrapper =
```
xlc, line 172: I don't think this is necessary for mock

modules/honzon-distribution/src/mock.rs
```diff
@@ -0,0 +1,275 @@
...
+	pub enum Runtime where
+		Block = Block,
+		NodeBlock = Block,
+		UncheckedExtrinsic = UncheckedExtrinsic
+	{
+		System: frame_system,
+		Balances: pallet_balances,
+		Tokens: orml_tokens,
+		Currencies: module_currencies,
+		Dex: module_dex,
```
xlc, line 232: why is dex needed?

zqhxuyuan: removed.

modules/honzon-distribution/src/mock.rs
zqhxuyuan: `StableAssetWrapper` is used to test stable asset mint and burn, it's a wrapper type.

modules/honzon-distribution/src/lib.rs
zqhxuyuan: move to module_support/mock.

modules/honzon-distribution/src/lib.rs
zqhxuyuan: What's the production runtime value should be? current I'm setting period=DAYS and offset=MINUTES.

xlc: period = 1 day, offset = 1 block

modules/honzon-distribution/src/lib.rs
```diff
@@ -0,0 +1,319 @@
...
+		ensure!(asset_index < asset_length, Error::<T>::InvalidDestination);
+
+		let current_rate = Ratio::saturating_from_rational(*ausd_supply, total_supply);
+		let target_rate = params
+			.target_min
+			.saturating_add(params.target_max)
+			.saturating_mul(Ratio::saturating_from_rational(1, 2));
+		let one: Ratio = One::one();
+		let remain_rate = one.saturating_sub(target_rate);
+		let remain_reci = one.div(remain_rate);
```
ferrell-code, line 266: `checked_div()` here, could panic

modules/honzon-distribution/src/lib.rs
```diff
@@ -0,0 +1,319 @@
...
+pub enum DistributionDestination<AccountId> {
+	StableAsset(DistributionToStableAsset<AccountId>),
+}
+
+/// Information needed when distribution to StableAsset.
+#[derive(Encode, Decode, Clone, RuntimeDebug, PartialEq, Eq, TypeInfo, MaxEncodedLen)]
+pub struct DistributionToStableAsset<AccountId> {
+	pub pool_id: StableAssetPoolId,
+	pub stable_token_index: u32,
+	pub stable_currency_id: CurrencyId,
```
ferrell-code, line 57: We could take this field and make it a config constant like `T::StableCurrencyId`. Do we want to use this pallet to mint something other than AUSD? if we do than make sense as is

modules/honzon-distribution/src/tests.rs
```diff
@@ -0,0 +1,237 @@
...
+				max_step: 0,
+				target_min: Default::default(),
+				target_max: Default::default(),
+			},
+		}));
+	});
+}
+
+#[test]
+fn stable_asset_mint_burn_works() {
```
ferrell-code, line 88: Maybe add another test with three pool to make sure it works with that case

modules/honzon-distribution/src/lib.rs
```diff
@@ -0,0 +1,319 @@
...
+				// update `DistributedBalance` of destination
+				DistributedBalance::<T>::try_mutate(destination, |maybe_balance| -> DispatchResult {
+					let old_val = maybe_balance.take().unwrap_or_default();
+					let new_val = if balance.is_positive() {
+						old_val
+							.checked_add(balance as Balance)
+							.ok_or(Error::<T>::InvalidUpdateBalance)
+					} else {
+						old_val
+							.checked_sub(balance.unsigned_abs() as Balance)
```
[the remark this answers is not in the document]
ferrell-code: nit

modules/honzon-distribution/src/tests.rs
zqhxuyuan: I'll do that in integration test instead of here.

modules/honzon-distribution/src/lib.rs
zqhxuyuan: I think it'll be only AUSD. added.

xlc (commented):
also need a `remove_distrubution(dest)` to remove the params and balance and withdraw everything

modules/honzon-distribution/src/lib.rs
```diff
@@ -0,0 +1,358 @@
...
+		let pool_info = T::StableAsset::pool(pool_id).ok_or(Error::<T>::InvalidDestination)?;
+		let account_id = stable_asset.account_id;
+		let asset_length = pool_info.assets.len();
+		let asset_index = stable_asset.stable_token_index as usize;
+		let total_supply = pool_info.total_supply;
+		let ausd_supply = pool_info
+			.balances
+			.get(asset_index)
+			.ok_or(Error::<T>::InvalidDestination)?;
+		ensure!(asset_index < asset_length, Error::<T>::InvalidDestination);
```
xlc, line 262: I don't think we need this check anymore

xlc (commented):
currently capacity is only enforced when minting but I think we should always enforce it.
i.e. if current balance is greater than capacity (must be due to we reduced the capacity via governance), then it should always redeem to reduce the current balance to match with capacity. this also means we don't need to special case `target_max = 0`

modules/honzon-distribution/src/lib.rs
ferrell-code: line #368 would panic if governance accidentally put a bad value.

I think check is still useful

modules/honzon-distribution/src/lib.rs
```diff
@@ -0,0 +1,378 @@
...
+		let capacity = params.capacity;
+
+		let current_rate = Ratio::saturating_from_rational(*ausd_supply, total_supply);
+		let one: Ratio = One::one();
+		if current_rate > params.target_max || capacity < distributed {
+			// current rate large than target_max, or capacity is less than distributed, burn aUSD
+			let burn_amount = if capacity < distributed {
+				let remain = distributed.saturating_sub(capacity).min(*ausd_supply);
+				let stable_balance = T::Currency::free_balance(StableAssetPoolToken(pool_id), &account_id);
+				remain.min(stable_balance)
```
zqhxuyuan, line 314: in some case when rate of aUSD and lp not maintain in 1. the result of `distributed-capacity` may large than lp share, we use lp share to redeem. because current `redeem_single` accept lp token amount. other case redeem works, although not strict precise, but it's acceptable.

PR #2314, opened 2022-08-06
Make close_dex benchmark Worst Case

ferrell-code:
Also added `debug_assert!` to catch invariant in homa

follow up of: #2303


modules/homa/src/lib.rs
```diff
@@ -362,6 +362,12 @@ pub mod module {
 			let bump_era_number = Self::era_amount_should_to_bump(T::RelayChainBlockNumber::current_block_number());
 			if !bump_era_number.is_zero() {
 				let _ = Self::bump_current_era(bump_era_number);
+				debug_assert!(
```
xlc, line 365: use `debug_assert_eq `

xlc (dismissed):

xlc (approved):

PR #2315, opened 2022-08-08
Immediate liquidation

shaunxw:


modules/cdp-engine/src/lib.rs
```diff
@@ -1334,47 +1329,127 @@ impl<T: Config> LiquidateCollateral<T::AccountId> for LiquidateViaDex<T> {
...
+				target_stable_amount,
+			);
+			let liquidated_price = Price::saturating_from_rational(repayment, amount);
+			let ratio = liquidated_price
+				.checked_div(&collateral_oracle_price)
+				.ok_or(ArithmeticError::DivisionByZero)?;
+			return Ok(ratio);
+		} else if repayment > 0 {
+			// insufficient repayment, refund
+			CurrencyOf::<T>::transfer(stable_coin, &repay_dest_account_id, &contract_account_id, repayment)?;
```
xlc, line 1389: I think we can just revert the previous liquidate call using `with_tranasction` so no need to refund

PR #2316, opened 2022-08-08
Support LCDOT for XCM

zjb0807:


modules/asset-registry/src/lib.rs
```diff
@@ -582,6 +582,27 @@ where
...
+{
+	fn calculate_rate(location: MultiLocation) -> Option<Ratio> {
+		let currency = key_to_currency(location);
+		match currency {
+			Some(CurrencyId::LiquidCrowdloan(lease)) => {
+				// The default rate for all leases is the same.
+				// LCDOT:DOT = 1.3:1
+				let rate = FixedU128::saturating_from_rational(13, 10);
+				log::debug!(target: "asset-registry::weight", "LiquidCrowdloan: {}, rate:{:?}", lease, rate);
+				Some(rate)
```
xlc, line 599: this should return the ratio to native token ED not DOT
we also have LCDOT registered so should just use similar logic

xlc (approved):

zqhxuyuan (approved):

shaunxw (approved):

PR #2341, opened 2022-09-11
Remove dex savings

wangjj9219:


modules/incentives/src/lib.rs
```diff
@@ -204,6 +203,14 @@ pub mod module {
 
 	#[pallet::hooks]
 	impl<T: Config> Hooks<T::BlockNumber> for Pallet<T> {
+		/// need remove it after next runtime upgrade
+		fn on_runtime_upgrade() -> Weight {
```
xlc, line 207: put this to runtime upgrade hook

runtime/common/src/precompile/incentives.rs
```diff
@@ -106,20 +106,11 @@ where
...
-				let pool_id = PoolId::Dex(pool_currency_id);
-
-				let value = <module_incentives::Pallet<Runtime> as IncentivesManager<
-					Runtime::AccountId,
-					Balance,
-					CurrencyId,
-					PoolId,
-				>>::get_dex_reward_rate(pool_id);
-
+				// NOTE: return default , or return PrecompileFailure?
```
xlc, line 109: this is not enabled so should be safe to remove this
cc @zjb0807

zjb0807: https://github.com/AcalaNetwork/Acala/blob/3f449446e4e2eea58582b958e8e1c967f301a55e/runtime/common/src/precompile/mod.rs#L136

`INCENTIVES` only enabled for mandala, it's safe

modules/incentives/src/tests.rs
```diff
@@ -924,7 +820,7 @@ fn on_initialize_should_work() {
 			TokensModule::free_balance(ACA, &VAULT::get()),
 			1300 + (1000 + 2000 + 100 + 200)
 		);
-		assert_eq!(TokensModule::free_balance(AUSD, &VAULT::get()), 509 + (500 + 9)); // 9 from debit_issue,  500 from RewardsSource
+		assert_eq!(TokensModule::free_balance(AUSD, &VAULT::get()), 500 + 500); // 9 from debit_issue,  500 from RewardsSource
```
shaunxw, line 823:
```suggestion
		assert_eq!(TokensModule::free_balance(AUSD, &VAULT::get()), 500 + 500); // 500 from RewardsSource
```

runtime/common/src/precompile/incentives.rs
xlc: let's remove this and the predeploy contract

xlc (approved):

PR #2343, opened 2022-09-16
Some improvement for security

wangjj9219:
1. use wrapped function to issue/burn for some assets instead of using Tokens.deposit/withdraw directly
2. config homa.LastEraBumpedBlock shouldn't cause immediately bump era


shaunxw (approved):
LGTM

PR #2345, opened 2022-09-16
Bounded params

shaunxw:
Introduce abstract `BoundedType` and `FractionalRate`. Will add other bounded types in follow up PRs.


modules/support/src/bounded.rs
```diff
@@ -0,0 +1,167 @@
...
+		let (min, max) = Range::get();
+		if value < min || value > max {
+			return Err(Error::OutOfBound);
+		}
+		Ok(Self(value, PhantomData))
+	}
+
+	/// Set the inner value. Returns `Err` if out of bound or the diff with current value exceeds
+	/// the max absolute value.
+	pub fn set(&mut self, value: T) -> Result<(), Error> {
```
xlc, line 77: `try_set`?

modules/cdp-engine/src/lib.rs
```diff
@@ -488,11 +493,21 @@ pub mod module {
 			T::UpdateOrigin::ensure_origin(origin)?;
 
 			let mut collateral_params = Self::collateral_params(currency_id).unwrap_or_default();
-			if let Change::NewValue(update) = interest_rate_per_sec {
-				collateral_params.interest_rate_per_sec = update;
+			if let Change::NewValue(maybe_rate) = interest_rate_per_sec {
+				if let (Some(existing), Some(rate)) = (collateral_params.interest_rate_per_sec.as_mut(), maybe_rate) {
```
xlc, line 497: better to make it a match

modules/support/src/bounded.rs
```diff
@@ -0,0 +1,167 @@
...
+impl<T: Default + Encode + Decode, Range: Get<(T, T)>, MaxChangeAbs: Get<T>> Default
+	for BoundedType<T, Range, MaxChangeAbs>
+{
+	fn default() -> Self {
+		let (min, _) = Range::get();
+		Self(min, PhantomData)
+	}
+}
+
+impl<T: Encode + Decode + CheckedSub + PartialOrd + Copy, Range: Get<(T, T)>, MaxChangeAbs: Get<T>>
```
xlc, line 63: should implement `TryFrom` and `Into` trait

shaunxw: `TryFrom` and `Into` cannot be implemented for bounded types, which conflict with impl in `core`, due to rust trait limitations https://github.com/rust-lang/rust/issues/50133

modules/cdp-engine/src/lib.rs
```diff
@@ -87,7 +88,7 @@ pub struct RiskManagementParams {
 	pub maximum_total_debit_value: Balance,
 
 	/// Extra interest rate per sec, `None` value means not set
-	pub interest_rate_per_sec: Option<Rate>,
+	pub interest_rate_per_sec: Option<FractionalRate>,
```
wangjj9219, line 91: the bound and change limit check of `FractionalRate` doesn't make much sense for the `interest_rate_per_sec` of `RiskManagementParams`,  the config value is too tiny(for example 0.000000000860244)

shaunxw: We can definitely further limit the range, if reasonable bounds can be defined. Any suggestion for the upper limit of `interest_rate_per_sec`?

modules/support/src/bounded.rs
```diff
@@ -0,0 +1,183 @@
...
+		let abs = if value > *old_value {
+			value
+				.checked_sub(old_value)
+				.expect("greater number subtracting smaller one can't underflow; qed")
+		} else {
+			old_value
+				.checked_sub(&value)
+				.expect("greater number subtracting smaller one can't underflow; qed")
+		};
+		if abs > max_change_abs {
```
wangjj9219, line 103: we may want to directly set the value to min for some situations, but if the change exceeds the  max_change_abs when try_set min, whether to bypass this check or not? @xlc

shaunxw: In this case you can make multiple calls in a batch, to gradually set the value until min is reached.

modules/cdp-engine/src/lib.rs
wangjj9219: we can define reasonable bound for system params in #2362

xlc (approved):

Issue #2342, opened 2022-09-15
Not check relayNumber when the evm.chainId is mandala/karura-testnet/acala-testnet


zjb0807:
https://github.com/AcalaNetwork/Acala/blob/ed9f42c1f9f202cabf68fd23c0c8a5219ed38bbc/runtime/acala/src/lib.rs#L1458
https://github.com/AcalaNetwork/Acala/blob/ed9f42c1f9f202cabf68fd23c0c8a5219ed38bbc/runtime/karura/src/lib.rs#L1474
https://github.com/AcalaNetwork/Acala/blob/ed9f42c1f9f202cabf68fd23c0c8a5219ed38bbc/runtime/mandala/src/lib.rs#L1691

PR #2346, opened 2022-09-16
Skip checking relay chain block numbers on testnet

zjb0807:
Closes: #2342

### DIFF
--- a/runtime/common/src/lib.rs
+++ b/runtime/common/src/lib.rs
@@ -22,6 +22,7 @@
 #![recursion_limit = "256"]
 
 use codec::{Decode, Encode, MaxEncodedLen};
+use cumulus_pallet_parachain_system::CheckAssociatedRelayNumber;
 use frame_support::{
 	parameter_types,
 	traits::{Contains, EitherOfDiverse, Get},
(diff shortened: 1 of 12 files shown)


runtime/common/src/lib.rs
```diff
@@ -133,6 +138,28 @@ impl Convert<Weight, u64> for WeightToGas {
 	}
 }
 
+pub struct CheckRelayNumber<EvmChainID, RelayNumberStrictlyIncreases>(EvmChainID, RelayNumberStrictlyIncreases);
+impl<EvmChainID: Get<u64>, RelayNumberStrictlyIncreases: CheckAssociatedRelayNumber> CheckAssociatedRelayNumber
+	for CheckRelayNumber<EvmChainID, RelayNumberStrictlyIncreases>
+{
+	fn check_associated_relay_number(current: RelayChainBlockNumber, previous: RelayChainBlockNumber) {
+		if EvmChainID::get() == CHAIN_ID_MANDALA
```
xlc, line 146: make this a match case

xlc (commented):
some test to confirm will be nice

xlc (approved):

runtime/integration-tests/src/runtime.rs
```diff
@@ -341,6 +341,29 @@ fn parachain_subaccounts_are_unique() {
...
+		set_relaychain_block_number(10);
+		assert_eq!(ParachainSystem::validation_data().unwrap().relay_parent_number, 10);
+
+		// testnet skip checking relay chain block number
+		set_relaychain_block_number(9);
+		assert_eq!(ParachainSystem::validation_data().unwrap().relay_parent_number, 9);
+
+		// set chain_id mainnet
+		#[cfg(feature = "with-mandala-runtime")]
+		module_evm::ChainId::<Runtime>::set(CHAIN_ID_KARURA_MAINNET);
```
shaunxw, line 358: it should be mandala chain ID.

zjb0807: only karura-mainnet and acala-mainnet check relay chain block number, so use karura mainnet chainid with mandala runtime

shaunxw: can you add this as comment in the code?

shaunxw (approved):

PR #2347, opened 2022-09-16
add option to pause evm precompiles

ermalkaleci:


zjb0807 (commented):
LGTM.
Do we need to suspend the functionality of the contract addresses that has loophole?

ermalkaleci:
> LGTM.
> 
> Do we need to suspend the functionality of the contract addresses that has loophole?

I don't think we should. If the team wants to pause something they can add the feature into their contract

xlc (approved):

PR #2357, opened 2022-09-26
reset incentives.PendingMultiRewards and rewards.SharesAndWithdrawnRewards for Pool aUSD-IBTC

wangjj9219:


xlc (dismissed):

wangjj9219:
I think this migration should reset rewards.PoolInfos for pool aUSD-IBTC, I will update code

zjb0807 (dismissed):

modules/incentives/src/migration.rs
```diff
@@ -19,29 +19,91 @@
...
+
+		let mut total_share: Balance = Default::default();
+
+		// reset SharesAndWithdrawnRewards
+		for (who, (share, _)) in orml_rewards::SharesAndWithdrawnRewards::<T>::iter_prefix(&pool_id) {
+			orml_rewards::SharesAndWithdrawnRewards::<T>::mutate(&pool_id, &who, |(_, withdrawn_rewards)| {
+				*withdrawn_rewards = WithdrawnRewards::new();
+			});
+
+			total_share = total_share.saturating_add(share);
```
zjb0807, line 78: The WithdrawnRewards is BTreeMap, e.g.
```
{
        {token:ACA}: 2,047,606,129,239,216
        {token:AUSD}: 62,077,309,016,549,947,911
        {foreignAsset:4}: 63,305,740,673,477
}
```

But the total_share is Balance, I'm not sure the result is correct.

zjb0807: https://github.com/open-web3-stack/open-runtime-module-library/blob/master/rewards/src/lib.rs#L80

wangjj9219: here is vaule definition for `SharesAndWithdrawnRewards` : 
https://github.com/open-web3-stack/open-runtime-module-library/blob/44fda4432b71f49ee59a650cf4775db895ab97af/rewards/src/lib.rs#L110

The previous one of the tuple is `Balance` type to record the staked LP token amount,  `WithdrawnRewards` records the withdrawn rewards for calculation unsettled rewards. 

And `rewards.PoolInfos` for aUSD-IBTC has been killed in before referendum. Therefore, in addition to clearing the unsettled rewards(`WithdrawnRewards` in `SharesAndWithdrawnRewards` ) and the rewards to be claimed(incentives.PendingMultiRewards), the total amount of staked LP token should be counted and reset in `rewards.PoolInfos`.

zjb0807 (dismissed):

modules/incentives/src/migration.rs
```diff
@@ -19,29 +19,91 @@
...
+	fn post_upgrade() -> Result<(), &'static str> {
+		let pool_id = GetPoolId::get();
+		let mut total_share = Balance::default();
+
+		for (_, (share, withdrawn_rewards)) in orml_rewards::SharesAndWithdrawnRewards::<T>::iter_prefix(&pool_id) {
+			assert_eq!(withdrawn_rewards, WithdrawnRewards::new());
+			total_share = total_share.saturating_add(share);
+		}
+
+		assert_eq!(orml_rewards::PoolInfos::<T>::get(&pool_id).total_shares, total_share);
```
shaunxw, line 101: I think we also need `assert_eq` to check `rewards` of each `PoolInfo` is empty btree map.

shaunxw (approved):

runtime/karura/src/lib.rs
```diff
@@ -1799,14 +1799,8 @@ pub type SignedPayload = generic::SignedPayload<Call, SignedExtra>;
...
-pub type Executive = frame_executive::Executive<
-	Runtime,
-	Block,
-	frame_system::ChainContext<Runtime>,
-	Runtime,
-	AllPalletsWithSystem,
-	module_incentives::migration::ClearDexSavingRewardRates<Runtime>,
->;
+pub type Executive =
+	frame_executive::Executive<Runtime, Block, frame_system::ChainContext<Runtime>, Runtime, AllPalletsWithSystem, ()>;
```
xlc, line 1803: migration is not yet executed on karura so I think we should leave this until it is executed

xlc (approved):

shaunxw (approved):

PR #2381, opened 2022-10-12
Statemine more test

zqhxuyuan:
statemine now support using self sufficent asset paying for xcm fee in this PR: https://github.com/paritytech/cumulus/pull/1278.

testcase support using self sufficent asset as fee on statemine
and also support using relaychain asset as fee on statemine
in some case(e.g. statemint) user may still want to use relaychain asset as fee


runtime/integration-tests/src/relaychain/statemine.rs
```diff
@@ -74,8 +77,14 @@ fn statemine_min_xcm_fee_matched() {
 }
 
 #[test]
-fn teleport_from_relay_chain() {
+fn teleport_between_relaychain_and_statemine_works() {
```
shaunxw, line 80: Use cases between Kusama and statemine isn't in the scope of our integration tests.

runtime/integration-tests/src/relaychain/statemine.rs
```diff
@@ -87,21 +96,243 @@ fn teleport_from_relay_chain() {
...
+		assert_eq!(UNIT - FEE_STATEMINE, Balances::free_balance(&AccountId::from(BOB)));
+	});
+
+	KusamaNet::execute_with(|| {
+		assert_eq!(2 * UNIT - FEE_KUSAMA, Balances::free_balance(&AccountId::from(ALICE)));
+	});
+}
+
+#[test]
+fn reserve_transfer_between_relaychain_and_statemine_should_not_allowed() {
```
shaunxw, line 135: Same here.

shaunxw (commented):
Thanks for the PR. I think the tests between Statemine and Kusama should to be removed, as it's not related to Acala/Karura.

xlc (approved):

PR #2388, opened 2022-10-19
Removes license check in rustfmt

zjb0807:
Removes license check in rustfmt as it is deprecated now (see https://github.com/rust-lang/rustfmt/issues/5103 )

Dependent: https://github.com/open-web3-stack/open-runtime-module-library/pull/829


xlc (approved):

xlc:
Should research some other tools to do the validation

xlc (approved):

Issue #2338, opened 2022-09-06
Future rust-evm update reminder


ermalkaleci:
when we update `rust-evm` we need to handle `invalid_code` on binary search https://github.com/paritytech/frontier/pull/840

ermalkaleci:
https://github.com/rust-blockchain/evm/commit/51b8c2ce3104265e1fd5bb0fe5cdfd2e0938239c

zjb0807:
Fixed in #2406

PR #2389, opened 2022-10-19
Update evm dependencies

zjb0807:
https://github.com/rust-blockchain/evm/pull/126/files#diff-363a23d98a6c2addecef2665cf68deec178a1dc19d7b625439d4cc50f7f4af50R131

The codec index changed, and this needs to publish a new client.


ermalkaleci:
We need to update RPC as well #2338

PR #2393, opened 2022-10-27
prevent arbitrary message from xcm pallet as in statemine

dzmitry-lahoda:
i known karura disabled most of extrinsics in that pallet via filter call, but hope at some point it will consider it secure to open up it


shaunxw (approved):
Acala will also need it. Will be good to have integration tests to cover this.

xlc:
replaced by https://github.com/AcalaNetwork/Acala/pull/2399

PR #2398, opened 2022-11-09
claim_rewards should not re-accumulate failed transferred rewards

### LINKED ISSUE #2402: Emit 'ClaimRewards' event only if rewards transfer succeeded
https://github.com/AcalaNetwork/Acala/blob/9b661ac6caf8362bbe7bf9fd46ad7e17fddccc38/modules/incentives/src/lib.rs#L426-L443

should be:

```rust
    if res.is_error() {
        // ...
    } else {
        // deposit event
    }
```

wangjj9219:
1. claim_rewards should not re-accumulate failed transferred rewards.
2. transfer rewards from the source account to the vault and update the rewards record atomically when accumulating incentives.

Closes #2402


modules/incentives/src/lib.rs
```diff
@@ -380,89 +380,113 @@ impl<T: Config> Pallet<T> {
...
+							);
+						}
+					};
+				}
+
+				// clear zero value item of BTreeMap
+				pending_multi_rewards.retain(|_, v| *v != 0);
+
+				// if pending_multi_rewards is default, clear the storage
+				if *pending_multi_rewards == BTreeMap::new() {
```
ermalkaleci, line 468:
```suggestion
				if pending_multi_rewards.is_empty() {
```

modules/incentives/src/tests.rs
```diff
@@ -418,68 +418,183 @@ fn payout_works() {
...
+			PoolId::Loans(BTC)
+		));
+
+		assert_eq!(TokensModule::free_balance(AUSD, &VAULT::get()), 1);
+		assert_eq!(TokensModule::free_balance(AUSD, &BOB::get()), 18);
+		assert_eq!(
+			RewardsModule::pool_infos(PoolId::Loans(BTC)),
+			PoolInfo {
+				total_shares: 300,
+				rewards: vec![(AUSD, (33, 33)), (DOT, (30, 30))].into_iter().collect(),
```
ermalkaleci, line 588: above claim failed so I think reward for AUSD should be `(33, 27)` cause 4 is still `pending_multi_rewards`

wangjj9219: `claim_rewards` will trigger `rewards.claim_rewards` first. 
It will settle new accumulated rewards(4 aUSD) for BOB,  then added it to the `incentives.PendingMultiRewards` of BOB (The withdrawn record of `rewards.PoolInfos` and BOB's `rewards.SharesAndWithdrawn` will also added the new settled rewards)

After settled,  will trying to payout the `PendingMultiRewards` to BOB.  But this step failed due to aUSD transfer failure,  BOB's `PendingMultiRewards` will not change at this step,  still keep 4 aUSD.

shaunxw (approved):

xlc (approved):

PR #2401, opened 2022-11-14
keep the buffer amount for debit pool  when offset_surplus_and_debit

wangjj9219:


runtime/acala/src/lib.rs
```diff
@@ -1531,6 +1531,23 @@ impl orml_unknown_tokens::Config for Runtime {
...
 
+pub struct GetTotalFrozenStableCurrency;
+impl frame_support::traits::Get<Balance> for GetTotalFrozenStableCurrency {
+	fn get() -> Balance {
+		let stable_currency_id = GetStableCurrencyId::get();
+		let mut total_frozen_stable_currency = Balance::default();
+		for (_, currency_id, orml_tokens::AccountData::<Balance> { free, frozen, .. }) in
+			orml_tokens::Accounts::<Runtime>::iter()
+		{
+			if currency_id == stable_currency_id {
```
xlc, line 1542: this iterates every account data for all users all tokens and could cause PoV issue

wangjj9219: em, the first key of `tokens.AccountData` is `AccountId` instead of `CurrencyId`, so can not iterate it by `CurrencyId` directly.
I think should iterate the `tokens.Locks` firstly to find those `AccountId`s with locked `aUSD`

xlc (approved):

xlc (approved):

PR #2417, opened 2022-11-24
Xtokens call restrict

wangjj9219:


modules/extrinsic-restrict/src/lib.rs
```diff
@@ -0,0 +1,336 @@
...
+	pub type Hihi<T: Config> = StorageMap<_, Twox64Concat, CurrencyId, XtokensTransferLimit, ValueQuery>;
+
+	#[pallet::pallet]
+	#[pallet::without_storage_info]
+	pub struct Pallet<T>(_);
+
+	#[pallet::hooks]
+	impl<T: Config> Hooks<T::BlockNumber> for Pallet<T> {
+		fn on_initialize(now: T::BlockNumber) -> Weight {
+			if now
```
xlc, line 111: we shouldn't need to clear the limit. just store the last update time is enough

modules/extrinsic-restrict/Cargo.toml
```diff
@@ -0,0 +1,52 @@
+[package]
+name = "module-extrinsic-restrict"
```
xlc, line 2: just call it `module-rate-limit`

modules/extrinsic-restrict/src/lib.rs
```diff
@@ -0,0 +1,336 @@
...
+mod mock;
+mod tests;
+pub mod weights;
+
+#[frame_support::pallet]
+pub mod module {
+	use super::*;
+
+	#[derive(PartialEq, Eq, Clone, Encode, Decode, RuntimeDebug, TypeInfo, Default)]
+	pub struct XtokensTransferLimit {
```
xlc, line 61:
```suggestion
	pub struct RateLimit {
```

no need to reference xtokens in this pallet

wangjj9219:
new impl #2419

PR #2422, opened 2022-11-29
Update github actions

zjb0807:
`
Node.js 12 actions are deprecated. For more information see: https://github.blog/changelog/2022-09-22-github-actions-all-actions-will-begin-running-on-node16-instead-of-node12/. Please update the following actions to use Node.js 16: actions-rs/toolchain@v1, jwlawson/actions-setup-cmake@v1.12, actions/setup-node@v2
`

Fix the warnings in https://github.com/AcalaNetwork/Acala/actions/runs/3569255093

`actions-rs/toolchain` is no longer maintained, switch to `dtolnay/rust-toolchain`

### DIFF
--- a/.github/workflows/coverage.yml
+++ b/.github/workflows/coverage.yml
@@ -38,15 +38,13 @@ jobs:
         with:
           submodules: recursive
       - name: Install toolchain
-        uses: actions-rs/toolchain@v1
+        uses: dtolnay/rust-toolchain@nightly
         with:
-          profile: minimal
           toolchain: nightly-2022-08-05
           components: rustfmt
           target: wasm32-unknown-unknown
-          default: true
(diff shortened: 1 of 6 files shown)


.github/workflows/coverage.yml
```diff
@@ -38,15 +38,13 @@ jobs:
         with:
           submodules: recursive
       - name: Install toolchain
-        uses: actions-rs/toolchain@v1
+        uses: dtolnay/rust-toolchain@nightly
```
ermalkaleci, line 41: can't we just add those using bash?

ermalkaleci: #2424 ^

xlc (approved):

PR #2425, opened 2022-11-29
setup toolchain script

ermalkaleci:


xlc (changes requested):
I don't think we want to reuse the same init script for development & CI server.

ermalkaleci:
> I don't think we want to reuse the same init script for development & CI server.
> 
> 

This is just installing toolchain

xlc:
for example, we want to use minimal profile for CI, but the default profile for dev

PR #2433, opened 2022-12-05
fix docker build

zjb0807:
Test action: https://github.com/AcalaNetwork/Acala/actions/runs/3616352492


zjb0807:
@ermalkaleci Why does the CustomRunner still need the `protoc`?
https://github.com/AcalaNetwork/Acala/actions/runs/3616187051

xlc (approved):

PR #2434, opened 2022-12-06
Fix mandala benchmarks

zjb0807:


runtime/mandala/src/benchmarking/nutsfinance_stable_asset.rs
```diff
@@ -166,7 +166,7 @@ runtime_benchmarks! {
 		let pool_id = StableAsset::pool_count() - 1;
 		StableAsset::mint(RawOrigin::Signed(tester.clone()).into(), pool_id, mint_args, 0u128)?;
 	}: {
-		StableAsset::redeem_single(RawOrigin::Signed(tester).into(), pool_id, 10_000u128, 0u32, 0u128, u);
+		let _ = StableAsset::redeem_single(RawOrigin::Signed(tester).into(), pool_id, 10_000u128, 0u32, 0u128, u);
```
xlc, line 169: why ignore the result?

zjb0807: https://github.com/AcalaNetwork/Acala/pull/2411
This PR should be to test the worst-case iteration weight. 
I just fixed the warnings.

PR #2454, opened 2023-01-13
Set evm origin for xcm

xlc:


modules/evm/src/lib.rs
```diff
@@ -345,7 +345,7 @@ pub mod module {
 	/// ExtrinsicOrigin: Option<AccountId>
 	#[pallet::storage]
 	#[pallet::getter(fn extrinsic_origin)]
-	pub type ExtrinsicOrigin<T: Config> = StorageValue<_, T::AccountId, OptionQuery>;
+	pub type ExtrinsicOrigin<T: Config> = StorageValue<_, Vec<T::AccountId>, OptionQuery>;
```
xlc, line 348: this is safe without migration as this value should never be written to state

runtime/common/src/xcm_impl.rs
```diff
@@ -235,6 +236,34 @@ impl<FixedRate: Get<u128>, R: TakeRevenue, M: BuyWeightRate> Drop for FixedRateO
...
+		origin: impl Into<MultiLocation>,
+		message: Xcm<Config::RuntimeCall>,
+		weight_limit: XcmWeight,
+		weight_credit: XcmWeight,
+	) -> Outcome {
+		let origin = origin.into();
+		let account = AccountIdConvert::convert(origin.clone());
+		if let Ok(account) = account {
+			EVMBridge::push_origin(account);
+			defer!(EVMBridge::pop_origin());
```
zjb0807, line 261: I don't feel like `ExtrinsicOrigin ` needs to be an array

xlc: yeah it doesn't have to. we can just add a clear_origin method

however I am slightly bit worried about reentrancy issue. e.g. if someone send a tx somehow trigger XcmExecution, the evm origin will be cleared and follow execution will not have the correct origin.

xlc: I think this defer may execute when the `if` out of scope so what wouldn't work...

shaunxw: i might miss something but here it seems nothing happened between pushing and popping the origin

xlc: pop is defered using `defer`macro

shaunxw: I think it's only defered in the `if` scope and will be executed before xcm executor

zjb0807: https://github.com/paritytech/substrate/blob/master/primitives/core/src/defer.rs#L65-L73
Yes. `defer` executes in the `if` scope

zjb0807:
LGTM

When `sibling parachain sovereign account`'s balance < storage fee, the XCM message will fail. We need to monitor the balance of the `sibling parachain sovereign account`.(`xcm::execute_xcm_in_credit: [Parachain] Execution errored at 0: FailedToTransactAsset("ReserveStorageFailed") (original_origin: MultiLocation { parents: 1, interior: X1(Parachain(2001)) }
)`)

image

PR #2459, opened 2023-01-16
Update srtool-build.sh

StrawberryFlavor:


xlc (dismissed):

scripts/srtool-build.sh
```diff
@@ -2,7 +2,7 @@
 
 set -xe
 
-RUSTC_VERSION=1.64.0;
+RUSTC_VERSION=1.66.1;
```
zjb0807, line 5:
```suggestion
RUSTC_VERSION=`curl -s https://raw.githubusercontent.com/paritytech/srtool/master/RUSTC_VERSION`
```

zjb0807: https://github.com/chevdor/srtool-actions/blob/7091873ccd3367aa53f3b1e6c910606ede95ece5/action.yml#L82-L84

fetch the latest version, this will be the same as the workflow
https://github.com/AcalaNetwork/Acala/actions/runs/3921663720/jobs/6704115590#step:5:42

Issue #2475, opened 2023-02-08
EVM: precompile to perform xtokens transfer


xlc:
Expose xtokens.transfer to EVM to allow smart contracts to XCM tokens out

PR #2489, opened 2023-03-14
Add Xtokens precompile

zjb0807:
Closes: #2475 

TODO:
- [x] merge v0.9.38 and upgrade the xcm v3


runtime/common/src/precompile/mod.rs
```diff
@@ -166,7 +170,8 @@ where
 				// HOMA,
 				EVM_ACCOUNTS,
 				/* HONZON
-				 * INCENTIVES */
+				 * INCENTIVES
+				 * XTOKENS */
```
xlc, line 174: this will make it hard to test this on karura / acala testnet
one solution is enable it here, but have a migration to disable it with transaction-puase pallet
then we can unpause it on testnet
and for mainnet, we don't need a runtime upgrade to enable it

runtime/common/src/precompile/xtokens.rs
```diff
@@ -0,0 +1,963 @@
...
+					from, currency_id, amount, dest, weight
+				);
+
+				let transferred =
+					<orml_xtokens::Pallet<Runtime> as XcmTransfer<Runtime::AccountId, Balance, CurrencyId>>::transfer(
+						from,
+						currency_id,
+						amount,
+						dest,
+						WeightLimit::Limited(Weight::from_parts(weight, DEFAULT_PROOF_SIZE)),
```
xlc, line 134: I think it is most safe if we just let invoker specify proof size instead of hardcode it here

runtime/common/src/precompile/xtokens.rs
```diff
@@ -0,0 +1,963 @@
...
+					Decode::decode(&mut dest_bytes).map_err(|_| PrecompileFailure::Revert {
+						exit_status: ExitRevert::Reverted,
+						output: "invalid dest".into(),
+						cost: target_gas_limit(target_gas).unwrap_or_default(),
+					})?;
+				let dest: MultiLocation = versioned_dest.try_into().map_err(|()| PrecompileFailure::Revert {
+					exit_status: ExitRevert::Reverted,
+					output: "dest bad version".into(),
+					cost: target_gas_limit(target_gas).unwrap_or_default(),
+				})?;
```
xlc, line 118: this could be abstracted into a helper method?

runtime/common/src/precompile/xtokens.rs
```diff
@@ -0,0 +1,963 @@
...
+	fn cost(
+		input: &Input<Action, Runtime::AccountId, Runtime::AddressMapping, Runtime::Erc20InfoMapping>,
+	) -> Result<u64, PrecompileFailure> {
+		let action = input.action()?;
+
+		let cost: u64 = match action {
+			Action::Transfer => {
+				let currency_id = input.currency_id_at(2)?;
+				let read_currency = InputPricer::<Runtime>::read_currency(currency_id);
+
```
xlc, line 573: we could expose the weight functions from xtokens and use them here

runtime/common/src/precompile/xtokens.rs
zjb0807: https://github.com/AcalaNetwork/predeploy-contracts/pull/141

xlc (approved):

xlc (approved):

PR #2515, opened 2023-04-13
Improve selfdestruct

zjb0807:
If the contract data <= 100 items, delete it directly, otherwise schedule deletion.


modules/evm/src/lib.rs
```diff
@@ -1363,16 +1367,51 @@ impl<T: Config> Pallet<T> {
...
-			T::IdleScheduler::schedule(
-				EvmTask::Remove {
-					caller: *caller,
-					contract: *contract,
-					maintainer: contract_info.maintainer,
+					// Remove account after all of the storages are cleared.
+					*maybe_account_info = None;
+				} else {
+					// shouldn't happen.
+					debug_assert!(false);
```
xlc, line 1401: use `defensive`

modules/evm/src/lib.rs
```diff
@@ -1363,16 +1367,51 @@ impl<T: Config> Pallet<T> {
...
+					target: "evm",
+					"Remove: [from: {:?}, contract: {:?}, maintainer: {:?}, backend: {:?}, unique: {:?}]",
+					caller, contract, contract_info.maintainer, r.backend, r.unique
+				);
+
+				if r.maybe_cursor.is_none() {
+					// AllRemoved
+					let result = Pallet::<T>::refund_storage(caller, contract, &contract_info.maintainer);
+					// We also remove the contract if refund storage failed.
+					debug_assert!(result.is_ok());
```
xlc, line 1390: use `defensive`

xlc (changes requested):
Another way to do this without having the remove contract logic implemented twice is that still schedule the remove task to idle scheduler, but modify the `schedule` to return task id and a new method to idle scheduler to dispatch the task with a weight limit.

If the task can be executed under given weight limit, then the contract and task will be removed. Otherwise the scheduler will keep the task and run it later.

This way the additional cost is constant and we only have one place of the contract remove code, that will be updated later once we fully integrated with 2d weight

modules/evm/src/lib.rs
```diff
@@ -1363,18 +1368,42 @@ impl<T: Config> Pallet<T> {
 				}
 			});
 
-			ContractStorageSizes::<T>::take(contract);
+			let total_size = ContractStorageSizes::<T>::take(contract);
+			// If there are less than 50 items, we will remove it directly
+			maybe_weight_limit = if total_size
```
xlc, line 1373: I think we can always dispatch the task with a hardocded weight limit regardless if it can purge the whole contract or not

zjb0807: In this way, if it cannot be completely removed, part of it will be removed, and the rest will be scheduled to remove.

modules/evm/src/lib.rs
```diff
@@ -1343,15 +1345,17 @@ impl<T: Config> Pallet<T> {
 	#[transactional]
 	pub fn remove_contract(caller: &EvmAddress, contract: &EvmAddress) -> DispatchResult {
 		let contract_account = T::AddressMapping::get_account_id(contract);
-
-		Accounts::<T>::try_mutate_exists(contract, |account_info| -> DispatchResult {
+		let mut task_id = Default::default();
+		Accounts::<T>::try_mutate_exists(contract, |maybe_account_info| -> DispatchResult {
```
xlc, line 1349: `try_mutate_exists` can return `task_id`, just change `DispatchResult` to `Result<Nonce, DispatchError>`

modules/evm/src/lib.rs
```diff
@@ -1363,18 +1367,29 @@ impl<T: Config> Pallet<T> {
...
+			)?;
+
+			Ok(())
 		})?;
 
+		// try to dispatch the task, limit 50 DB writes
+		let weight_limit = Weight::from_ref_time(
+			<T as frame_system::Config>::DbWeight::get()
+				.write
+				.saturating_mul(REMOVE_LIMIT.saturating_div(2).into()),
```
xlc, line 1389: I think it is better to just have two constant. This could be `IMMEDIATE_REMOVE_LIMIT`

xlc (approved):

PR #2519, opened 2023-04-24
Add AccountKey20Aliases to support AccountKey20

zjb0807:


runtime/mandala/src/xcm_config.rs
```diff
@@ -66,8 +67,41 @@ pub type LocationToAccountId = (
 	SiblingParachainConvertsVia<Sibling, AccountId>,
 	// Straight up local `AccountId32` origins just alias directly to `AccountId`.
 	AccountId32Aliases<RelayNetwork, AccountId>,
+	// Convert `AccountKey20` to `AccountId`
+	AccountKey20Aliases<RelayNetwork, AccountId, EvmAddressMapping<Runtime>>,
 );
 
+/// Convert `AccountKey20` to `AccountId`
+pub struct AccountKey20Aliases<Network, AccountId, AddressMapping>(PhantomData<(Network, AccountId, AddressMapping)>);
```
xlc, line 75: it should be possible to put this to common runtime and avoid duplicated implementation?

runtime/acala/src/xcm_config.rs
```diff
@@ -70,8 +71,41 @@ pub type LocationToAccountId = (
 	SiblingParachainConvertsVia<Sibling, AccountId>,
 	// Straight up local `AccountId32` origins just alias directly to `AccountId`.
 	AccountId32Aliases<RelayNetwork, AccountId>,
+	// Convert `AccountKey20` to `AccountId`
+	AccountKey20Aliases<RelayNetwork, AccountId, EvmAddressMapping<Runtime>>,
```
xlc, line 75: but relay shouldn't have account 20?

zjb0807: I'm not sure what form it is. It can be added when needed

xlc (approved):

PR #2520, opened 2023-04-25
Fix bugs in stable asset balance conversion logic

ukby1234:
Fix bugs in stable asset balance conversion logic

### DIFF
--- a/runtime/acala/src/lib.rs
+++ b/runtime/acala/src/lib.rs
@@ -1602,9 +1602,9 @@ impl orml_tokens::ConvertBalance<Balance, Balance> for ConvertBalanceHoma {
 
 	fn convert_balance(balance: Balance, asset_id: CurrencyId) -> Balance {
 		match asset_id {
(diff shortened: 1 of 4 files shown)


runtime/acala/src/lib.rs
```diff
@@ -1602,9 +1602,9 @@ impl orml_tokens::ConvertBalance<Balance, Balance> for ConvertBalanceHoma {
 
 	fn convert_balance(balance: Balance, asset_id: CurrencyId) -> Balance {
 		match asset_id {
-			CurrencyId::Token(TokenSymbol::LDOT) => {
-				Homa::get_exchange_rate().checked_mul_int(balance).unwrap_or_default()
-			}
+			CurrencyId::Token(TokenSymbol::LDOT) => Homa::get_exchange_rate()
+				.checked_mul_int(balance)
```
xlc, line 1606: use `saturating_mul_int`

ukby1234: done.

xlc (approved):

PR #2526, opened 2023-05-06
Fix some integration tests

wangjj9219:
It's weird that staking_unbond failed because execute the xcm message occurs `XcmError::MaxWeightInvalid`,  
though the weight limit  should be sane. I try to set huge weight_limit and fee,  but it is still the same mistake.


xlc:
it is possible there is some proof size weight bug in xcm

xlc (approved):

PR #2530, opened 2023-05-12
Fix block limit test

shunjizhan:
somehow `createContractLoop` now only works with 50 items, otherwise it will exceed storage limit

not sure if this is bug caused by unexpectedly high storage usage (previously it works with 350, now 50, so 7X more storage used)?


ermalkaleci:
The storage limit hasn't changed. Test should not be changed. Must be something wrong with bodhi

PR #2532, opened 2023-05-15
add proof size to weight output

wangjj9219:


zjb0807:
https://github.com/AcalaNetwork/Acala/blob/3bd537c16358fc1ce4b6e17151d2bb9b2b041429/templates/precompile-weight-template.hbs#L33

should update and rerun `bench-evm`

wangjj9219:
> https://github.com/AcalaNetwork/Acala/blob/3bd537c16358fc1ce4b6e17151d2bb9b2b041429/templates/precompile-weight-template.hbs#L33
> 
> should update and rerun `bench-evm`

It seems that orml-bencher currently does not support proof size.

ermalkaleci:
> > https://github.com/AcalaNetwork/Acala/blob/3bd537c16358fc1ce4b6e17151d2bb9b2b041429/templates/precompile-weight-template.hbs#L33
> > 
> > should update and rerun `bench-evm`
> 
> It seems that orml-bencher currently does not support proof size.

yes it doesn't but it should be fine for now because those weights are used to calculate evm gas usage

runtime/acala/src/weights/module_cdp_engine.rs
```diff
@@ -49,125 +49,220 @@ use sp_std::marker::PhantomData;
...
+	// Proof: Auction Auctions (max_values: None, max_size: Some(70), added: 2545, mode: MaxEncodedLen)
 	/// The range of component `b` is `[1, 50]`.
 	fn liquidate_by_auction(b: u32, ) -> Weight {
-		// Minimum execution time: 182_413 nanoseconds.
-		Weight::from_parts(187_291_949, 0)
-			// Standard Error: 22_731
-			.saturating_add(Weight::from_parts(11_624_158, 0).saturating_mul(b.into()))
+		// Proof Size summary in bytes:
+		//  Measured:  `3484`
+		//  Estimated: `103387`
```
xlc, line 152: why estimated is so much higher than measured?

PR #2543, opened 2023-05-22
use evm-tests without lock file

ermalkaleci:


Makefile
```diff
@@ -215,15 +215,15 @@ cargo-update:
 
 .PHONY: build-wasm-mandala
 build-wasm-mandala:
-	./scripts/build-only-wasm.sh --profile production -p mandala-runtime --features=on-chain-release-build --workspace --exclude runtime-integration-tests --exclude e2e-tests
+	./scripts/build-only-wasm.sh --profile production -p mandala-runtime --features=on-chain-release-build --package mandala-runtime
```
xlc, line 218: why this change? there is already `-p mandala-runtime`

xlc (changes requested):

Makefile
ermalkaleci: Because all we want is the mandala-runtime, so it has to compile one package and not entire workspace

ermalkaleci: Old command will build relaychain runtime as well

ermalkaleci: Oh I think I missed it

ermalkaleci: Will fix tomorrow

xlc (approved):

PR #2547, opened 2023-05-26
inc nonce if evm-tx fails

ermalkaleci:


zjb0807:
I think we should do this in `post_dispatch` here
 https://github.com/AcalaNetwork/Acala/blob/c5a2d54e57911f28c4048cf8857871accf8166a0/runtime/common/src/check_nonce.rs#L107-L141

ermalkaleci:
> I think we should do this in `post_dispatch` here
> 
> https://github.com/AcalaNetwork/Acala/blob/c5a2d54e57911f28c4048cf8857871accf8166a0/runtime/common/src/check_nonce.rs#L107-L141

ues but we need this on fail and dispatch result is always Ok

xlc (changes requested):
inc_nonce is added to many places but only 1 test change means most of the changes are not unit tested.

I don't think we care about failing operational tx such as create_predeploy_contract

ermalkaleci:
> inc_nonce is added to many places but only 1 test change means most of the changes are not unit tested.
> 
> 
> 
> I don't think we care about failing operational tx such as create_predeploy_contract

I'm not done yet. Pushed this for @shunjizhan can test

xlc (approved):

zjb0807 (approved):

PR #2549, opened 2023-05-29
fix decode_gas_limit

zjb0807:
https://github.com/AcalaNetwork/bodhi.js/pull/740


shunjizhan:
LGTM, when using 21 instead of 22, one corner case is 2^21 < required storage < blocklimit, tx will fail. But I think no tx will actually use that much

ermalkaleci (approved):

Issue #2495, opened 2023-03-20
RefProvider


xlc:
With https://github.com/paritytech/substrate/pull/12951, reserved balance no longer add provider ref.

Update the module-support to add features to reserve balance and manage provider ref.

- fn reserve(who: Account, id: ReserveId, amount: Balance)
  - reserve amount.max(ED)
  - inc provider for who
- fn unreserve(who: Account, id: ReserveId, amount: Option<Balance>)
  - amount = amount.unwrap_or(current_reserve_amount)
  - ensure remaining is zero or >= ED
  - if remaning == 0
    - dec provider for who

Review all pallets that uses reserve and replace the usage with this

This should be fully backward compatible

In future, we should be able to modify those two helpers to migrate to the `fungible` trait

wangjj9219:
Now, as long as the `reserve_amount` is not zero, even if `free_balance` is zero, the account will not be dusted. 

But reserve operation may fail for provider/consumer:
if an account only free: 2*ED, reserved: 0,  its provider,consumer: 1,0

reserve 2ED, the expected result is free: 0, reserved: 2*ED,  will inc consumer first, then dec provider.  dec provider will fail for ConsumerRemaining. 

we can:
use wrapper reserve function, which do inc_provider before call reserve 
use wrapper unreserve function , which do dec_provider after call unreserve and check

In this way, regardless of whether free_balance has ED or not, 
before reserve step, additional provider additional was provided for dec. It seems a feasible solution, but for these account has reserved already, it should require more judgment to decide whether to  inc provider or  dec provider.

PR #2560, opened 2023-06-15
Polkadot v0.9.42

wangjj9219:
# pallet-balances  breaking change
substrate@0.9.42 uses new provider/consumers mechanism, there's breaking change: now pallet-balances Account supply provider based on its `free_balance` instead of `total_balance`, and no-zero `reserved_balance` will as consumer.  

It will affect some case, if the ED of native token is 100: 

if a account endowed by `200`(>ed) free_balance, its provider inc 1, the `(provider, consumer)` is `(1, 0)` . Then reserve its `200` free_balance to reserved balance, according to new rules, consumer will inc 1, and free_balance changes to `0`, it's below ED, provider will dec 1, However, due to  consumer is 1, the provider cannot become 0. So the reserve operation will fail.

For users, when they do some operations,  their account generally has sufficient Native token, or has other tokens assets as providers.  BUT some functions of Acala will use some escrow accounts, which are only used to store reserved ACA/KAR:

## 1. create NFT class
It will do:
1. transfer xxx ACA as deposit to the class account
2. then class account reserve  xxx ACA

now it will fail when reserve. 

### Solution
For create NFT class,  transfer total deposit + ACA ED to class account. after reserve total deposit, free_balance is still >= ED (`provider, consumer` is `1, 0` ) . 
For mint nft and transfer nft,  if the receiver's free_balance is <= ed,  transfer ED to receiver.

## 2. deploy contract or call contract in EVM+
the situation is much more complicated：

+ deploy a contract or call contract will charge(reserve) storage deposit fee
+ when deploy a contract on EVM+,  will inc_provider directly for contract account (`provider, consumer` is `1, 0` )  before charge storage. 
+ charge storage will do:
1. if the storage deposit `delta` is positive, 
   1.1 `transfer` `delta` from origin caller to contract account
   1.2  `reserve` `delta` amount  free_balance to reserved_balance
2. if the storage deposit `delta` is negative, 
   2.1. `repatriate_reserved` the `delta` reserved balance from contract account to origin caller

Here are the results for each case：
a. deploy a new contract, the `delta` is positive and >= ED, it will succeed. before the `reserve` operation , `provider, consumer` is `2, 0`, `1` is added by `inc_provider` directly,   `1` is added by transfer operation.  After `reserve` operation, `provider, consumer` is `1, 1`

b. call contract, the `delta` is positive and >= ED, it will succeed. `provider, consumer`  will change from `1, 1` to `2, 1` when `transfer ` operation,  after `reserve` operation, it becomes to `1, 1` 

c. deploy a new contract or call contract ,  `delta` is positive , but below ED. !!! It will fail !!! , due to the `TokenError::BelowMinimum` error when  transfer `delta` to  contract account.  the latest impl for `transfer` not allowed receiver's free_balance is below ED after transfer,  even if it's total balance is >= ED.

d. call contract, `delta` is negative. Usually, the free balance of the origin caller is greater than ED,  so here it will be successful.

### solution
We add `SystemAccountStore` to instead of the impl for `StoredMap` of `System`,  it regards existed `frame_system::Account` also exists `pallet_balances::Account`, even if it's AccountData is default.
so that `repatriate_reserved` can transfer reserved balance to initialized contract account, which is created
by inc_provider has default `AccountData`. 

Then use `repatriate_reserved` to instead of following operation:
1. unreserve storage deposit from caller
2. caller transfer storage deposit to contract account
3. contract account reserve storage deposit

# module-currencies refactor fungibles traits impl
In this pr module-currencies refactor the impl for fungibles traits：
1. In impl traits, each function should judge the asset_id, if it is erc20, it will be processed separately, if it is orml-tokens asset or native token, it will be passed to the corresponding function of Tokens and Balances respectively.

2. the fungibles in the substrate change the functions at the fund operation level to the default implementation(such as transfer, burn_from, mint_into...). module-currencies still overwrite these functions, it's necessary to compatible fungibles traits with ERC20. So, if  use fungibles traits in code, should only use these overwrited impl functions.


xlc:
we should do this to handle ref count https://github.com/AcalaNetwork/Acala/issues/2495

wangjj9219:
> we should do this to handle ref count #2495 and replace most of our reserves to RefProvider

How to prevent `reserve` operation fail for free_balance < ED after reserve, then dec_provider fail? 

Now we have come to a crossroads, should we use the current PR solution or do this:
https://github.com/AcalaNetwork/Acala/issues/2495#issuecomment-1598006142

xlc:
we also need to kill dust on account killed

modules/currencies/src/tests.rs
```diff
@@ -2479,73 +2913,82 @@ fn sweep_dust_native_currency_works() {
...
-
-		assert_noop!(
-			Currencies::sweep_dust(RuntimeOrigin::signed(bob()), NATIVE_CURRENCY_ID, accounts.clone()),
-			DispatchError::BadOrigin
+		// TODO: seems the insert directly does not work now, it's probably because of the new machanism of
+		// provider and consumer: https://github.com/paritytech/substrate/blob/569aae5341ea0c1d10426fa1ec13a36c0b64393b/frame/system/src/lib.rs#L1692
+		// consider deposit_creating alive account, then decrease the ED to fix this test!
+		assert_eq!(
+			<Runtime as pallet_balances::Config>::AccountStore::get(&bob()),
+			Default::default()
```
zjb0807, line 2926: TODO: fix tests

runtime/karura/src/lib.rs
```diff
@@ -1826,10 +1848,6 @@ pub type Executive = frame_executive::Executive<
 		pallet_xcm::migration::v1::MigrateToV1<Runtime>,
 		orml_unknown_tokens::Migration<Runtime>,
 		// Note: The following Migrations do not use the StorageVersion feature, must to be removed after the upgrade
-		module_asset_registry::migrations::MigrateV1MultiLocationToV3<Runtime>,
-		module_xcm_interface::migrations::MigrateXcmDestWeightAndFee<Runtime>,
-		module_transaction_pause::migrations::MigrateEvmPrecompile<Runtime>,
-		MigrateSetXcmVersionForKusama,
```
zjb0807, line 1832: clean all migrations

modules/currencies/src/lib.rs
```diff
@@ -740,36 +773,59 @@ impl<T: Config> fungibles::Inspect<T::AccountId> for Pallet<T> {
 	type Balance = BalanceOf<T>;
 
 	fn total_issuance(asset_id: Self::AssetId) -> Self::Balance {
-		<Self as MultiCurrency<_>>::total_issuance(asset_id)
+		match asset_id {
+			CurrencyId::Erc20(_) => <Self as MultiCurrency<_>>::total_issuance(asset_id),
+			id if id == T::GetNativeCurrencyId::get() => <T::NativeCurrency as fungible::Inspect<_>>::total_issuance(),
+			_ => <T::MultiCurrency as fungibles::Inspect<_>>::total_issuance(asset_id),
+		}
```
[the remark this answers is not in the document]
wangjj9219: Although `MultiCurrency` and `fungibles::Inspect` are same in the implementation for these functions, we hope to separate the two sets of interfaces, try not to mix them with each other

zjb0807 (approved):

modules/evm/src/tests.rs
```diff
@@ -921,18 +921,29 @@ fn create_predeploy_contract_works() {
...
-				1000000,
-				vec![],
-			),
-			Error::<Runtime>::ContractNotFound
-		);
+
+		// the call is ok, but actually deploy failed, will trige CreatedFailed event
+		// if contract is empty, will skip inc_provider for contract account, so it
+		// fail at charge storage.
+		assert_ok!(EVM::create_predeploy_contract(
```
ermalkaleci, line 928: this is supposed to cover `ContractNotFound` error

PR #2575, opened 2023-07-12
precompile for liquid-crowdloan

xlc:


runtime/common/src/precompile/liquid_crowdloan.rs
```diff
@@ -0,0 +1,286 @@
...
+					exit_status: ExitSucceed::Returned,
+					cost: gas_cost,
+					output: Output::encode_uint(redeem_amount),
+					logs: Default::default(),
+				})
+			}
+			Action::GetRedeemCurrency => {
+				let currency_id = <module_liquid_crowdloan::Pallet<Runtime>>::redeem_currency();
+				let address = <Runtime as module_prices::Config>::Erc20InfoMapping::encode_evm_address(currency_id)
+					.unwrap_or_default();
```
zjb0807, line 95: It's better to return `Revert` here

runtime/common/src/precompile/mod.rs
```diff
@@ -172,6 +176,7 @@ where
 				HONZON,
 				INCENTIVES,
 				XTOKENS,
+				LIQUID_CROWDLOAN,
```
zjb0807, line 179: `LIQUID_CROWDLOAN ` only for acala runtime

runtime/common/src/precompile/liquid_crowdloan.rs
ermalkaleci: This is a read operation. It better to return empty address and let contract handle it

xlc: in this particular case, it is something should never happen and I think revert is fine

ermalkaleci: just for consistency. every read operation shouldn't revert

runtime/common/src/precompile/mod.rs
```diff
@@ -347,7 +355,7 @@ where
 			} else if address == XTOKENS {
 				Some(XtokensPrecompile::<R>::execute(input, target_gas, context, is_static))
 			} else {
-				None
+				E::execute(&Default::default(), address, input, target_gas, context, is_static)
```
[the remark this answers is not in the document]
xlc: I guess we can but I really want to avoid feature flags if possible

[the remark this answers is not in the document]
ermalkaleci: yes but this is adding more complexity, more code to manage

runtime/common/src/precompile/liquid_crowdloan.rs
```diff
@@ -0,0 +1,290 @@
...
+				RuntimeOrigin::root(),
+				LiquidCrowdloanPalletId::get().into_account_truncating(),
+				DOT,
+				1_000_000_000
+			));
+
+			// redeem(address,uint256) -> 1e9a6950
+			// who
+			// amount 1e9
+			let input = hex!(
```
ermalkaleci, line 185:
```suggestion
			let input = hex! { "
```
to make fmt ignore this

runtime/common/src/precompile/mod.rs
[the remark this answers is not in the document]
xlc: not really compare to feature flags. there is no easy way to modify where clause with feature flag

zjb0807 (approved):

ermalkaleci (approved):

PR #2592, opened 2023-08-21
update evm-tests

ermalkaleci:


xlc:
Does this catch the nonce issue we had?

ermalkaleci:
> Does this catch the nonce issue we had?

no, this asserts invalid tx and enables a performance test

xlc (approved):

PR #2601, opened 2023-09-01
remove with_fee_paid_by

xlc:
Current implementation is unsafe and too complicated to do it safely


modules/transaction-payment/src/lib.rs
```diff
@@ -639,22 +639,6 @@ pub mod module {
 			call.dispatch(origin)
 		}
 
-		/// Wrap call with fee paid by other account
-		#[pallet::call_index(5)]
```
ermalkaleci, line 643: @xlc should we leave func name and call_index as comment here?

xlc: good point. that will ensure we never reuse the index

PR #2602, opened 2023-09-01
new wasm-bencher

ermalkaleci:


modules/evm/src/runner/stack.rs
```diff
@@ -772,7 +772,7 @@ impl<'vicinity, 'config, T: Config> StackStateT<'config> for SubstrateStackState
 	}
 
 	fn reset_storage(&mut self, address: H160) {
-		let _ = <AccountStorages<T>>::clear_prefix(address, u32::MAX, None);
```
ermalkaleci, line 775: new bencher will count limit as written operation so I'm changing this to drain_prefix to avoid counting any write operation

zjb0807:
@ermalkaleci The bot doesn‘t push the result, is it because the file has not been changed?

ermalkaleci:
> @ermalkaleci The bot doesn‘t push the result, is it because the file has not been changed?

Bot uses master version of benchmark.yml which is missing weight-gen. That's why it's failing

modules/evm/src/runner/stack.rs
xlc: best to leave a comment explain this for future readers

xlc (approved):

Issue #2634, opened 2023-10-20
EVM precompile to transfer token to Substrate address


xlc:
Update currency precompile to add function to transfer to Substrate address (bytes32).
Update tokens predeploy contract to add new methods:
- `transferToAccountId32(bytes32 dest, uint256 amount)`

PR #2636, opened 2023-10-24
transfer2 precompile

xlc:
Part of #2634

Transfer to Substrate account directly so that we don't need to require user to have evm account

### DIFF
--- a/runtime/common/src/precompile/multicurrency.rs
+++ b/runtime/common/src/precompile/multicurrency.rs
@@ -59,7 +59,7 @@ pub enum Action {
 	QueryTotalIssuance = "totalSupply()",
 	QueryBalance = "balanceOf(address)",
 	Transfer = "transfer(address,address,uint256)",
-	Transfer2 = "transfer2(address,bytes32,uint256)",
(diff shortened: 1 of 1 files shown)


zjb0807 (approved):

runtime/common/src/precompile/multicurrency.rs
```diff
@@ -56,12 +59,14 @@ pub enum Action {
 	QueryTotalIssuance = "totalSupply()",
 	QueryBalance = "balanceOf(address)",
 	Transfer = "transfer(address,address,uint256)",
+	Transfer2 = "transfer2(address,bytes32,uint256)",
```
ermalkaleci, line 62: can we have a better name?

xlc: do you have a suggestion? also this is just kinda following the naming convention 💁‍♂️ like create & create2

ermalkaleci: TransferToAccountId, TransferNativeDestination

PR #2640, opened 2023-11-06
validate evm account id

ermalkaleci:


xlc (approved):

modules/evm/src/tests.rs
```diff
@@ -37,9 +37,7 @@ use std::str::FromStr;
 #[test]
 fn fail_call_return_ok_and_inc_nonce() {
 	new_test_ext().execute_with(|| {
-		let mut data = [5u8; 32];
-		data[0..4].copy_from_slice(b"evm:");
-		let signer: AccountId32 = AccountId32::from(data);
+		let signer: AccountId32 = AccountId32::from([5u8; 32]);
```
xlc, line 40: this changes the test. we want this to be a valid evm account

xlc (changes requested):
need to fix test

modules/evm/src/tests.rs
ermalkaleci: This doesn't change anything. `data[0..4].copy_from_slice(b"evm:");` was probably copy paste. Had nothing to do with this fix. I just clean it up

xlc: but `evm:` prefix means the account is EVM address. the meaning of the test changed

ermalkaleci: yes but it's calling `get_or_create_evm_address`. result is the same

ermalkaleci: before fix it will return addr from `get_evm_address` because it starts with `evm:`. after fix this will create the address. So the rest of the test will work the same

xlc: ok in that case why this is not the same as the setup in `inc_nonce_with_revert`?

```rs
let alice = alice();
let account = MockAddressMapping::get_account_id(&alice);
let origin = RuntimeOrigin::signed(account);
```

ermalkaleci: I can make it the same as `inc_nonce_with_revert` for consistency

xlc (approved):

PR #2644, opened 2023-11-15
update to 1.3.0

xlc:
- [ ] trigger lazy migration for preimage https://github.com/paritytech/polkadot-sdk/pull/1363
- [ ] ensure we don't have any existing holds
- [ ] reimplement removed e2e-tests using ts-tests


wangjj9219 (dismissed):

ermalkaleci (approved):

modules/xcm-interface/src/mocks/mod.rs
```diff
@@ -194,7 +194,8 @@ macro_rules! impl_mock {
 			type MaxReserves = ();
 			type ReserveIdentifier = [u8; 8];
 			type WeightInfo = ();
-			type RuntimeHoldReason = ();
+				type RuntimeHoldReason = RuntimeHoldReason;
+	type RuntimeFreezeReason = RuntimeFreezeReason;
```
zjb0807, line 198: format here

PR #2646, opened 2023-11-21
[XNFT PoC] xnft module - XCM asset transactor for module nft

mrshiposha:
This PR adds a new `xnft` module.
This module serves as a PoC of the XCM NFT Asset Transactor on Karura. This code will be used as a base to implement a more general Asset Transactor for Substrate chains.

### Foreign Collection Registration
* To keep things simple in the PoC, a foreign NFT collection can be registered manually by a technical committee member invoking the `xnft.registerAsset` extrinsic. In the future, we need to come up with a mechanism that will allow an owner of an original collection on a remote chain to register the corresponding foreign collection themselves without involving a governance body. 

### Notes:
* `xnft` holds the mapping between the XCM identification of an NFT asset and the local identification in the `nft` module.
The mapping is bidirectional, though only one side is used now. The backward mapping can be used in the future for extrinsics like `xtokens.transfer` where we identify assets using parachain-local IDs.
* `xnft` pallet account serves as the location for NFTs when they are located in the Holding Register to avoid burning an NFT and losing the associated data.
* Also, the `xnft` pallet account will hold a derivative of an NFT when the original NFT is transferred elsewhere instead of burning the derivative. This serves two goals:
    * If the original NFT returns, it will correspond to the same derivative NFT with the same Karura-local ID.
    * Suppose the reserve chain erroneously sends the `ReserveAssetDeposited` containing an NFT that was already present on Karura. In that case, it will fail due to the failed transfer from the xnft module account to the beneficiary because the derivative NFT will belong to someone else's account, avoiding creating a duplicate.

### Tests
The common tests are located in the [xnft-tests](https://github.com/uniqueNetwork/xnft-tests) repository.

### DIFF
--- a/modules/xnft/src/lib.rs
+++ b/modules/xnft/src/lib.rs
@@ -1,0 +1,154 @@
+// This file is part of Acala.
+
+// Copyright (C) 2023 Unique Network.
+// SPDX-License-Identifier: GPL-3.0-or-later WITH Classpath-exception-2.0
+
+// This program is free software: you can redistribute it and/or modify
+// it under the terms of the GNU General Public License as published by
+// the Free Software Foundation, either version 3 of the License, or
+// (at your option) any later version.
+
+// This program is distributed in the hope that it will be useful,
+// but WITHOUT ANY WARRANTY; without even the implied warranty of
+// MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE. See the
+// GNU General Public License for more details.
+
+// You should have received a copy of the GNU General Public License
+// along with this program. If not, see <https://www.gnu.org/licenses/>.
+
+#![cfg_attr(not(feature = "std"), no_std)]
+#![allow(clippy::unused_unit)]
+
+use cumulus_primitives_core::ParaId;
+use frame_support::{ensure, pallet_prelude::*};
+use frame_system::pallet_prelude::*;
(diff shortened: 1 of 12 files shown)


modules/xnft/src/lib.rs
```diff
@@ -0,0 +1,154 @@
...
+	#[pallet::getter(fn class_to_foreign_asset)]
+	pub type ClassToForeignAsset<T: Config> = StorageMap<_, Twox64Concat, ClassIdOf<T>, xcm::v3::AssetId, OptionQuery>;
+
+	#[pallet::storage]
+	#[pallet::getter(fn asset_instance_to_item)]
+	pub type AssetInstanceToItem<T: Config> = StorageDoubleMap<
+		_,
+		Twox64Concat,
+		ClassIdOf<T>,
+		Twox64Concat,
```
wangjj9219, line 92: I'm not sure that `xcm::v3::AssetInstance` type key should use `Blake2_128Concat` hasher  @xlc

xlc:
1.3.0 is merged. can you update this to point to master please

modules/xnft/src/lib.rs
xlc: yeah better to make it blake2 as user have control of the content of the asset

mrshiposha:
Here is the new PR pointing to the master branch: https://github.com/AcalaNetwork/Acala/pull/2649

PR #2649, opened 2023-11-23
[XNFT PoC] xnft module - XCM asset transactor for module nft

mrshiposha:
This PR adds a new `xnft` module.
This module serves as a PoC of the XCM NFT Asset Transactor on Karura. This code will be used as a base to implement a more general Asset Transactor for Substrate chains.

### Foreign Collection Registration
* To keep things simple in the PoC, a foreign NFT collection can be registered manually by a technical committee member invoking the `xnft.registerAsset` extrinsic. In the future, we need to come up with a mechanism that will allow an owner of an original collection on a remote chain to register the corresponding foreign collection themselves without involving a governance body. 

### Notes:
* `xnft` holds the mapping between the XCM identification of an NFT asset and the local identification in the `nft` module.
The mapping is bidirectional, though only one side is used now. The backward mapping can be used in the future for extrinsics like `xtokens.transfer` where we identify assets using parachain-local IDs.
* `xnft` pallet account serves as the location for NFTs when they are located in the Holding Register to avoid burning an NFT and losing the associated data.
* Also, the `xnft` pallet account will hold a derivative of an NFT when the original NFT is transferred elsewhere instead of burning the derivative. This serves two goals:
    * If the original NFT returns, it will correspond to the same derivative NFT with the same Karura-local ID.
    * Suppose the reserve chain erroneously sends the `ReserveAssetDeposited` containing an NFT that was already present on Karura. In that case, it will fail due to the failed transfer from the xnft module account to the beneficiary because the derivative NFT will belong to someone else's account, avoiding creating a duplicate.

### Tests
The common tests are located in the [xnft-tests](https://github.com/uniqueNetwork/xnft-tests) repository.


modules/nft/src/lib.rs
```diff
@@ -65,7 +65,7 @@ pub struct ClassData<Balance> {
 	pub attributes: Attributes,
 }
 
-#[derive(Encode, Decode, Clone, RuntimeDebug, PartialEq, Eq, TypeInfo, Serialize, Deserialize)]
+#[derive(Encode, Decode, Clone, RuntimeDebug, PartialEq, Eq, TypeInfo, Serialize, Deserialize, Default)]
```
xlc, line 68: is the Default necessary? I want it to be explicit

modules/xnft/src/lib.rs
```diff
@@ -0,0 +1,156 @@
...
+	pub struct Pallet<T>(_);
+
+	#[pallet::call]
+	impl<T: Config> Pallet<T> {
+		#[pallet::call_index(0)]
+		#[pallet::weight(Weight::from_parts(1_000_000, 0)
+			.saturating_add(<module_nft::weights::AcalaWeight<T>>::create_class())
+			.saturating_add(T::DbWeight::get().reads(1))
+			.saturating_add(T::DbWeight::get().writes(2)))]
+		pub fn register_asset(origin: OriginFor<T>, foreign_asset: Box<AssetId>) -> DispatchResult {
```
xlc, line 122:
```suggestion
		pub fn register_asset(origin: OriginFor<T>, foreign_asset: Box<VersionedAssetId>) -> DispatchResult {
```

all the extrinsic should take Versioned XCM struct to ensure backward and forward compatibility

the storages, however, should stick with a particular version (v3 here) and use migration to upgrade to new version if needed

modules/xnft/src/xcm_helpers.rs
```diff
@@ -0,0 +1,97 @@
...
+use crate::*;
+use xcm::v3::AssetId::Concrete;
+use xcm_executor::traits::Error as MatchError;
+
+impl<T: Config> Pallet<T>
+where
+	TokenIdOf<T>: TryFrom<u128>,
+	ClassIdOf<T>: TryFrom<u128>,
+{
+	pub fn asset_to_collection(asset: &AssetId) -> Result<(ClassIdOf<T>, bool), MatchError> {
```
xlc, line 28: what is the bool in result? it is not obvious and either need to introduce a struct or very minimal lots of comments to explain it

modules/nft/src/lib.rs
mrshiposha: `Default` removed

modules/xnft/src/lib.rs
mrshiposha: Done

modules/xnft/src/xcm_helpers.rs
mrshiposha: `ClassLocality` enum added. Refactored.

modules/xnft/src/xcm_helpers.rs
```diff
@@ -0,0 +1,117 @@
...
+		match Self::asset_instance_to_item(asset, asset_instance) {
+			Some(token_id) => <ModuleNftPallet<T>>::do_transfer(&Self::account_id(), to, (asset, token_id))
+				.map_err(|_| XcmError::FailedToTransactAsset("non-fungible foreign item deposit failed")),
+			None => {
+				let token_id = <OrmlNftPallet<T>>::mint(
+					to,
+					asset,
+					Vec::new(),
+					module_nft::TokenData::<BalanceOf<T>> {
+						deposit: 0u32.into(),
```
xlc, line 76: It will be great if there is a way to take a deposit here. There needs to be another assets to buy weights so in theory we can use it for deposit. But I am not sure if the current XCM API allows such action.

xlc (dismissed):
Mostly good except the token deposit. But should be fine for an MVP with very limited usage.

mrshiposha:
Strangely, it didn't build in the CI. I tested it before the last push. Will fix.

modules/xnft/src/xcm_helpers.rs
```diff
@@ -0,0 +1,118 @@
...
+			_ => None,
+		}
+	}
+
+	pub fn deposit_foreign_asset(to: &T::AccountId, asset: ClassIdOf<T>, asset_instance: &AssetInstance) -> XcmResult {
+		match Self::asset_instance_to_item(asset, asset_instance) {
+			Some(token_id) => <ModuleNftPallet<T>>::do_transfer(&Self::account_id(), to, (asset, token_id))
+				.map_err(|_| XcmError::FailedToTransactAsset("non-fungible foreign item deposit failed")),
+			None => {
+				let token_id = <OrmlNftPallet<T>>::mint(
```
zjb0807, line 72: I'm just curious why there is only `mint` but no `burn`.  Is there any reason to transfer to pallet instead of burning?
@mrshiposha

mrshiposha: I tried to describe the reasons in the `Notes` of the description of the PR. They might be debatable, yet currently, I can't see a better solution for the following issues.

Let's define:

* local NFTs = NFTs minted on Karura
* foreign NFTs = NFTs minted on a remote chain
* derivative NFT = a local NFT corresponding to a foreign NFT

So, the issues:

1. We don't want to lose the data associated with local NFTs. We will lose the metadata and attributes if we burn a local NFT.
2. We don't want a local NFT to obtain different IDs. If we burn a local NFT and mint it again, it will be a new NFT with a new ID.
3. We **might** not want the derivative NFTs to change their IDs over time. When the reserve location deposits a foreign NFT, we mint a derivative NFT. If we burn such a derivative NFT, and later, if the reserve deposits the foreign NFT again, we'll mint a new derivative NFT with a new ID. Yet, maybe dApps must track such correspondence via some API and not rely on the derivative NFT's ID. So, perhaps, we could burn derivative NFTs.

The issues with IDs could be solved by implementing the minting with a predefined ID. Yet, it requires a change to the local NFT pallet logic, which is against the goal of creating a base implementation for a more general solution in the future.
The issue of losing the associated data of local NFTs might be solved by using a relaxed variant of burning operation: not removing the NFT data when burning. But this solution also looks like a debatable one.

mrshiposha: Also, we are solving another issue for free by using a transfer operation when depositing an existing derivative NFT.
The reserve location could erroneously send the same foreign NFT again. We need to avoid minting a duplicate derivative NFT. The transfer operation will fail if the pallet is not the owner of the derivative thus we won't create a duplicate derivative corresponding to the same foreign NFT.

Of course, we could solve this issue without using transfer, even if we were using mint/burn only. However, we would have to add a separate check for this.

mrshiposha: We could burn only derivative NFTs. So, only local NFTs will be transferred to the pallet's account.

However, we still could waste the storage since the XCM `BurnAsset` [does nothing but take assets from the Holding Register](https://github.com/paritytech/polkadot-sdk/blob/dbd8d20b25b29abf138e3ad35c40df4c55bd7c5c/polkadot/xcm/xcm-executor/src/lib.rs#L754-L757). So, if we don't burn an NFT on withdrawal, it will stay somewhere. But we don't want to burn a local NFT to avoid losing its data. And if we burn a local NFT partially (i.e., without deleting its data), the `BurnAsset` will waste the data nonetheless.

We could introduce a change to the XCM executor API to add a "callback" for burning assets. However, I'm not sure if it is enough since the `BurnAsset` is infallible [by the spec](https://github.com/paritytech/xcm-format#burnasset). Maybe on some occasions, we can't burn an NFT infallibly (I mean on other chains, especially those that use smart contracts).

mrshiposha: Anyways, following the PoC, we will develop a generalized "xnft" pallet where such behavior will be configurable.

xlc:
not 100% sure why the e2e-tests are failing. I will take a look

modules/xnft/src/xcm_helpers.rs
xlc: Ideally, we have some XCM ID for NFT (basically the MultiAsset) and we only ever display that to user. This means the local NFT ID doesn't matter and it is ok for it to change if user transfer out and in the same NFT because the XCM ID remains the same.

Local NFT also shouldn't hold any extra data, at least those data should transfer back to origin chain so it should be safe to delete them then the NFT is transferred back.

But yeah we can think about those later. For Karura, we could have a migration to cleanup unused NFT data once we upgraded to the new xnft pallet. We also don't expect too much usage so we shouldn't have too much redundant data.

xlc:
can you do a merge master to fix the e2e-tests issue

mrshiposha:
@xlc 
>Local NFT also shouldn't hold any extra data, at least those data should transfer back to origin chain so it should be safe to delete them then the NFT is transferred back.

We can't drop local NFTs' data if we use reserve-based transfers. Because Karura is the reserve location for these NFTs, Karura is the only source of truth of any aspect of these assets. We had a long thread about this on the Polkadot forum. [Here is Keith's post](https://forum.polkadot.network/t/cross-chain-nft-transfer/2080/31?u=mrshiposha).

Karura, as the reserve location, can't trust another chain with Karura's own NFTs' data.

xlc (approved):

PR #2654, opened 2023-11-29
fix ts test for me

xlc:
not 100% sure why localhost doesn't work but 127.0.0.1 works fine on my machine


ermalkaleci:
Use always ip

Issue #2648, opened 2023-11-22
reimplement removed e2e-tests using ts-tests


xlc:
Followup of #2644 to reimplement removed tests

PR #2655, opened 2023-11-30
Add e2e tests

zjb0807:
Closes: #2648


ts-tests/tests/test-evm-fill-block.ts
```diff
@@ -0,0 +1,132 @@
...
+
+        const creates = Array(15).fill(context.provider.api.tx.evm.create(
+            contract,
+            0,
+            2_000_000,
+            100_000,
+            []
+        ));
+
+        const tx = context.provider.api.tx.utility.batchAll(creates);
```
ermalkaleci, line 34: this is not the same as it was

zjb0807: `test-service` can manually generate blocks, send all of the transactions to mempool, and then generate blocks, but mandala cannot, so I use batchAll to test

ermalkaleci: can't we start manual-sealing, feed with tx and then build blocks and see how many tx are included per block?

ts-tests/tests/test-evm-call-fill-block.ts
```diff
@@ -0,0 +1,53 @@
...
+        }
+
+        while (true) {
+            const currentHeight = await context.provider.api.query.system.number();
+
+            if (currentHeight.toNumber() > beforeHeight) {
+                break;
+            }
+
+            setTimeout(() => { }, 1000);
```
xlc, line 42: this doesn't wait, I think you want `await new Promise(resolve => setTimeout(resolve, 1000))`

ts-tests/tests/test-gas.ts
```diff
@@ -18,7 +18,7 @@ describeWithAcala("Acala RPC (Gas)", (context) => {
 			from: alice.getAddress(),
 			data: "0x" + Block.bytecode,
 		});
-		expect(gas.toNumber()).to.be.eq(11301014);
+		expect(gas.toNumber()).to.be.eq(11301114);
```
xlc, line 21: we should convert those into snapshot tests at some point

zjb0807: use `closeTo` instead

xlc (approved):

xlc (approved):

PR #2696, opened 2024-01-28
Update ecrecover precompile

zjb0807:
ref: https://github.com/polkadot-evm/frontier/pull/964

### DIFF
--- a/modules/evm/src/precompiles/blake2/mod.rs
+++ b/modules/evm/src/precompiles/blake2/mod.rs
@@ -36,6 +36,7 @@ impl Precompile for Blake2F {
 		const BLAKE2_F_ARG_LEN: usize = 213;
 
 		let input = handle.input();
+
 		if input.len() != BLAKE2_F_ARG_LEN {
 			return Err(PrecompileFailure::Error {
 				exit_status: ExitError::Other(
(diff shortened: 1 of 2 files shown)


ermalkaleci:
we already have the correct version, what does this change?

zjb0807:
@ermalkaleci 

https://github.com/AcalaNetwork/Acala/blob/0b8c9b4b174401214f8ab4465bffaaa497527db9/modules/evm/src/precompiles/ecrecover.rs#L42-L47

If v > 26, then sig[64] = v - 27, let it be 0 or 1.  
while geth version only accept 27 and 28.

ermalkaleci:
That's all we need

ermalkaleci:
if this is fixing an issue then there should be a test. where's `v - 27` https://github.com/ethereum/go-ethereum/blob/a907d7e81aaeea15d80b2d3209ad8e08e3bf49e0/core/vm/contracts.go#L174?

zjb0807:
It checked that `input[63] > 26`, so if it is `input[63] = 30`, it need to continue to verify the signature.
Now strictly check input[63] is 27/28.

ermalkaleci:
> It checked that `input[63] > 26`, so if it is `input[63] = 30`, it need to continue to verify the signature.
> 
> Now strictly check input[63] is 27/28.

If v > 26 then v - 27

zjb0807:
> If v > 26 then v - 27

So v can be `27~255` to call `secp256k1_ecdsa_recover`. Now it is only allowed on 27/28.

Issue #2650, opened 2023-11-23
Update EVM RPC runtime APi to allow estimate gas from Substrate account


xlc:
need new version of the `call` and `create` that takes AccountId as origin

this is needed to estimate gas usage for account that have not yet binded

PR #2698, opened 2024-01-29
Add account_call and account_create in EVM Runtime RPC

zjb0807:
Closes: #2650


xlc (approved):

xlc (approved):

shunjizhan:
currently when adding new runtime code, we need to copy and paste the same code for each of `{ mandala, karura, acala }` runtime definition, I am curious if there is any way to keep the code more DRY? Maybe like a shared runtime definition, and each of the network can extend it with overrides

xlc:
with this https://github.com/polkadot-js/api/issues/5725, there will be no need to supply runtime API defs anymore

shunjizhan:
I mean in this PR, we need to write the following runtime codes 3 times. In substrate there is no shortcut to write "shared runtime implementation"? 
```rust
		// required by xtokens precompile
		#[transactional]
		fn account_call(
			from: AccountId,
			to: H160,
			data: Vec<u8>,
			value: Balance,
			gas_limit: u64,
			storage_limit: u32,
			access_list: Option<Vec<AccessListItem>>,
			estimate: bool,
		) -> Result<CallInfo, sp_runtime::DispatchError> {
			let from = EvmAddressMapping::<Runtime>::get_or_create_evm_address(&from);

			Self::call(from, to, data, value, gas_limit, storage_limit, access_list, estimate)
		}

		fn account_create(
			from: AccountId,
			data: Vec<u8>,
			value: Balance,
			gas_limit: u64,
			storage_limit: u32,
			access_list: Option<Vec<AccessListItem>>,
			estimate: bool,
		) -> Result<CreateInfo, sp_runtime::DispatchError> {
			let from = EvmAddressMapping::<Runtime>::get_or_create_evm_address(&from);

			Self::create(from, data, value, gas_limit, storage_limit, access_list, estimate)
		}
	}
```

xlc:
oh. right. no. but we could write some macro to help it but macro in macro is going to make things complicated and it is mostly the method signature are duplicated and implementation are shared

PR #2716, opened 2024-02-22
evm tracing

ermalkaleci:


xlc (commented):
one issue is that this won't work well with indirect EVM calls. e.g. currency.trasnfer ERC20 asset

.github/workflows/tracing-runtime.yml
```diff
@@ -0,0 +1,48 @@
+name: Build and publish evm tracing runtime
+on:
+  release:
+    types: [published]
+  workflow_dispatch:
+    inputs:
+      tag:
+        description: Existing tag to build runtime. i.e. 1.0.0
```
ermalkaleci, line 8: this won't actually work because old tags don't have tracing feature :(

.github/workflows/tracing-runtime.yml
```diff
@@ -0,0 +1,55 @@
+name: Build and publish evm tracing runtime
+on:
+  release:
+    types: [published]
+  # Allow manual trigger for old releases
+  workflow_dispatch:
```
ermalkaleci, line 6: This is useful for the current version. We'll need to create a branch from old tag and include tracing feature and then trigger this action to build tracing runtime and push to specified tag.

xlc (dismissed):

zjb0807 (dismissed):

xlc (approved):

ermalkaleci:
> one issue is that this won't work well with indirect EVM calls. e.g. currency.trasnfer ERC20 asset

#2731 will address this

PR #2724, opened 2024-03-19
only make contract unpublished when deployer is in developer mode

### LINKED ISSUE #2683: new flow for deploying evm contract
The developer mode and contract publish thing has introduced a lot of confusion for developers, and it's better to an "opt in" feature when needed, instead of an "extra task" for every developer.

A potential better flow can be something like this
- by default every EOA have developer mode off, and the contracts deployed by them are auto published. This will simplify our current work flow, and align with classic evm flow
- when a developer wants to test a contract before anyone else can interact with it, he can then turn developer mode on for his account. So that every contract by this developer account is unpublished at deployment, and needs manual publish.

xlc:
closes #2683

### DIFF
--- a/modules/evm/src/lib.rs
+++ b/modules/evm/src/lib.rs
@@ -1126,7 +1126,6 @@ pub mod module {
 					}
 
 					if info.exit_reason.is_succeed() {
-						Self::mark_published(contract, Some(source))?;
 						Pallet::<T>::deposit_event(Event::<T>::ContractPublished { contract });
 					}
 
@@ -1777,7 +1776,7 @@ impl<T: Config> Pallet<T> {
 
 	fn is_developer_or_contract(caller: &H160) -> bool {
 		let account_id = T::AddressMapping::get_account_id(caller);
-		Self::query_developer_status(account_id) || Self::is_contract(caller)
+		Self::query_developer_status(&account_id) || Self::is_contract(caller)
 	}
 
 	fn reserve_storage(caller: &H160, limit: u32) -> DispatchResult {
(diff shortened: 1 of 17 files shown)


modules/evm/src/lib.rs
```diff
@@ -1126,7 +1126,6 @@ pub mod module {
 					}
 
 					if info.exit_reason.is_succeed() {
-						Self::mark_published(contract, Some(source))?;
 						Pallet::<T>::deposit_event(Event::<T>::ContractPublished { contract });
 					}
```
zjb0807, line 1130: remove the event

zjb0807 (commented):
I feel that the developer status will become useless, no one will use this function, and it may be deprecated in the future.

xlc:
yeah we can check onchain status and see if anyone is using it

modules/evm/src/tests.rs
```diff
@@ -892,8 +874,6 @@ fn create_predeploy_contract_works() {
 			used_storage: 284,
 		}));
 
-		System::assert_last_event(RuntimeEvent::EVM(crate::Event::ContractPublished { contract: addr }));
```
ermalkaleci, line 895: are we changing auto publish for predeploy?
https://github.com/AcalaNetwork/Acala/pull/2724/files#diff-db3ad6e2ca083dd40dacda29e393a5d47e24baae25ab8efcab355ed09c5db7efL1128-L1132

zjb0807 (approved):

modules/evm/src/tests.rs
ermalkaleci: oh yes, NetworkContractOrigin is not developer so it will be published

ermalkaleci (approved):

Issue #1764, opened 2022-01-10
EVM Tracing RPCs


shaunxw:
https://openethereum.github.io/JSONRPC-trace-module

PR #2731, opened 2024-03-29
improve tracing

ermalkaleci:


modules/evm/src/runner/state.rs
```diff
@@ -873,15 +873,6 @@ impl<'config, 'precompiles, S: StackState<'config>, P: PrecompileSet> StackExecu
...
 
-		event!(Create {
-			caller,
-			address,
-			scheme,
-			value,
-			init_code: &init_code,
-			target_gas
-		});
-
```
zjb0807, line 884: We plan to implement [debug_traceTransaction](#1764), these events may be useful, it is best to keep it

ermalkaleci: it is moved on create method. I believe its the correct place to be. It has enter and exit. Tracing is implemented on chopsticks https://github.com/AcalaNetwork/chopsticks/pull/673

ermalkaleci: First enter event is `transact_create` which forwards the call to `create_inner` and then `emit_exit`
Second is `fn create(` which is called during execution. It will emit create and exit event. Same logic for call method

zjb0807 (approved):

xlc (approved):

Issue #2693, opened 2024-01-24
implement `block.prevrandao`


zjb0807:
add an issue to implement this

_Originally posted by @xlc in https://github.com/AcalaNetwork/Acala/pull/2691#discussion_r1462521729_

PR #2736, opened 2024-04-11
Impl block prevrandao

zjb0807:
Closes: #2693


runtime/acala/src/lib.rs
```diff
@@ -1469,6 +1469,33 @@ impl<I: From<Balance>> frame_support::traits::Get<I> for TxFeePerGasV2 {
 	}
 }
 
+pub struct RandomnessSource<T>(sp_std::marker::PhantomData<T>);
+impl<T: frame_system::Config> Randomness<T::Hash, BlockNumber> for RandomnessSource<T> {
+	fn random(subject: &[u8]) -> (T::Hash, BlockNumber) {
+		let relay_storage_root = ParachainSystem::validation_data()
+			.expect("set in `set_validation_data`")
```
xlc, line 1476: ValidationData is removed at `on_initialize` and set at the inherent, this means it could be empty in the `on_initialize` hook for some pallets and some other inherents
so this could fail when invoked by scheduler or some other pallet's `on_initialize` hook

runtime/acala/src/lib.rs
```diff
@@ -1469,6 +1469,33 @@ impl<I: From<Balance>> frame_support::traits::Get<I> for TxFeePerGasV2 {
 	}
 }
 
+pub struct RandomnessSource<T>(sp_std::marker::PhantomData<T>);
```
xlc, line 1472: we should be able to make this reusable across runtime by having `ParachainSystem` to be passed as a generic parameter

runtime/acala/src/lib.rs
xlc: we better to avoid expect here, when no random value is available, we can return an insecure random value and block number of zero

ermalkaleci:
I have 2 concerns with this:
1. Is this going to work for tx executive on_initialize?
2. This is a read and doing hashing every time is called. Any risk this can be abused?

runtime/common/src/lib.rs
```diff
@@ -408,6 +412,55 @@ where
...
+						.ok()
+						.flatten()
+					{
+						randomness = current_block_randomness;
+						// maybe have a - 4 will make it secure enough
+						block_number = frame_system::Pallet::<T>::block_number()
+							.saturating_sub(One::one())
+							.saturating_sub(One::one())
+							.saturating_sub(One::one())
+							.saturating_sub(One::one());
```
xlc, line 448:
```suggestion
						// the randomness is from relaychain so there is a delay
						// have a - 4 to indicate the randomness is from previous relay block
						block_number = frame_system::Pallet::<T>::block_number()
							.saturating_sub(4u.into());
```

runtime/common/src/lib.rs
```diff
@@ -408,6 +412,55 @@ where
...
+					}
+				}
+			}
+		}
+
+		let mut subject = subject.to_vec();
```
xlc, line 454:
```suggestion
        // Unable to access relay randomness so use an insecure one
        // and mark it as stale with block number of zero
		let mut subject = subject.to_vec();
```

xlc (approved):

xlc:
> 1. Is this going to work for tx executive on_initialize?

An insecure random value that's based on parent block hash will be provided. But there are only few limited ways to trigger EVM call in `on_initialize` including scheduler and XCM. XCM processing will be moved out from on_initialize in future and scheduler usage is governance protected so it should be fine.

> 2. This is a read and doing hashing every time is called. Any risk this can be abused?

I guess we can cache the value? Otherwise we need to ensure we charge enough gas.

modules/currencies/src/mock.rs
```diff
@@ -286,8 +290,8 @@ pub fn deploy_contracts() {
 				H256::from_slice(&buf).as_bytes().to_vec()
 			},
 		}],
-		used_gas: 1235455,
-		used_storage: 5131,
+		used_gas: 1215220,
```
ermalkaleci, line 293: why is gas changing?

ermalkaleci:
> > 1. Is this going to work for tx executive on_initialize?
> 
> An insecure random value that's based on parent block hash will be provided. But there are only few limited ways to trigger EVM call in `on_initialize` including scheduler and XCM. XCM processing will be moved out from on_initialize in future and scheduler usage is governance protected so it should be fine.
> 
> > 2. This is a read and doing hashing every time is called. Any risk this can be abused?
> 
> I guess we can cache the value? Otherwise we need to ensure we charge enough gas.

I think we need to be careful with this, at least enable only on mandala & karura first

xlc:
It is known prevrandao have its limitation and cannot be used solely for cases requires strong random number, extra randomness source are required so I am not too worried about abuse.

ts-tests/contracts/ExplicitRevertReason.sol
```diff
@@ -1,6 +1,6 @@
 // SPDX-License-Identifier: GPL-3.0-or-later
 
-pragma solidity 0.8.2;
+pragma solidity ^0.8.2;
```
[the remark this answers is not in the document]
zjb0807: We need to upgrade the solidity compiler Version. A unified version will be easier to maintain.

xlc (approved):

PR #2741, opened 2024-04-24
Polkadot sdk 1.9.0

zjb0807:
https://github.com/nutsfinance/stable-asset/pull/54

~~TODO:~~
~~- [ ] The front end needs to be updated to use xcm v4 to call the xtokens contract~~


runtime/acala/src/lib.rs
```diff
@@ -1285,11 +1301,18 @@ impl module_asset_registry::Config for Runtime {
 	type WeightInfo = weights::module_asset_registry::WeightInfo<Runtime>;
 }
 
+parameter_type_with_key! {
+	pub MinimalShares: |_pool_id: PoolId| -> Balance {
+		0 // TODO: config
```
xlc, line 1306: this should read the current from PoolId and return the ED of it

xlc (approved):

Issue #2766, opened 2024-05-30
migrate all of the benchmarks with `frame-benchmarking`


zjb0807:
enable this after all of the benchmarks migrated to `frame-benchmarking`

_Originally posted by @zjb0807 in https://github.com/AcalaNetwork/Acala/pull/2765#discussion_r1620164451_

PR #2767, opened 2024-06-05
migrate asset-registry benchmark

zjb0807:
#2766

### DIFF
--- a/modules/asset-registry/Cargo.toml
+++ b/modules/asset-registry/Cargo.toml
@@ -17,21 +17,28 @@ primitives = { workspace = true }
 xcm = { workspace = true }
 
 module-support = { workspace = true }
+module-evm = { workspace = true }
+serde_json = { workspace = true }
+hex = { workspace = true }
+
+
+# Benchmarks
+frame-benchmarking = { workspace = true, optional = true }
 
 [dev-dependencies]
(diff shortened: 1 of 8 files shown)


zjb0807:
* /bench runtime mandala *

modules/asset-registry/Cargo.toml
```diff
@@ -17,21 +17,28 @@ primitives = { workspace = true }
 xcm = { workspace = true }
 
 module-support = { workspace = true }
+module-evm = { workspace = true }
+serde_json = { workspace = true }
+hex = { workspace = true }
```
zjb0807, line 22: The new benchmark requires importing the associated module as a dependency, which I’m not sure is what we want.
https://github.com/AcalaNetwork/Acala/blob/a6bb42cab4312a2d3c935abcc741b2ca6f3dc93c/runtime/mandala/src/benchmarking/dex_oracle.rs#L51-L52
the dex-oracle need to import the `module-currencies`, `module-dex`, `pallet-aura`, `pallet-balances`, `pallet-timestamp`, `sp-consensus-aura`,  it feels too bulky

xlc: can we mock it?

xlc (approved):

PR #2776, opened 2024-07-08
earning precompile

wangjj9219:


primitives/src/bonding/ledger.rs
```diff
@@ -29,9 +29,9 @@ use frame_support::pallet_prelude::*;
 #[derive(PartialEq, Eq, Clone, Encode, Decode, RuntimeDebug, MaxEncodedLen, TypeInfo)]
 pub struct UnlockChunk<Moment> {
 	/// Amount of funds to be unlocked.
-	value: Balance,
+	pub value: Balance,
```
xlc, line 32: instead of making the fields pub, we can have a pub method to return the values as a tuple or ethabi encoded value

runtime/common/src/precompile/earning.rs
```diff
@@ -0,0 +1,691 @@
...
+			assert_eq!(Currencies::free_balance(ACA, &alice()), 100_000_000_000_000);
+
+			// bond(address,uint256) -> 0xa515366a
+			// who 0x1000000000000000000000000000000000000001
+			// amount 20_000_000_000_000
+			let input = hex! {"
+                a515366a
+                000000000000000000000000 1000000000000000000000000000000000000001
+                00000000000000000000000000000000 0000000000000000000012309ce54000
+            "};
```
zjb0807, line 378: format this and other places

runtime/common/src/precompile/earning.rs
```diff
@@ -0,0 +1,691 @@
...
+	Permill, RuntimeDebug,
+};
+use sp_std::{marker::PhantomData, prelude::*};
+
+/// The Earning precompile
+///
+/// `input` data starts with `action`.
+///
+/// Actions:
+/// - Mint. `input` bytes: `who`.
```
zjb0807, line 45:
```suggestion
/// - Bond. `input` bytes: `who`.
```

xlc (approved):

zjb0807 (approved):

Issue #2777, opened 2024-07-19
EVMTraceApi throws `{ invalid: { stale: null } }`


shunjizhan:
when I try to call `evmtraceApi.traceExtrinsic`, it throws `{ invalid: { stale: null } }`. It seems like the api didn't skip validity check for the tx, and so marked it as outdated? 

For trace runtime api, we probably should skip all these checks, and execute the tx directly

## reproduce
first run a chopsticks acala fork, and overrides the runtime with [runtime 2250 with tracing](https://api.github.com/repos/AcalaNetwork/Acala/releases/assets/168198977)

then run the following code, with the latest acala types that already includes types for the trace api
```
    "@acala-network/api": "^6.1.3-0",
    "@acala-network/types": "^6.1.3-0",
```
```ts
import '@acala-network/types'
import { withAcalaTypes } from '@acala-network/api';
import { ApiPromise, WsProvider } from '@polkadot/api';

async function main() {
  const provider = new WsProvider('ws://127.0.0.1:8000');
  const api = await ApiPromise.create(withAcalaTypes({ provider }));

  // randomly selected evm tx
  const extrinsicHex = '0x41038400421c7a4b5839385d9453cca4dc139f9aba94e426befc5376dde8510b86d8a06803fa61d4950fe1841c7b274262de97c53628ff006da1b007d4bde9cfed6b1f2fdc063efe67b500d983c1602b63cc93ec2a7e3a938db7b0cd72da92a9af2e0650291b000c00820f007fe92ec600f15cd25253b421bc151c51b0276b7d11017b0472f000000000000000000000000000000000000000000000000000000000000000030000000000000000000000000000000000000000000000000000020a3d37de000007b276d2481742d4240000';
  
  const traceConf = { CallTracer: null };
  const res = await api.call.evmTraceApi.traceExtrinsic(extrinsicHex, traceConf);
  console.log(res.asErr.toJSON()) // { invalid: { stale: null } }

  await api.disconnect();
}

main();
```

PR #2778, opened 2024-07-23
Skip check nonce and valid when enable the tracing feature

zjb0807:
Closes: #2777

### DIFF
--- a/runtime/karura/src/lib.rs
+++ b/runtime/karura/src/lib.rs
@@ -2560,9 +2560,12 @@ impl Convert<(RuntimeCall, SignedExtra), Result<(EthereumTransactionMessage, Sig
 				access_list,
 				valid_until,
 			}) => {
-				#[cfg(not(feature = "tracing"))]
 				if System::block_number() > valid_until {
-					return Err(InvalidTransaction::Stale);
+					if cfg!(feature = "tracing") {
+						// skip check when enable tracing feature
+					} else {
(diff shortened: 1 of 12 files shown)


runtime/karura/src/lib.rs
```diff
@@ -2560,6 +2560,7 @@ impl Convert<(RuntimeCall, SignedExtra), Result<(EthereumTransactionMessage, Sig
 				access_list,
 				valid_until,
 			}) => {
+				#[cfg(not(feature = "tracing"))]
```
xlc, line 2563: can we only disable the check only if the code is being traced? so that it is harder to accidentally disable this check in production execution

runtime/karura/src/lib.rs
```diff
@@ -2606,6 +2610,11 @@ impl Convert<(RuntimeCall, SignedExtra), Result<(EthereumTransactionMessage, Sig
 					decode_gas_price(gas_price, gas_limit, TxFeePerGasV2::get()).ok_or(InvalidTransaction::Stale)?;
 
 				if System::block_number() > valid_until {
+					if cfg!(feature = "tracing") {
+						// skip check when enable tracing feature
+					} else {
+						return Err(InvalidTransaction::Stale);
+					}
 					return Err(InvalidTransaction::Stale);
```
xlc, line 2618: this is not removed?

zjb0807: This is a mistake

Issue #2467, opened 2023-01-23
state trie to version 1


zjb0807:
https://github.com/paritytech/substrate/pull/10073
https://github.com/paritytech/polkadot/pull/6336
https://github.com/paritytech/cumulus/pull/1424

xlc:
Let's do this after Statemint successfully migrated

PR #2799, opened 2024-08-28
update state version to 1 on acala

zjb0807:
#2467

follow up #2759 #2762

statetriemigration on karura mainnet: https://karura.subscan.io/event?module=statetriemigration&page=1&time_dimension=date&page_size=25


runtime/karura/src/lib.rs
```diff
@@ -1969,8 +1947,6 @@ construct_runtime!(
 		// Parachain System, always put it at the end
 		ParachainSystem: cumulus_pallet_parachain_system = 30,
 
-		StateTrieMigration: pallet_state_trie_migration = 254,
```
xlc, line 1972: do we need a migration to cleanup storages?

zjb0807:
tested `--sync=warp` on karura mainnet.
image

xlc (approved):

PR #2803, opened 2024-09-02
allow signed migration

xlc:
will include this into 2260 runtime

### DIFF
--- a/runtime/karura/src/lib.rs
+++ b/runtime/karura/src/lib.rs
@@ -1837,6 +1837,10 @@ impl orml_parameters::Config for Runtime {
 	type WeightInfo = ();
 }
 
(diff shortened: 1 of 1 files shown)


zjb0807 (approved):
It's better to adjust the deposit
```	
pub MigrationSignedDepositPerItem: Balance = dollar(KAR);
pub MigrationSignedDepositBase: Balance = dollar(KAR);
```

https://github.com/polkadot-fellows/runtimes/blob/cf15699e1527f3bf4c76abd0287e62721bee428c/relay/polkadot/src/lib.rs#L1540-L1541

xlc:
only a single account can trigger migration so high deposit is not required

PR #2858, opened 2025-02-03
Support DOT reseve on AH and Snowfork bridged assets

xlc:
Accept AH as DOT reserve
Accept AH as bridges asset reserve

This enables Snowfork integration


alistair-singh (approved):
Looks good!

zjb0807 (approved):

PR #2906, opened 2025-08-17
Update polkadot-stable2503

zjb0807:
https://github.com/nutsfinance/stable-asset/pull/58


modules/evm/src/lib.rs
```diff
@@ -643,6 +660,7 @@ pub mod module {
 				gas_limit,
 				storage_limit,
 				access_list.into_iter().map(|v| (v.address, v.storage_keys)).collect(),
+				vec![], // TODO
```
xlc, line 663: what is this? is this needs to be done now or later?

PR #2915, opened 2025-09-24
Update xtoken reserve provider

zjb0807:
* add `XtokensParameters` to store `ReserveLocation` with `orml_parameters` pallet
* add `ParametersMigrateToV1` to setup `ReserveLocation`  of `XtokensParameters`

Since the `chain_part` and `non_chain_part` called by `xtokens` are [hard-coded](https://github.com/open-web3-stack/open-runtime-module-library/blob/c02a0baca184d91071e5516de4439df0eddd9d6d/traits/src/location.rs#L22), if we want to use xtokens after AHM, we need to upgrade the runtime first.

### DIFF
--- a/.github/workflows/test.yml
+++ b/.github/workflows/test.yml
@@ -134,7 +134,7 @@ jobs:
         with:
           node-version: 18.x
       - name: Install deps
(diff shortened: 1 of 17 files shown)


zjb0807:
It seems that adding `MigratedStatus` storage to `xtokens` to control it would be better. No runtime upgrade

PR #2926, opened 2025-10-01
add xcm migration for xcm v5

xlc:


zjb0807:
#2877 This migration has already been executed. But here should be no problem.

xlc:
we need it for every new XCM version. this release upgrades from v4 to v5

